PR #43, opened 2018-02-06
Make build target path platform agnostic

BrianAllred:
This fixes building Blazor on file systems and platforms that use the forward-slash character as a path separator.

Tested with

* ArchLinux / DotNet SDK 2.1.4
* Windows 10 / DotNet SDK 2.1.4
* Windows 10 / Visual Studio 2017 version 15.5.6

PR #58, opened 2018-02-07
Fixed issue #44 with build when using a directory path with a space in it

MarkGaze:
Added **\&quot;** characters to the paths in the build targets and properties to take into account directory paths with a space in them

Trimmed any trailing backslashes from the ProjectDir path as this would escape the quotemark.

Also, fixed a backslash that was missed in #43

Addresses #44 

Tested on:

- Windows 10 / DotNet SDK 2.1.4
- Ubuntu 16.04 / DotNet SDK 2.1.4

And tested with a directory without a space.


d-dizhevsky:
Also fixes an issue #59

SteveSandersonMS:
Thanks very much @MarkGaze!

PR #67, opened 2018-02-07
Setup our patched illink.exe

javiercn:
* Added the mono linker submodule under src/mono/tools/illink
* Applied the patches as in https://github.com/SteveSanderson/Blazor/tree/6624bd45c9bbade1995ec3d8f0196d51b7b3b629/samples/MicroApp/tools/illink
* Added the instructions on how to build the patched version of the linker.
* Added a ready to use version of the linker under src/mono/tools


SteveSandersonMS:
Great - getting illink in makes sense as a first step. A couple of questions though:

[1] The version built this way will rely on .NET Framework, won't it? For it to be involved in a cross-platform build, we'll need a .NET Core build.
[2] Are the patches I made still required? I seem to recall they were originally because of missing references to `Mono.Security`, but that assembly is present in the current Mono BCL in our repo, so it might be even that it works without patching. Just checking!

For [1], I guess maybe you're putting this in only as a step towards optimisation investigation and aren't planning to invoke them from the MSBuild publish task at this stage. If so we can probably do with the .NET Framework build in the short term.

javiercn:
@SteveSandersonMS I'm only planning to work on the investigation for now. We'll have another task to make this part of the build proper. I think we should focus on doing all the invesgations/manual work first and then plug-it into the build system.

SteveSandersonMS:
Makes sense - thanks for the info!

rynowak:
@javiercn should this be merged so the code doesn't rot?

PR #93, opened 2018-02-09
[WIP] injecting blazor boot code and/or config data more intelligently

ebekker:
Taking a stab at this TODO:

This allows for either scenario in the TODO comments -- it injects either the boot loader or just the boot config data.

* If their exists a user-owned `script` tag with a attribute `blazor="boot"` it is assumed the user has the responsibility of loading the boot script and we inject a new script tag at the very top of body that contains only the boot config data (`type="blazor-config"`).
* Otherwise, we inject the full blazor boot + config data script tag (same as before), 
  * just before the first user-owned script tag, or
  * at the very end of the body if there is none, to prevent blocking.

I've tested this out in the 3 supported scenarios, seems to work as expected.

Looking for feedback.

### DIFF
--- a/samples/HostedInAspNet.Client/wwwroot/index.html
+++ b/samples/HostedInAspNet.Client/wwwroot/index.html
@@ -6,5 +6,6 @@
 </head>
 <body>
   <app>Loading...</app>
+  <script type="blazor-boot"></script>
 </body>
 </html>
--- a/samples/MonoSanity/wwwroot/index.html
+++ b/samples/MonoSanity/wwwroot/index.html
@@ -57,6 +57,7 @@
 
   <p id="loadingIndicator">Loading...</p>
 
+  <script type="blazor-boot"></script>
   <script src="loader.js"></script>
   <script>
     initMono(['/_framework/_bin/MonoSanityClient.dll'], function () {
--- a/samples/StandaloneApp/wwwroot/index.html
+++ b/samples/StandaloneApp/wwwroot/index.html
@@ -5,6 +5,7 @@
     <title>Blazor standalone</title>
 </head>
 <body>
-  <app>Loading...</app>
+    <app>Loading...</app>
+    <script type="blazor-boot"></script>
 </body>
 </html>
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Boot.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Boot.ts
@@ -6,13 +6,15 @@ import './GlobalExports';
 async function boot() {
   // Read startup config from the <script> element that's importing this file
   const allScriptElems = document.getElementsByTagName('script');
-  const thisScriptElem = allScriptElems[allScriptElems.length - 1];
-  const entryPoint = thisScriptElem.getAttribute('main');
+  const thisScriptElem = document.currentScript || allScriptElems[allScriptElems.length - 1];
+  // Try to find the script element that has our bootstrap configuration, or default to this
+  const configScriptElem = document.querySelector('script[type=="blazor-config"]') || thisScriptElem;
+  const entryPoint = configScriptElem.getAttribute('main');
   if (!entryPoint) {
-    throw new Error('Missing "main" attribute on Blazor script tag.');
+    throw new Error('Missing "main" attribute on Blazor Config script tag.');
   }
   const entryPointAssemblyName = getAssemblyNameFromUrl(entryPoint);
-  const referenceAssembliesCommaSeparated = thisScriptElem.getAttribute('references') || '';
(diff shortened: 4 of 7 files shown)


ebekker:
This is a build-time operation, it generates the index.html file which is statically stored and served up.

SteveSandersonMS:
Thanks very much for contributing this! It looks like a great start.

Based on your progress, I've been thinking about the design a bit more, and wonder if we could simplify it further. I think that:

* We should make it obvious to developers how to control where the `<script>` tag goes, without them having to look up anything in docs. So it would be better if, by default, projects always have something that explicitly shows where it's going, rather than some rule like "insert above first developer-controlled script tag"
* We should also make it obvious how to control whether the `<script>` tag has `async` or `defer` or whatever else.

**Proposed design simplification:**

What if we say your `index.html` must contain a script tag of type `blazor-boot`, i.e.,

    <script type="blazor-boot"></script>

Then in default project template, we put that at the end of `<body>`, but the developer can freely move it anywhere else they want (e.g., into `<head>` or whatever they like).

And on build, we process it like this:

 1. Remove the `type` attribute altogether (since `javascript` is the default when not specified)
 2. Set the `src` attribute and the other existing Blazor boot config attribute data
 3. Leave any other attributes (e.g., `async` or `defer` or anything else) unchanged

Then we don't need to inject extra script tags with custom config formats - it's always just one script tag, where the developer controls placement and any extra attributes.

@ebekker, what do you think?

ebekker:
Thanks @SteveSandersonMS, so I wanted to make sure I understand fully.

So the idea is that whenever a Blazor project is initialized, we would put some guiding details for the user, something along these lines:

**`wwwroot/index.html`:**
```html
<body>
<head>...</head>
<body>
  <!-- place this tag before any of your code calls Blazor APIs -->
  <script type="blazor-boot" userTag1 userTag2="..."></script>
</body>
</html>
```
and after a build, this would be the generated dist output:
**`.../dist/wwwroot/index.html`:**
```html
<body>
<head>...</head>
<body>
  <!-- place this tag before any of your code calls Blazor APIs -->
  <script src="(blazor-overwritten)" main="(blazor-overwritten)" references="(blazor-overwritten)" userTag1 userTag2="..."></script>
</body>
</html>
```

So a couple questions,
* What do you think should be the response if the "blazor-boot" tag is missing?  Should the tools just fail (throw) or make a best effort?
* If best effort?  Is that where the existing logic might be used?  (i.e. insert it just before the first user-defined script tag or at the end of the body if there is none)

SteveSandersonMS:
@ebekker Yes, that sounds about perfect! We probably don't need the comment tag in the examples as people will understand the `<script>` tag placement anyway, but other than that, it looks ideal.

> What do you think should be the response if the script tag with "blazor-boot" is missing? Should the build tool just fail (throw) or make a best effort?

I'd suggest if there's no `<script type="blazor-boot">`, then we just do nothing, i.e., we output the `index.html` file without changes. The developer might have come up with their own special way of booting Blazor, and we don't need to interfere or consider it an error. It makes it simpler to understand and explain as well, rather than having any further rules about fallback logic that are hard to anticipate.

ebekker:
@SteveSandersonMS, I've incorporated your feedback and re-design into the PR.

Additionally, I've fixed up existing, or added new, tests to accommodate this change, and also updated all the samples to work with this new boot injection strategy.

One thing to note with this change, because I'm using AngleSharp to parse, manipulate and emit the index.html file, it's no longer possible to preserve completely the formatting, and even the content of the original input file.  For example, a poorly structured input file, that's missing <html>, <head> and <body> tags will now be *corrected*.  The semantic structure of the input file is still intact, but AngleSharp does do some *fixing up* of the document.

SteveSandersonMS:
This looks perfect - thanks so much @ebekker! I'll merge it on Monday if that's OK.

SteveSandersonMS:
Thanks again - it's [now merged](https://github.com/aspnet/Blazor/commit/4c7480a8406156510bcd44ed52454d89d53755b5)

> One thing to note with this change, because I'm using AngleSharp to parse, manipulate and emit the index.html file, it's no longer possible to preserve completely the formatting, and even the content of the original input file.

Yes, that is a really important point, and I'm glad you pointed it out. While merging, I realised there may be some scenarios where people find it problematic if we modify their HTML formatting. So I've altered the logic so it [injects the tag using `HtmlTokenizer` instead of `HtmlParser`](https://github.com/aspnet/Blazor/blob/e83f6029df5ff47aa5697c98bfff8c79eaf42beb/src/Microsoft.AspNetCore.Blazor.Build/Core/FileSystem/IndexHtmlFileProvider.cs#L57), which now means it retains the original source formatting. I hope you don't mind me making that change!

PR #94, opened 2018-02-09
bug fix: unused parameter in the constructor

agametov:


SteveSandersonMS:
Thanks for proposing this!

For now I'd like to leave this file unchanged, because it's actually copied from https://github.com/aspnet/Razor/blob/ff40124594b58b17988d50841175430a4b73d1a9/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CodeWriterExtensions.cs, and at some stage it might be possible to reuse the existing implementation without copying it (so it's useful to know they are identical).

If you wanted to submit a PR for the original file at https://github.com/aspnet/Razor/blob/ff40124594b58b17988d50841175430a4b73d1a9/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CodeWriterExtensions.cs that would be great though!

agametov:
ok! 👍

agametov:
https://github.com/aspnet/Razor/pull/2044

PR #97, opened 2018-02-10
Added fast activator for generic and types

vbondaryuk:
Added FastActivator implementation. This implementation two-time faster than standart 
```
Activator.CreateInstance(typeof(A))
```
I think this changes will help for up performance
```
BenchmarkDotNet=v0.10.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-3632QM CPU 2.20GHz, ProcessorCount=8
Frequency=2143575 Hz, Resolution=466.5104 ns, Timer=TSC
Host Runtime=Clr 4.0.30319.42000, Arch=32-bit RELEASE [AttachedDebugger]
GC=Concurrent Workstation
JitModules=clrjit-v4.7.2600.0
Job Runtime(s):
Clr 4.0.30319.42000, Arch=32-bit RELEASE


               Method |       Mean |    StdDev |     Median |
--------------------- |----------- |---------- |----------- |
  FastActivatorByType | 47.5745 ns | 1.3975 ns | 47.2293 ns |
 GenericFastActivator | 10.7458 ns | 0.3119 ns | 10.6596 ns |
    StandartActivator | 92.7500 ns | 2.3080 ns | 92.2107 ns 
```

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser/Interop/Json.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Browser/Interop/Json.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using Microsoft.AspNetCore.Blazor.Activators;
 
 namespace Microsoft.AspNetCore.Blazor.Browser.Interop
 {
@@ -32,7 +33,7 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Interop
                 // Return default value for type
                 if (typeOfT.GetTypeInfo().IsValueType)
                 {
-                    return Activator.CreateInstance(typeOfT);
+                    return FastActivator.CreateInstance(typeOfT);
                 }
                 else
                 {
@@ -112,7 +113,7 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Interop
             }
             else if (deserializedValue is Dictionary<string, object>)
             {
-                var result = Activator.CreateInstance(typeOfT);
+                var result = FastActivator.CreateInstance(typeOfT);
                 var deserializedPropertyDict = (Dictionary<string, object>)deserializedValue;
                 foreach (var propInfo in typeOfT.GetRuntimeProperties())
                 {
--- a/src/Microsoft.AspNetCore.Blazor/Activators/DynamicModuleLambdaCompiler.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Activators/DynamicModuleLambdaCompiler.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq.Expressions;
+using System.Reflection.Emit;
+
+namespace Microsoft.AspNetCore.Blazor.Activators
+{
+	public static class DynamicModuleLambdaCompiler
+	{
+		public static Func<T> GenerateFactory<T>() where T : new()
+		{
+			Expression<Func<T>> expr = () => new T();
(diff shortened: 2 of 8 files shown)


SteveSandersonMS:
Thanks for this. Would you be able to benchmark it while running on Mono on WebAssembly? I didn't think that Mono on WebAssembly did anything like JIT compilation for Reflection.Emit, so if the Reflection.Emit is working, it might be going through a slower interpreted implementation.

vbondaryuk:
@SteveSandersonMS This result for core and mono, also i am going to test on webassembly

```
BenchmarkDotNet=v0.10.12, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.192)
Intel Core i7-3632QM CPU 2.20GHz (Ivy Bridge), 1 CPU, 8 logical cores and 4 physical cores
Frequency=2143567 Hz, Resolution=466.5121 ns, Timer=TSC
.NET Core SDK=2.1.4
  [Host]         : .NET Core 2.0.5 (Framework 4.6.26020.03), 64bit RyuJIT  [AttachedDebugger]
  Core           : .NET Core 2.0.5 (Framework 4.6.26020.03), 64bit RyuJIT
  Mono 5.8.0.108 : Mono 5.8.0 (Visual Studio), 32bit 


               Method |            Job |        Runtime |      Mean |      Error |     StdDev |
--------------------- |--------------- |--------------- |----------:|-----------:|-----------:|
  FastActivatorByType |           Core |           Core |  46.85 ns |  1.0883 ns |  1.1644 ns |
 GenericFastActivator |           Core |           Core |  12.43 ns |  0.3691 ns |  0.7785 ns |
    StandartActivator |           Core |           Core |  73.78 ns |  1.5792 ns |  1.8186 ns |
  FastActivatorByType | Mono 5.8.0.108 | Mono 5.8.0.108 |  75.24 ns |  1.5494 ns |  1.2938 ns |
 GenericFastActivator | Mono 5.8.0.108 | Mono 5.8.0.108 |  18.13 ns |  0.4436 ns |  0.3932 ns |
    StandartActivator | Mono 5.8.0.108 | Mono 5.8.0.108 | 973.88 ns | 15.2055 ns | 14.2232 ns |
```

A have tested with this code:
```csharp
[CoreJob]
[MonoJob("Mono 5.8.0.108", @"C:\Program Files (x86)\Mono\bin\mono.exe")]
public class Bench
{
	[Benchmark]
	public object FastActivatorByType()
	{
		return FastActivator.CreateInstance(typeof(A));
	}

	[Benchmark]
	public object GenericFastActivator()
	{
		return FastActivator.CreateInstance<A>();
	}

	[Benchmark]
	public object StandartActivator()
	{
		return System.Activator.CreateInstance(typeof(A));
	}
}
```

SteveSandersonMS:
Thanks for the updated benchmarks. Indeed the WebAssembly results would be the key ones for Blazor.

However, before you go much further with this, I've realised that I'm not sure we'll be able to control the activator calls from inside Blazor code directly. Once we integrate with ASP.NET's standard dependency injection feature, I think the DI system will be responsible for instantiating component objects (so it can pass constructor parameters where needed). It's quite possible then that we won't have any `Activator.CreateInstance` calls (or FastActivator equivalents) in the Blazor core library anyway. Does that seem plausible?

vbondaryuk:
Yep, you are right, and I think that it should be better solution to use ASP.NET's standard dependency injection. But I'm not sure we'll use DI in current json parser implementation. May be I'm wrong. 
I'm ready to help in anyway.

vbondaryuk:
@SteveSandersonMS A have checked this changes on mono on WebAssembly and it is does not work(have not supported on this platform yet).
Also I want to ask it is possible to help your team with solving some small bugs and develop small minor features it is very interesting for me?

Issue #16, opened 2018-01-25
Publishing


danroth27:
- [ ] Support `dotnet publish` and produce a useful result for standalone apps
- [ ] Support `dotnet publish` and produce a useful result when hosted in ASP.NET
- [x] Linking (combining and stripping assemblies, eliminating facades)

SteveSandersonMS:
Done

iambipinpaul:
Few days before , I tried to publish it on GitHub page and it was not able to reference  the files inside of  ` 
_framework ` folder. how ever it worked on azure when I tried using Visual Studio publish tools to publish it on azure.  

can we serve this from a static server like GitHub page or GitLab Page?

danroth27:
Yup, in fact https://blazor-demo.github.io is a Blazor app hosted on GitHub Pages as static files.

iambipinpaul:
is there any extra tweak  we have to do or simply hosting all content of `dist` folder which will be get after running `dotnet publish` is enough?

let me try it again.

danroth27:
When tried this out I had to make a couple of tweaks. For example, I think GitHub pages ignores files/folders that start with `_` by default.

danroth27:
Also you need to enable support for SPA patterns on GitHub pages using something like https://github.com/rafrex/spa-github-pages. See https://github.com/blazor-demo/blazor-demo.github.io for an example deployment

iambipinpaul:
Thanks it worked ! 👍

rstropek:
@danroth27 Thank you for the tip regarding GitHub pages. I added it to https://learn-blazor.com/getting-started/static-hosting/. It now covers nginx/Docker, Webserver for Chrome, and GitHub Pages.

PR #99, opened 2018-02-10
Refactoring middleware configuration API

ebekker:
Trying to clean up the API for configuring server-side middleware for working with Blazor.

Refactoring the API to mimic the typical middleware options pattern, such as the simple [UseStaticFiles](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.staticfileextensions.usestaticfiles?view=aspnetcore-2.0)
or the [UseExceptionHandler](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.exceptionhandlerextensions.useexceptionhandler) extensions or even the more complex `UsMvc` and all its variations.

This pattern allows for easier growth, as any additional options can simply be added to the options class, and where it makes sense, we can create alternate convenience overloads.  But complete *configurability* can always be accessed by using the options parameter directly.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.DevHost/Server/Startup.cs
+++ b/src/Microsoft.AspNetCore.Blazor.DevHost/Server/Startup.cs
@@ -22,7 +22,7 @@ namespace Microsoft.AspNetCore.Blazor.DevHost.Server
             app.UseDeveloperExceptionPage();
 
             var clientAssemblyPath = FindClientAssemblyPath(app);
-            app.UseBlazorInternal(clientAssemblyPath);
+            app.UseBlazor(new BlazorOptions { ClientAssemblyPath = clientAssemblyPath });
         }
 
         private static string FindClientAssemblyPath(IApplicationBuilder app)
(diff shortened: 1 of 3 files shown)


ebekker:
A bit unrelated, but I noticed the config file parsing (`.blazor.config`) in this area is parsing the file using simple line positions.  This seems a bit fragile, should this be refactored, for example, into a JSON format.

So right now the first line is `SourceMSBuildPath` and the second line is the relative  `SourceOutputAssemblyPath`, but instead could be encoded as the JSON:

```json
{
    "SourceMSBuildPath": "...",
    "SourceOutputAssemblyPath": "..."
}
```
The corresponding generation of the config file in `All.targets` would be updated to match.

Just a thought...

SteveSandersonMS:
Thanks for this @ebekker! It's now merged.

> A bit unrelated, but I noticed the config file parsing (.blazor.config) in this area is parsing the file using simple line positions. This seems a bit fragile -- should this be refactored, for example, into a JSON format?

Yes, this area could definitely be improved. I think it was done as a plain text file just as a very quick way of being able to generate the file from MSBuild. But you're right that if more config ends up getting passed this way, JSON would be cleaner.

It's possible that we might not have `.blazor.config` forever. We'll know more when we come to implementing publishing (#16). We might either need something more sophisticated than `.blazor.config`, or maybe come up with a different approach altogether. That's why I didn't mind just doing it in a very simple way with a plain text file for now.

PR #101, opened 2018-02-11
[WIP] Adding ability to qualify the entry point of the main assembly

ebekker:
Trying to resolve the correct entry point of the main assembly.  I don't think this can be done the absolute *right way* without further support from mono-wasm -- i.e. a way to read the `EntryPoint` metadata from the assembly.

Until then, I'm proposing an alternative -- resolve the EntryPoint MethodInfo during the build process, and supply that detail in the boot config data, along with the entry point assembly.  I think this a pretty reasonable compromise and removes the *hard-codedness* of assuming an entry point with a namespace == assembly name, and fixed class and method names.

I'm submitting this PR to start a discussion on this approach -- these changes are tested and working, but they're a bit hacky, so if this approach is acceptable, there is a bit more clean up left.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .vs/
 bin/
 obj/
+
+*.csproj.user
--- a/samples/HostedInAspNet.Client/wwwroot/index.html
+++ b/samples/HostedInAspNet.Client/wwwroot/index.html
@@ -6,5 +6,6 @@
 </head>
 <body>
   <app>Loading...</app>
+  <script type="blazor-boot"></script>
 </body>
 </html>
--- a/samples/MonoSanity/wwwroot/index.html
+++ b/samples/MonoSanity/wwwroot/index.html
@@ -57,6 +57,7 @@
 
   <p id="loadingIndicator">Loading...</p>
 
+  <script type="blazor-boot"></script>
   <script src="loader.js"></script>
   <script>
     initMono(['/_framework/_bin/MonoSanityClient.dll'], function () {
--- a/samples/StandaloneApp/Program.cs
+++ b/samples/StandaloneApp/Program.cs
@@ -5,7 +5,15 @@ using Microsoft.AspNetCore.Blazor.Browser.Rendering;
 
 namespace StandaloneApp
 {
-    public class Program
+    public class ProgramX
+    {
+        public static void Main(string[] args)
+        {
+            new BrowserRenderer().AddComponent<Home>("app");
+        }
+    }
+
+    public class ProgramY
     {
         public static void Main(string[] args)
         {
--- a/samples/StandaloneApp/StandaloneApp.csproj
+++ b/samples/StandaloneApp/StandaloneApp.csproj
@@ -5,6 +5,7 @@
 
     <!-- Local alternative to <RunArguments>blazor serve</RunArguments> -->
     <RunArguments>run --project ../../src/Microsoft.AspNetCore.Blazor.DevHost serve</RunArguments>
+    <StartupObject>StandaloneApp.ProgramY</StartupObject>
   </PropertyGroup>
 
   <ItemGroup>
--- a/samples/StandaloneApp/wwwroot/index.html
+++ b/samples/StandaloneApp/wwwroot/index.html
@@ -5,6 +5,7 @@
     <title>Blazor standalone</title>
 </head>
 <body>
-  <app>Loading...</app>
+    <app>Loading...</app>
+    <script type="blazor-boot"></script>
 </body>
 </html>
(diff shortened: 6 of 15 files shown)


SteveSandersonMS:
Thanks - this makes sense! I definitely agree it would be good to stop having the hardcoded assumption about the entrypoint being `{assemblyname}.Program::Main`.

**Suggestion 1:**

It might be preferable not to use `Assembly.Load(clientApp)` inside the build process, just to avoid any possibility that loading the assembly into the build process has unexpected side-effects. Ideally this would be by using `Assembly.ReflectionOnlyLoad`, but that's not available on .NET Core as far as I know. But we could use [Mono's Cecil library to determine the entry point](https://github.com/jbevain/cecil/blob/master/Mono.Cecil/AssemblyDefinition.cs#L57) without using `Assembly.Load` - we're already using Cecil during the build process to discover the chain of references (https://github.com/aspnet/Blazor/blob/dev/src/Microsoft.AspNetCore.Blazor.Build/Core/FileSystem/FrameworkFiles/ReferencedAssemblyFileProvider.cs#L63).

**Suggestion 2:**

One minor point is that instead of the custom `{assemblyname}@{namespace},{type},{method}` format it would be nice if we used the standard CLR-style `[{assemblyname}]{namespace}.{type}::{method}()` format. Examples:

* `[YourApp]YourApp.Program::Main()`
* `[YourCompany.YourApp]YourCompany.YourNamespace1.YourNamespace2.AppStartup::StartupApp()`

I know it's a bit more work on the parsing side, but will at least be familiar at a glance to .NET developers.

Your thoughts?

grahamehorner:
I’m thinking why not decorate the entry point with a attribute, this would then allow better descriptive code generation and commenting around usage and if the entrypoint accepts any args, which at present is also hardcoded to an empty string array. 

attribute could be 
```
namespace myCompany
{

public static class myApp
{
[WebAssemblyEntryPoint()]
[WebAssemblyEntryPointAcceptsArgs(true)]
public static void main(string[] args)
{}
}
}
```
IMHO this would also open the door for having multiple entrypoints per wasm and could allow switching of wasm without additional downloading, assuming the generation of blazor.js was don’t in such away that the scripts supported it.   

This is simply a brain dump of an idea, would love to hear your thoughts before doing a POC/PR

ebekker:
@SteveSandersonMS, thanks for the feedback.

**Regarding suggestion 1:**  yes, already discovered that `ReflectionOnlyLoad` is not supported on Core.  Also looked at `System.Reflection.Metadata`, but that didn't include the needed metadata.  I didn't think of Cecil -- great idea, thanks, I'll look into incorporating that.

**Regarding suggestion 2:**  actually the way that I currently have it formatted was just a quick hack to experiment with the idea and test it out with minimal changes to the existing config structure and touch as little as possible of the existing code.  Moving forward with this approach, what I was thinking is actually keeping the `main` assembly attribute untouched and adding a new `entry-point` attribute to the boot config data.  This new attribute would use the format `{namespace}.{type}.{method}`, so a full example:

```html
<script src="/_framework/blazor.js"
    main="Sample.BlzrApp.dll" entry-point="Sample.BlzrApp.Program"
    references="lib1.dll,lib2.dll,..."></script>
```

This format is more consistent with the way the entry point is declared within an MSBuild file (the `StartupObject` property) or the compiler CLI parameter `/main`.  They simply let you qualify the class, the method name is assumed (required?) to be `Main`.

Now this does *force* you to use a Main method, but that is actually stipulated by the platform, so not something we're introducing in Blazor.  Thoughts?

ebekker:
@grahamehorner, if we stay with a console-style project format, I don't think it's necessary to add a new attribute for this purpose.  As the project format already has indicators in the form of the `StartupObject` property in the MSBuild file, which in turn gets passed on to the CSC compiler as a [parameter](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/main-compiler-option) to qualify the entry point.

However, I see that @SteveSandersonMS has indicated in #7 that he wants to move from `netcoreapp2.0` project type to `netstandard2.0` -- if that's the case, then we will need to come up with a way to flag the entry point.

Regarding entry point args, with the changes in #93, I think the simplest way to support that would be to allow a user-specified `args="arg1,arg2,arg3"` attribute in the boot `<script>` tag.  If found and the entry point method takes an array of string as a parameter, the entry point would get invoked with these args passed in.

grahamehorner:
@ebekker IMHO the args should simply follow the same as a commandline which is split on a space and not a comma the args can the be passed to the aspnet ConfigurationBuilder or what ever the developer wishes just like a console app; with regards to moving to netstandard 2.0 I can understand why and think any work/changes/development that is carried out should be mindful of this probability and should be done in such away that such a change wouldn’t have massive implications or a redevelopment.

I also believe that the injection of the script tag into the index.html may be something that limits options, ie. if I wish my hosting page to dynamically inject multiple wasm entrypoints with some server side user context/args I would be using a RazorPages *.cshtml with some code behind that injected the blazor-configure for each of the entrypoints

SteveSandersonMS:
Thanks again @ebekker - this is now merged :)

PR #114, opened 2018-02-14
Split Razor extensibility into its own assembly

rynowak:
This functionality will need to cross-compile to desktop framework
(net461) so that we can use it in VS. VS doesn't yet have netstandard2.0
support.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RenderTreeBuilder.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RenderTreeBuilder.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Blazor.Razor
+{
+    // Constants for method names used in code-generation
+    // Keep these in sync with the actual RenderTreeBuilder definitions
+    internal static class RenderTreeBuilder
+    {
+        public static readonly string OpenElement = nameof(OpenElement);
+
+        public static readonly string CloseElement = nameof(CloseElement);
+
+        public static readonly string OpenComponent = nameof(OpenComponent);
+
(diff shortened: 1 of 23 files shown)


rynowak:
No really cool changes here, just separating the Razor stuff into its own project so it can be included in VS. This helps prevent code rot in my VS prototyping work.

rynowak:
The next step for this will be to introduce latest builds of Razor and start checking in actual working VS bits. Let's discuss this at tomorrow's sync up.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RenderTreeBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public static readonly string OpenComponent = nameof(OpenComponent);
+
+        public static readonly string CloseComponent = nameof(CloseElement);
+
+        public static readonly string AddText = nameof(AddText);
+
+        public static readonly string AddAttribute = nameof(AddAttribute);
+        
+        public static readonly string Clear = nameof(Clear);
```
SteveSandersonMS: I suspect we won't need `Clear` or `GetFrames` here, as they would never be used in the codegenned code.

test/Microsoft.AspNetCore.Blazor.E2ETest/Infrastructure/BrowserFixture.cs
```diff
@@ -16,7 +16,18 @@ public BrowserFixture()
...
+            try
+            {
+                Browser = new RemoteWebDriver(opts);
+            }
+            catch (WebDriverException ex)
+            {
+                var message =
+                    "Failed to connect to the web driver. Please see the readme and follow the instructions to install selenium." +
+                    "Remember to start the web driver with `selenium-standalone start` before running the end-to-end tests.";
+                throw new InvalidOperationException(message, ex);
```
SteveSandersonMS: This is better than the uninformative error we had before.  Later on we could look into launching and stopping selenium server automatically from the tests. This is fine for now though :)

PR #132, opened 2018-02-17
Fixing build instructions in readme

MartynJones87:
The current build instructions the readme do not work as there are two solution files in the directory, so I've included the `Blazor.sln` onto the end of the build command.


danroth27:
Thanks!

PR #147, opened 2018-02-21
Components with children

SteveSandersonMS:
Work in progress


muqeet-khan:
In current implementation, I can do this in Home.cshtml to include a component from Shared:

`    <c:Shared.Header />`

However, I need to include the Full type name. Could you consider including `using Shared` when adding components so that anything in shared can be included in components in Pages by default?

SteveSandersonMS:
> However, I need to include the Full type name. Could you consider including using Shared when adding components so that anything in shared can be included in components in Pages by default?

That is already implemented. You can write:

    @using Shared
    <c:Header />

muqeet-khan:
Oh! I had tried that, then backed away because VS was showing it as an error. 

image

SteveSandersonMS:
Then your namespace probably isn't just `Shared`. Do you need to write something like `using StandaloneApp.Shared` instead?

PR #148, opened 2018-02-22
Add new Razor bits

rynowak:
- Adds the aspnetcore-dev feed as a package source
- Updates Razor to a relatively new build
- Fixes a few issues encountered with new bits

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,5 @@
+<Project>
+  <PropertyGroup>
+    <RazorPackageVersion>2.1.0-preview2-30159</RazorPackageVersion>
+  </PropertyGroup>
+</Project>
--- a/src/Microsoft.AspNetCore.Blazor.Build/Core/RazorCompilation/RazorCompiler.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Build/Core/RazorCompilation/RazorCompiler.cs
@@ -96,7 +96,7 @@ namespace Microsoft.AspNetCore.Blazor.Build.Core.RazorCompilation
                 var engine = new BlazorRazorEngine();
                 var blazorTemplateEngine = new BlazorTemplateEngine(
                     engine.Engine,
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.Blazor.Build/Core/RazorCompilation/RazorCompiler.cs
```diff
@@ -96,7 +96,7 @@ public class RazorCompiler
                 var engine = new BlazorRazorEngine();
                 var blazorTemplateEngine = new BlazorTemplateEngine(
                     engine.Engine,
-                    RazorProject.Create(inputRootPath));
+                    RazorProjectFileSystem.Create(inputRootPath));
```
rynowak: `RazorProject` has been obsoleted.

Deriving for `RazorTemplateEngine` has also been deprecated, and is in the process of being phased out from our repos. I have these code changes at home, I'll push them later as a follow up.

Directory.Build.props
```diff
@@ -0,0 +1,5 @@
+<Project>
+  <PropertyGroup>
+    <RazorPackageVersion>2.1.0-preview2-30159</RazorPackageVersion>
```
rynowak: I just grabbed a relatively new build. We can update this when we need new stuff. Due to the reliance on VS bits, we'll need to be a bit careful about it.

Ideally we can make improvements to a 'target' build of VS and then pull packages that include the same stuff after it makes its way into VS.

The alternative would be to install private builds of the Razor extension, which shouldn't be too bad either.

SteveSandersonMS: Does this have any impact on people wanting to use this with VS 15.5? My guess is no impact since we're not yet involved in VS internally. For our first distribution in ~2 weeks we want to remain compatible with VS15.5.

rynowak: Your guess is correct, anyone can use the runtime or msbuild-based compiler with just .NET CLI.

PR #151, opened 2018-02-22
Dependency injection

SteveSandersonMS:
Work in progress

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/InjectDirective.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/InjectDirective.cs
@@ -1,0 +1,116 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.CodeGeneration;
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Blazor.Razor
+{
+    // Much of the following is equivalent to Microsoft.AspNetCore.Mvc.Razor.Extensions's InjectDirective,
+    // but this one outputs properties annotated for Blazor's property injector, plus it doesn't need to
+    // support multiple CodeTargets.
+
+    internal class InjectDirective
+    {
+        const string InjectAttributeTypeName = "Microsoft.AspNetCore.Blazor.Components.InjectAttribute";
+
+        public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
+            "inject",
+            DirectiveKind.SingleLine,
+            builder =>
+            {
+                builder.Tokens.Add(DirectiveTokenDescriptor.CreateToken(
+                    DirectiveTokenKind.Type, optional: false));
+                builder.Tokens.Add(DirectiveTokenDescriptor.CreateToken(
+                    DirectiveTokenKind.Member, optional: false));
+                builder.Usage = DirectiveUsage.FileScopedMultipleOccurring;
+                builder.Description = "Inject a service from the application's service container into a property.";
+            });
+
+        public static void Register(IRazorEngineBuilder builder)
+        {
+            builder.AddDirective(Directive);
+            builder.Features.Add(new Pass());
+        }
+
+        private class Pass : IntermediateNodePassBase, IRazorDirectiveClassifierPass
+        {
+            protected override void ExecuteCore(
+                RazorCodeDocument codeDocument,
+                DocumentIntermediateNode documentNode)
+            {
+                var visitor = new Visitor();
+                visitor.Visit(documentNode);
+
+                var properties = new HashSet<string>(StringComparer.Ordinal);
+                var classNode = documentNode.FindPrimaryClass();
+
+                for (var i = visitor.Directives.Count - 1; i >= 0; i--)
+                {
+                    var directive = visitor.Directives[i];
+                    var tokens = directive.Tokens.ToArray();
+                    if (tokens.Length < 2)
+                    {
+                        continue;
+                    }
+
+                    var typeName = tokens[0].Content;
+                    var memberName = tokens[1].Content;
+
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Blazor/Components/InjectAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+
+namespace Microsoft.AspNetCore.Blazor.Components
+{
+    /// <summary>
+    /// Indicates that the associated property should have a value injected from the
+    /// service provider during initialization.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public class InjectAttribute : Attribute
```
rynowak: Will you want the `@inject` directive also?

I'm doing some of the directive stuff now so it will be easy to add.

src/Microsoft.AspNetCore.Blazor/Components/ComponentFactory.cs
```diff
@@ -0,0 +1,104 @@
...
+            return instance;
+        }
+
+        private void PerformPropertyInjection(IComponent instance)
+        {
+            var instanceType = instance.GetType();
+            if (!_cachedInitializers.TryGetValue(instanceType, out var initializer))
+            {
+                initializer = CreateInitializer(instanceType);
+                _cachedInitializers[instanceType] = initializer;
```
rynowak: Is thread safety a concern here?

src/Microsoft.AspNetCore.Blazor/Components/ComponentFactory.cs
```diff
@@ -0,0 +1,104 @@
...
+                    property.SetMethod)
+            )).ToArray();
+
+            // Return an action whose closure can write all the injected properties
+            // without any further reflection calls (just typecasts)
+            return instance =>
+            {
+                foreach (var injectable in injectables)
+                {
+                    var serviceInstance = _serviceProvider.GetService(injectable.propertyType);
```
rynowak: Any reason not to just reuse https://github.com/aspnet/DependencyInjection/blob/88c3bd6fe2786dd759b4a6c6d7c410e895336b6c/src/DI.Abstractions/ServiceProviderServiceExtensions.cs#L38

src/Microsoft.AspNetCore.Blazor/Components/InjectAttribute.cs
SteveSandersonMS: Ah, I just did implement that before I saw your comment: https://github.com/aspnet/Blazor/pull/151/commits/f57975118d0766e751ae70a4bb830948ee4f41b4#diff-c3f8f3bf88777effa8c797c34024165d

Does that look OK to you? I know it doesn't support multiple CodeTargets, but I'm not sure whether we need that (and it's not hard to add if so).

src/Microsoft.AspNetCore.Blazor/Components/ComponentFactory.cs
SteveSandersonMS: Good point. I was thinking no because all use of a single renderer occurs within the scope of a single thread, and a `ComponentFactory` is owned by a renderer. But maybe in the future our server-side prerenderer may choose to reuse a `ComponentFactory` across renderers (because why wouldn't you) so that would have to be thread-safe. I'll add the necessary locks.

SteveSandersonMS: Well, I'll make the underlying storage a `ConcurrentDictionary`.

src/Microsoft.AspNetCore.Blazor/Components/ComponentFactory.cs
SteveSandersonMS: I was just trying to make the error message clearer (i.e., by specifying which property on which type we failed to bind, rather than only saying which service type isn't registered).

rynowak: OK cool. I think that's a good thing

src/Microsoft.AspNetCore.Blazor/Components/InjectAttribute.cs
rynowak: I also commented elsewhere. Looks good to me.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/InjectDirective.cs
```diff
@@ -0,0 +1,114 @@
...
+                public override void VisitDirective(DirectiveIntermediateNode node)
+                {
+                    if (node.Directive == Directive)
+                    {
+                        Directives.Add(node);
+                    }
+                }
+            }
+
+            internal class InjectIntermediateNode : ExtensionIntermediateNode
```
rynowak: I think you already have a good intuition about this, but FYI:

The design intent of the IR is to provide nodes for all of the 'fundamental truths' of Razor. That includes, html, html attribute values (which have some special semantics), c# expressions, c# 'blobs', as well as generic c# structure like classes and namespaces. 

As such, all of these concepts in the IR are intentionally inflexible. What we define needs to work for everyone because they are fundamental to the language. You can implement these concepts in a different way (as you have done) by replacing the node writer.

------

Tag Helpers also fall into this category of 'fundamental truth' and represent the invocation syntax of an html tag that will **run code** to output html. They are intentionally abstracted from the runtime implementation, and can map to different implementations inside the framework (MVC already does this).

When we get further along (maybe tomorrow) - we'll have derived extension nodes that represent the specifics of components, and probably also a target extension that renders them to c#. Example - all of the VCTH stuff [here](https://github.com/aspnet/Razor/tree/dev/src/Microsoft.AspNetCore.Mvc.Razor.Extensions)

Using a target extension and defining lots of nodes is a little more ceremony, but provides some design benefits when doing something complex. In particular it's helpful when you need to separate editor codegen from runtime code gen.

-----

The third category of stuff is extension nodes that represent framework-specific concepts (such as `@page` from MVC).

It always a cost benefit choice of whether to implement a target extension for these things.  We made the tradeoff in terms of implementing a target extension for most cases, but I can't say it's always paid off. It can always be added later.

I was bored while making coffee 😆

SteveSandersonMS: OK, thanks for the info!

PR #152, opened 2018-02-22
Adopt more of Razor Exensibility

rynowak:
Removes some workarounds and uses Razor extensibility in a few more
places.

_ViewImports now works in VS


src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.csproj
```diff
@@ -19,6 +19,14 @@
     <None Remove="tools\**" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- This is a newer version than is in the shared framework, so copy it -->
+    <Content Include="..\Microsoft.AspNetCore.Blazor.Razor.Extensions\bin\$(Configuration)\netstandard2.0\Microsoft.AspNetCore.Razor.Language.dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Visible>false</Visible>
+    </Content>
+  </ItemGroup>
```
rynowak: This is the kind of thing that pops up when you're using the build output to create a CLI tool. Any package reference you have that diverges from the shared framework won't be around for you to use. 

Technically the output of dotnet build is not designed to be portable, that's supposed to require a publish. In the long run for Blazor this shouldn't matter if we can just use the Razor build server.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorApi.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Blazor.Razor
+{
+    // Constants for method names used in code-generation
+    // Keep these in sync with the actual definitions
+    internal static class BlazorApi
```
rynowak: I decided to group these because otherwise they are scattered across a bunch of different files.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorCodeTarget.cs
```diff
@@ -11,11 +13,47 @@ namespace Microsoft.AspNetCore.Blazor.Razor
...
+            Extensions = extensions.ToArray();
+        }
+
+        public ICodeTargetExtension[] Extensions { get; }
+
         public override IntermediateNodeWriter CreateNodeWriter()
-            => new BlazorIntermediateNodeWriter();
+        {
+            return _options.DesignTime ? (IntermediateNodeWriter)new DesignTimeNodeWriter() : new BlazorIntermediateNodeWriter();
+        }
```
rynowak: We need to properly implement codetarget for tooling. Looking for a way to make this not needed, but it seems OK for now.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorCodeTarget.cs
```diff
@@ -11,11 +13,47 @@ namespace Microsoft.AspNetCore.Blazor.Razor
...
+            _options = options;
+            Extensions = extensions.ToArray();
+        }
+
+        public ICodeTargetExtension[] Extensions { get; }
+
         public override IntermediateNodeWriter CreateNodeWriter()
-            => new BlazorIntermediateNodeWriter();
+        {
+            return _options.DesignTime ? (IntermediateNodeWriter)new DesignTimeNodeWriter() : new BlazorIntermediateNodeWriter();
```
rynowak: Using the stock `DesignTimeNodeWriter` in VS for now. So far I'm not aware of any problems.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.AspNetCore.Blazor.Razor
+{
+    public static class BlazorConfiguration 
+    {
+        public static readonly RazorConfiguration Default = new RazorConfiguration(
+            RazorLanguageVersion.Version_2_1,
+            "Blazor-0.1",
+            Array.Empty<RazorExtension>());
```
rynowak: This should be the way we plug in our extensibility, need to enable a few things on the Razor side first.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
```diff
@@ -0,0 +1,81 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Extensions;
+
+namespace Microsoft.AspNetCore.Blazor.Razor
+{
+    public class BlazorExtensionInitializer : RazorExtensionInitializer
```
rynowak: This is the recipe for creating a razor 'plugin'.

I grouped a bunch of the old configuration code here for now for ease of maintenance.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
```diff
@@ -0,0 +1,81 @@
...
+            public void Configure(RazorCodeGenerationOptionsBuilder options)
+            {
+                if (options == null)
+                {
+                    throw new ArgumentNullException(nameof(options));
+                }
+
+                // These metadata attributes require a reference to the Razor.Runtime package which we don't
+                // otherwise need.
+                options.SuppressMetadataAttributes = true;
```
rynowak: This configures the metadata that MVC uses the find views/pages at runtime. If we need something like this it's fairly easy to build a blazor-specific version.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorImportProjectFeature.cs
```diff
@@ -0,0 +1,79 @@
...
+            private readonly byte[] _defaultImportBytes;
+
+            private VirtualProjectItem()
+            {
+                var preamble = Encoding.UTF8.GetPreamble();
+                var content = @"
+@using System
+@using System.Collections.Generic
+@using System.Linq
+@using System.Threading.Tasks
```
rynowak: Let me know if you want me to add anything else here.

rynowak: This can have any non-block directives - ie: no code, only declarations.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorLoweringPhase.cs
```diff
@@ -1,75 +0,0 @@
...
-using Microsoft.AspNetCore.Razor.Language.CodeGeneration;
-using Microsoft.AspNetCore.Razor.Language.Intermediate;
-
-namespace Microsoft.AspNetCore.Blazor.Razor
-{
-    /// <summary>
-    /// A <see cref="RazorEngine"/> phase that builds the C# document corresponding to
-    /// a <see cref="RazorCodeDocument"/> for a Blazor component.
-    /// </summary>
-    internal class BlazorLoweringPhase : IRazorCSharpLoweringPhase
```
rynowak: Replaced by the document classifier.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -0,0 +1,99 @@
...
+            return true;
+        }
+
+        protected override void OnDocumentStructureCreated(
+            RazorCodeDocument codeDocument, 
+            NamespaceDeclarationIntermediateNode @namespace, 
+            ClassDeclarationIntermediateNode @class, 
+            MethodDeclarationIntermediateNode method)
+        {
+            @namespace.Content = (string)codeDocument.Items[BlazorCodeDocItems.Namespace];
```
rynowak: We will want to replace stuff like this asap. We don't have a great solution for this yet in VS. I have a few ideas.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            @class.Modifiers.Clear();
+            @class.Modifiers.Add("public");
+
+            // Note that DefaultDocumentWriter's VisitMethodDeclaration in 2.0 is hardcoded to
+            // emit methods with no parameters.
+            //
+            // This was added in 2.1, but we still need a workaround to support 2.0. In that case we
+            // inject the parameter later in RazorCompiler.
```
rynowak: Will remove this comment, I already addressed that.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -0,0 +1,99 @@
...
+            var callBase = new CSharpCodeIntermediateNode();
+            callBase.Children.Add(new IntermediateToken
+            {
+                Kind = TokenKind.CSharp,
+                Content = $"base.{BlazorApi.BlazorComponent.BuildRenderTree}(builder);" + Environment.NewLine
+            });
+            method.Children.Insert(0, callBase);
+        }
+
+        #region Workaround
```
rynowak: This is simplest, most isolated workaround I could find for not being able to replace the intermediate node writer.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Blazor.Razor;
+using Microsoft.AspNetCore.Razor.Language;
+
+[assembly: ProvideRazorExtensionInitializer("Blazor-0.1", typeof(BlazorExtensionInitializer))]
```
rynowak: This hooks up the razor tooling's plugin discovery

tooling/Microsoft.VisualStudio.LanguageServices.Blazor/BlazorProjectEngineFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System;
+using Microsoft.AspNetCore.Blazor.Razor;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.VisualStudio.LanguageServices.Blazor
+{
+    [ExportCustomProjectEngineFactory("Blazor-0.1", SupportsSerialization = false)]
+    internal class BlazorProjectEngineFactory : IProjectEngineFactory
```
rynowak: This hooks up VS.

src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.csproj
SteveSandersonMS: Glad you're anticipating that sort of thing!

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorApi.cs
SteveSandersonMS: OK sounds good. BTW would anything bad happen if `Microsoft.AspNetCore.Blazor.Razor.Extensions` took a dependency on `Microsoft.AspNetCore.Blazor` (which is just a netstandard2.0 package with no dependencies)? If we did that then we could use a lot of `nameof` and `typeof(...).FullName` instead of duplicating. I'm guessing you probably didn't for an important reason, but just checking.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        public static void Register(RazorProjectEngineBuilder builder)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            FunctionsDirective.Register(builder);
+            InheritsDirective.Register(builder);
```
SteveSandersonMS: If #151 gets merged first, please add `InjectDirective.Register(builder);` here

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
```diff
@@ -0,0 +1,81 @@
...
+        // we need into the RazorProjectEngine (namespace).
+        public static void Register(IRazorEngineBuilder builder)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            FunctionsDirective.Register(builder);
+            InheritsDirective.Register(builder);
```
SteveSandersonMS: (and here)

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorImportProjectFeature.cs
SteveSandersonMS: Great. I'm sure we will want a couple of the most commonly used Blazor namespaces here too, but we can work out exactly which ones later.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -0,0 +1,94 @@
...
+            });
+
+            // We need to call the 'base' method as the first statement.
+            var callBase = new CSharpCodeIntermediateNode();
+            callBase.Children.Add(new IntermediateToken
+            {
+                Kind = TokenKind.CSharp,
+                Content = $"base.{BlazorApi.BlazorComponent.BuildRenderTree}(builder);" + Environment.NewLine
+            });
+            method.Children.Insert(0, callBase);
```
SteveSandersonMS: We can probably remove this `base.blah()` call soon. I'll change the base class so that it doesn't need you to call the base implementation of the thing it wants you to override. But for now this is still needed.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorImportProjectFeature.cs
rynowak: BTW the strategy we take in MVC to dump all of the names applications are likely to use into 1-2 namespaces so that the strategy is straightforward. For MVC views that's `M.A.Mvc` and `M.A.Mvc.Rendering`.

Also, fully qualify everything we use as a codegen target, so that there's no potential ambiguity.

Mentioning it because I've seen 3-4 different namespaces in tests.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
rynowak: Cool beans!

PR #155, opened 2018-02-22
Implement layout and implements with directives

rynowak:
- Remove haxxxx
- Add proper directives with tooling support


rynowak:
A few previews:
image

image

test/Microsoft.AspNetCore.Blazor.Build.Test/RazorCompilerTest.cs
```diff
@@ -525,23 +525,23 @@ public void SupportsLayoutDeclarationsViaTemporarySyntax()
             Assert.NotNull(layoutAttribute);
             Assert.Equal(typeof(TestLayout), layoutAttribute.LayoutType);
             Assert.Collection(frames,
-                frame => AssertFrame.Text(frame, "\nHello"));
+                frame => AssertFrame.Text(frame, "Hello"));
```
rynowak: The extra leading newline looked to me like it's an artifact of your workaround for not having proper directive support. If this isn't the case, let me know and I'll look into it.

SteveSandersonMS: Exactly. Nobody really wants spurious extra newlines :)

SteveSandersonMS:
Those screenshots look so good!

SteveSandersonMS:
This looks excellent, and I'm really looking forwards to having this in.

If we want our first release (in ~2 weeks) to be compatible with VS15.5/15.6, then do we need to be putting these tooling changes in a separate branch? I'm guessing that although it would work fine at compile time in 15.5/6. you'd see design-time errors because it doesn't know the new syntax. Is that correct?

rynowak:
> I'm guessing that although it would work fine at compile time in 15.5/6. you'd see design-time errors because it doesn't know the new syntax. Is that correct?

Yeah that's correct. 

> Thoroughly approved. But do we need to merge it into a post-0.0.5 branch so that 0.0.5 retains compatibility with VS 15.5/6?

I see what you mean. I'll do that for now. 

The other option is to have folks install a private Razor VSIX, which I don't think is a great option for obvious reasons.

I'm also going to add instructions to the readme for how to do that. The current Razor bits that are going into 15.7 are **compatible** with 15.6 so you can always set yourself up to do it if you know what to install.

rynowak:
>Thoroughly approved. But do we need to merge it into a post-0.0.5 branch so that 0.0.5 retains compatibility with VS 15.5/6?

Another thought that I had, we could also do the reverse. Move the 0.0.1 work into `release/0.0.1` and keep dev dirty. I think the main pivot is where we expect most changes to land. 

If dev is the branch for 0.0.1 it means that there will be a 'big merge' in the future as all of the 0.0.5 work will need to join back after 0.0.1 ships.

If dev is the primary branch for 0.0.5 then it means that changes go into `release/0.0.1` and then can be immediately merged back to dev and adapted to the 'future' shape.

You're doing the majority of the 0.1 work so I'll defer to you.

SteveSandersonMS:
I expect there will be a lot of commits over the next week in whichever branch represents 0.0.1. After that, 0.0.1 should be final.

If your 0.0.5 work was kept in a separate branch, I'd be happy to keep rebasing it on 0.0.1 once a day or so, so it never gets far off (if you don't mind the force-pushes to 0.0.5 after each daily rebase, or if you have a different preferred Git strategy for feature branches).

rynowak:
Alright that sounds fine. I'm going to create `feature/0.0.5` and then once that 0.0.1 ships we can merge everything to `dev`

rynowak:
Merged. feature/0.0.5 has all of this stuff now.

PR #156, opened 2018-02-23
Initial HttpClient

SteveSandersonMS:
We don't know how soon Mono will be ready to support the real `System.Net.Http.HttpClient`. Hopefully it will be very soon, but in case it's not soon enough for our 0.0.5 release, this PR is to implement a sufficiently similar `HttpClient` that people can use in the meantime. It has `Temporary` in its namespace, but otherwise you can use it similarly.

Most of this work isn't wasted, because even when Mono does support the real `System.Net.Http.HttpClient`, the bulk of our work to integrate with it will still be the same:

 * Implementing our own browser-compatible HTTP message handler
 * Making use of `HttpClient` in sample apps
 * Providing whatever JSON-related APIs are necessary to do anything useful with HTTP requests
 * E2E tests around all that

So even when we switch to the real `HttpClient`, we'll still be using most of the code in this PR.


src/Microsoft.AspNetCore.Blazor.Browser/Services/Temporary/HttpClient.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// an instance of <see cref="HttpResponseMessage"/> in an asynchronous
+        /// operation.
+        /// </summary>
+        /// <param name="requestUri">The URI the request is sent to.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public Task<HttpResponseMessage> GetAsync(string requestUri)
+        {
+            var tcs = new TaskCompletionSource<HttpResponseMessage>();
+            int id;
+            lock (_idLock)
```
TylerBrinkley: @SteveSandersonMS are thread synchronization locks necessary in the browser? Isn't WebAssembly single threaded like JavaScript?

SteveSandersonMS: https://gist.github.com/SteveSandersonMS/ba34011dd4ee8bb3021c8837ff68ef28#gistcomment-2360374

grahamehorner: @SteveSandersonMS looking good 👍 is the plan to move the interop into Typescript for everything or is it to be mixed? vanilla JavaScript/Typescript etc.  also what would be your thoughts around using RxJs in the interop layer for async/event based operations?

PR #166, opened 2018-02-26
Rynowak/vsix improvements

rynowak:
These changes should make us 15.6-oriented

### DIFF
--- a/tooling/Microsoft.VisualStudio.BlazorExtension/Microsoft.VisualStudio.BlazorExtension.csproj
+++ b/tooling/Microsoft.VisualStudio.BlazorExtension/Microsoft.VisualStudio.BlazorExtension.csproj
@@ -22,7 +22,7 @@
 
   <PropertyGroup>
     <!-- Following VS convention of using the VS release # as a convention for the vsix version. -->
-    <VsixVersion>15.7</VsixVersion>
+    <VsixVersion>15.6</VsixVersion>
     <VsixVersion Condition="'$(BuildNumber)'!=''">$(VsixVersion).$(BuildNumber)</VsixVersion>
     <VsixVersion Condition="'$(BuildNumber)'==''">$(VsixVersion).999999</VsixVersion>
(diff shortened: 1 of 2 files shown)


tooling/Microsoft.VisualStudio.BlazorExtension/Microsoft.VisualStudio.BlazorExtension.csproj
```diff
@@ -99,18 +99,6 @@
       this problem, we're only using the P2P references to force build ordering. Then we include the build
       output of those projects as content.
     -->
-    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Blazor.Razor.Extensions\Microsoft.AspNetCore.Blazor.Razor.Extensions.csproj">
-      <Name>Microsoft.AspNetCore.Blazor.Razor.Extensions</Name>
-      <Private>False</Private>
-      <IncludeOutputGroupsInVSIX></IncludeOutputGroupsInVSIX>
-      <IncludeOutputGroupsInVSIXLocalOnly></IncludeOutputGroupsInVSIXLocalOnly>
-    </ProjectReference>
-    <Content Include="..\..\src\Microsoft.AspNetCore.Blazor.Razor.Extensions\bin\$(Configuration)\net461\Microsoft.AspNetCore.Blazor.Razor.Extensions.dll">
```
alexgav: Could we leave this one in? If I have time to do the in-product survey this week, it probably would be better to put it into the VSIX assembly rather than language service assembly. I think currently it doesn't have much other than a VSPackage (and I would need a package for the in-product survey).

rynowak: `Microsoft.AspNetCore.Blazor.Razor.Extensions` is an assembly that contains Blazor's customizations for Razor codegen, it's not the thing you're worried about 👍

PR #182, opened 2018-02-28
[WIP] Component discovery

rynowak:
Just a sneak preview....

This uses the Razor SDK stuff to implement 2-phase compilation and component/tag helper discovery. 

Example component data here https://gist.github.com/rynowak/5e6f588b41afb6161c9b74e3d3838de0


samples/HostedInAspNet.Client/HostedInAspNet.Client.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Razor/2.1.0-preview2-30159">
```
rynowak: This is the way to bring in an SDK from a .nupkg.

Thoughts on this as a temporary solution? This requires a 2.1 CLI, but not a particular one as long as it's as new as preview1 (I think)

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/BuildRazorCommand.cs
```diff
@@ -1,85 +0,0 @@
...
-using Microsoft.AspNetCore.Blazor.Razor;
-using Microsoft.Extensions.CommandLineUtils;
-using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-
-namespace Microsoft.AspNetCore.Blazor.Build.Cli.Commands
-{
-    internal class BuildRazorCommand
```
rynowak: These are not needed anymore. These are pretty similar to what's in the Razor SDK

src/Microsoft.AspNetCore.Blazor.Build/Core/RazorCompilation/RazorCompiler.cs
```diff
@@ -25,21 +27,23 @@ public class RazorCompiler
         /// </summary>
         /// <param name="inputRootPath">Path to a directory containing input files.</param>
         /// <param name="inputPaths">Paths to the input files relative to <paramref name="inputRootPath"/>. The generated namespaces will be based on these relative paths.</param>
+        /// <param name="referencePaths">Paths to reference assemblies. Used to discovery components.</param>
```
rynowak: We should talk about this class. I want to see if we can just do without it.

src/Microsoft.AspNetCore.Blazor.Build/ReferenceFromSource.props
```diff
@@ -10,6 +10,10 @@
   
   <Import Project="$(MSBuildThisFileDirectory)targets\All.targets" />
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="$(RazorPackageVersion)"/>
+  </ItemGroup>
```
rynowak: This package delivers a specific version of the razor toolset. Whichever is the 'root' package for Blazor client code should include this and also include its MSBuild assets.

rynowak: You need this **AND** the Razor SDK

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -1,18 +1,27 @@
...
+    public class ComponentDocumentClassifierPass : DocumentClassifierPassBase, IRazorDocumentClassifierPass
     {
-        protected override string DocumentKind => "Blazor.Component-0.1";
+        public static readonly string ComponentDocumentKind = "Blazor.Component-0.0.5";
+
+        private static readonly char[] PathSeparators = new char[] { '/', '\\' };
+
+        // This will be set by our CLI tool when running at the command line. The default value
+        // will be used by the IDE.
+        public string BaseNamespace { get; set; } = "HostedInAspNet.Client"; //"__BlazorGenerated";
```
rynowak: This is the one sticky thing I ran into. Razor doesn't provide the base namespace today and it probably should.

I think it makes more sense to try and work around this rather than blocking on it. Let's discuss at the sync.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -70,6 +76,38 @@ protected override bool IsMatch(RazorCodeDocument codeDocument, DocumentIntermed
...
+            for (var i = 0; i < segments.Length - 1; i++)
+            {
+                builder.Append('.');
+                builder.Append(CSharpIdentifier.SanitizeClassName(segments[i]));
+            }
+
+            @namespace = builder.ToString();
+            @class = Path.GetFileNameWithoutExtension(relativePath);
+
+            return true;
```
rynowak: This is very similar to your existing code, and MVC's implementation of `@namespace`

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,247 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.AspNetCore.Blazor.Razor
+{
+    internal class ComponentTagHelperDescriptorProvider : RazorEngineFeatureBase, ITagHelperDescriptorProvider
```
rynowak: I kept this as simple as possible for now. We should dig in more once the E2E is done.

test/Microsoft.AspNetCore.Blazor.Build.Test/RazorCompilationIntegrationTest.cs
```diff
@@ -10,18 +17,12 @@
...
-using System.Runtime.InteropServices;
-using System.Text;
+using Microsoft.Extensions.DependencyModel;
 using Xunit;
 
-namespace Microsoft.AspNetCore.Blazor.Build.Test
+namespace Microsoft.AspNetCore.Blazor.Build
 {
-    public class RazorCompilerTest
+    public class RazorCompilationIntegrationTest
```
rynowak: This part is all WIP, ignore the rest. I broke the world 👍

PR #189, opened 2018-02-28
Add BlazorTooling.sln to instructions

rynowak:


README.md
```diff
@@ -60,6 +60,16 @@ Install prerequisites for E2E tests
...
 
+## Build VS Tooling
+
+Prerequisites:
+- Visual Studio 2017 15.5 (or newer)
+- Visual Studio extension development features (install via Visual Studio Installer)
+
+Open a VS Developer Command Prompt
+
+Run `msbuild BlazorTooling.sln` from the solution directory.
```
rynowak: It doesn't seem useful to include instructions about how to debug the tools until there's something to actually debug, so I left it out 😆

README.md
```diff
@@ -60,6 +60,16 @@ Install prerequisites for E2E tests
 
 Run `dotnet test test\AllTests.proj`
 
+## Build VS Tooling
+
+Prerequisites:
+- Visual Studio 2017 15.5 (or newer)
```
rynowak: I would imagine that it's possible to build with earlier VS builds, but anyone with old VS installs is going to have a bad time for other reasons, so I'd rather say it's not supported.

PR #192, opened 2018-03-01
Implement basic Component discovery

rynowak:
This is the basics of component discovery along with some tests.

The next set of changes will integrate it into the compilation process.

### DIFF
--- a/test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/ComponentTagHelperDescriptorProviderTest.cs
+++ b/test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/ComponentTagHelperDescriptorProviderTest.cs
@@ -1,0 +1,295 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.Extensions.DependencyModel;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Blazor.Razor.Extensions
+{
+    public class ComponentTagHelperDescriptorProviderTest
+    {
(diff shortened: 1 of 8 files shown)


test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/ComponentTagHelperDescriptorProviderTest.cs
```diff
@@ -0,0 +1,295 @@
...
+            // Act
+            provider.Execute(context);
+
+            // Assert
+            var components = ExcludeBuiltInComponents(context);
+            var component = Assert.Single(components);
+
+            // These are features Components don't use. Verifying them once here and
+            // then ignoring them.
+            Assert.Empty(component.AllowedChildTags);
```
SteveSandersonMS: Does this being empty mean that all child markup is allowed, or that none is? We'll want to allow any markup inside a component, e.g., for cases like

    <MyDialog>
        Some arbitrary HTML goes here, possibly containing more component tag helpers
    </MyDialog>

test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/ComponentTagHelperDescriptorProviderTest.cs
```diff
@@ -0,0 +1,295 @@
...
+            Assert.Equal("System.Boolean", attribute.TypeName);
+
+            Assert.False(attribute.HasIndexer);
+            Assert.True(attribute.IsBooleanProperty);
+            Assert.False(attribute.IsEnum);
+            Assert.False(attribute.IsStringProperty);
+        }
+
+        [Fact] // enum properties have some special intellisense behavior
+        public void Excecute_EnumProperty_CreatesDescriptor()
```
SteveSandersonMS: Not necessary to fix, but typo in test name

SteveSandersonMS:
@rynowak Looks good!

test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/ComponentTagHelperDescriptorProviderTest.cs
rynowak: All properly nested markup is allowed

PR #196, opened 2018-03-01
remove mpf

LunicLynx:
MPF seems like an unnecessary dependency. While it is present you cannot install the vsix on its own.


rynowak:
> While it is present you cannot install the vsix on its own.

Can you provide a little more information about this? We expect this extension to be installable and work on VS 15.6+ with the Web Workload installed (includes C#, web tools, and Razor).

LunicLynx:
Sure, this is the error i get when it is included:
image

And it does work without that dependency. I also found other extension online with the same issue, which had to remove this dependency as well.

rynowak:
> But maybe i'm missing something ...

Nah, I think I must be the one who's missing something 😆 - if it doesn't work then it doesn't work. Your configuration is supposed to be supported. 

I think the problem might be the 15.6 instead of 15.0. I had intended to enforce the 15.6 version requirement, but perhaps that package isn't versioned.

I'll merge your PR right now to unblock folks and see what I can do to get to the bottom of this. Thanks for letting us know about this!

PR #199, opened 2018-03-01
Implement 2-phase compilation

rynowak:
This is a working (all tests passing) implementation of the two-phase
compilation system we will need for component discovery.

This builds on top of work we've doing in Razor, including the Razor
SDK, MSBuild tasks, and CLI/server.

This currently *does* discovery components during the build process, but
it doesn't use that data for anything yet.

It works like this:
1. Generate class declarations (structure only, no method bodies)
2. Compile a 'temp' assembly using the .cs files and output of 1.
3. Do component discovery using the 'temp' assembly
4. Generate class definitions (including method bodies)
5. Compile the 'real' assembly using the .cs files and output of 4.


samples/HostedInAspNet.Client/HostedInAspNet.Client.csproj
```diff
@@ -1,7 +1,8 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Razor/2.1.0-preview2-30159">
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <OutputType>Exe</OutputType>
```
rynowak: /cc @SteveSandersonMS  currently the 'build' command expects that this project will be an exe (`GetEntryPoint`) not a library. WebSDK forces all projects to be an exe, which is the reason for the delta.

Any thoughts on this? bug or intentional? I can investigate if needed

SteveSandersonMS: Correct, the assembly does need to indicate its entrypoint.

> Any thoughts on this? bug or intentional? I can investigate if needed

Sorry, I don't follow - what might be a bug? The absence of "<OutputType>exe</OutputType>" before? It was only omitted because it was redundant (like you say, the Web SDK put it there by default).

If we're changing to the Razor SDK - which I agree we should do for now until we have a Blazor SDK that extends the Razor one - then it looks totally reasonable to put `<OutputType>Exe</OutputType>` back.

Or am I missing the question?

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/BuildRazorCommand.cs
```diff
@@ -1,96 +0,0 @@
...
-            var violations = options.Where(o => !o.HasValue()).ToList();
-            foreach (var violation in violations)
-            {
-                Console.WriteLine($"ERROR: No value specified for required option '{violation.LongName}'.");
-            }
-
-            return !violations.Any();
-        }
-    }
-}
```
SteveSandersonMS: Really cool that we don't need any of this any more :)

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -105,7 +105,7 @@ private bool TryComputeNamespaceAndClass(string filePath, string relativePath, o
 
             var builder = new StringBuilder();
             builder.Append(baseNamespace); // Don't sanitize, we expect it to contain dots.
-
+
```
SteveSandersonMS: Typo

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/EliminateMethodBodyPass.cs
```diff
@@ -25,7 +25,42 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
...
+                    new IntermediateToken()
+                    {
+                        Content = "#pragma warning disable 0649" + Environment.NewLine,
+                        Kind = TokenKind.CSharp,
+                    },
+
+                    // Field is never used
+                    new IntermediateToken()
+                    {
+                        Content = "#pragma warning disable 0169" + Environment.NewLine,
```
SteveSandersonMS: We don't need corresponding `restore` pragmas because these are all separate files now, is that right?

rynowak: The restore parameters are really just about being nice, and we try to put them **around **specific things when we do pragmas.  In this case the code that triggers the warnings could be anywhere in the file, so we can't really scope it down, it has to surround the whole class.

Also these case of these things, they are only for the temp-compile, so if you have code that would cause these warnings legitimately you will still get them when you build.

I also considered turning off warnings globally when invoking CSC - that may be a better approach, but it would have taken me more work to figure out than this took to code up.

samples/HostedInAspNet.Client/HostedInAspNet.Client.csproj
rynowak: No, I think you we're talking about the same thing. Makes sense to me now.

PR #203, opened 2018-03-02
VSIX Changes to have Blazor templates included into VSIX and show up …

alexgav:
…in One ASP.NET Dialog. Changing the template project target to 4.6 to be able to reference it from VSIX (shouldn't be a problem AFAIK, Phil?). Adding template nuget to the VSIX. Adding pkgdef file with a regkey to allow template discovery.

### DIFF
--- a/tooling/Microsoft.VisualStudio.BlazorExtension/Microsoft.VisualStudio.BlazorExtension.csproj
+++ b/tooling/Microsoft.VisualStudio.BlazorExtension/Microsoft.VisualStudio.BlazorExtension.csproj
@@ -19,7 +19,6 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\build\Key.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
-
   <PropertyGroup>
     <!-- 
       Following VS convention of using the VS release # as a convention for the vsix version.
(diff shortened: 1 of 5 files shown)


tooling/Microsoft.VisualStudio.BlazorExtension/Microsoft.VisualStudio.BlazorExtension.csproj
```diff
@@ -103,12 +109,24 @@
       this problem, we're only using the P2P references to force build ordering. Then we include the build
       output of those projects as content.
     -->
+    <Content Include="Templates.pkgdef" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Blazor.Templates\Microsoft.AspNetCore.Blazor.Templates.csproj">
+      <Project>{edd21533-c6e6-4f85-be4f-10e06756e24c}</Project>
+      <Name>Microsoft.AspNetCore.Blazor.Templates</Name>
+      <Private>False</Private>
```
rynowak: If you're still having trouble getting this to build, you could do something like this in the templates project

<PropertyGroup> 
  <TargetFrameworks Condition="'$(OS)'=='Windows_NT'">net46;netstandard2.0</TargetFrameworks>
  <TargetFrameworks Condition="'$(OS)'!='Windows_NT'">netstandard2.0</TargetFrameworks>
</PropertyGroup

alexgav: Thanks Ryan! No, netstandard2.0 fix it. We are  good to go if everyone is good.

PR #209, opened 2018-03-06
Misc build fixes

rynowak:


src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Client/BlazorHosted.CSharp.Client.csproj
```diff
@@ -1,10 +1,12 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Razor/2.1.0-preview2-30159">
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.1.0-preview2-30159" PrivateAssets="all" />
```
rynowak: FYI we will need to add these packages to the Blazor feed before making this live

SteveSandersonMS:
@rynowak Looks great to me!

> FYI we will need to add these packages to the Blazor feed before making this live

Is that because the aspnetcore-dev feed's retention policy means we can't rely on fetching it from there? If so, @Eilon, do we have the ability to upload one-off nupkgs to MyGet feeds directly without it coming from CI?

Eilon:
aspnetcore-dev currently is set to be indefinite retention, but I don't know that we'll keep that because it might not be sustainable. We can certainly upload one-off packages whenever we want.

PR #210, opened 2018-03-06
Implement components as tag helpers

rynowak:
Implements Component code generation and tooling support end to end
udditionally adds some default `@addTagHelper` directives to make
programming in Blazor a little nicer.

Components are discovered as Tag Helpers using Razor's extensibility
during the build/IDE process. This drives the code generation during
build and lights up a bunch of editor features.

Add


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
```diff
@@ -373,6 +363,173 @@ private void WriteElementAttributeToken(CodeRenderingContext context, HtmlTagTok
...
+                node.Children[0] is HtmlContentIntermediateNode htmlNode && htmlNode.Children.Count != 1 ||
+                node.Children[0] is CSharpExpressionIntermediateNode cSharpNode && cSharpNode.Children.Count != 1)
+            {
+                // We don't expect this to happen, we just want to know if it can.
+                throw new InvalidOperationException("Attribute nodes should either be minimized or a single content node.");
+            }
+            else if (node.BoundAttribute.IsUIEventHandlerProperty())
+            {
+                // This is a UIEventHandler property. We do some special code generation for this
+                // case so that it's easier to write for common cases.
```
rynowak: Note, this isn't up to date based on our conversation this morning.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -0,0 +1,499 @@
...
+            // because each component creates a lambda scope for its child content.
+            //
+            // We're hacking it a bit here by just forcing every component to have an empty lambda
+            _scopeStack.OpenScope(node.TagName, isComponent: true);
+            _scopeStack.IncrementCurrentScopeChildCount(context);
+            context.RenderChildren(node);
+            _scopeStack.CloseScope(context, node.TagName, isComponent: true, source: node.Source);
+        }
+
+        public override void WriteComponentAttribute(CodeRenderingContext context, ComponentAttributeExtensionNode node)
```
rynowak: @NTaylorMullen @ajaybhargavb - I feel like I was able to really simplify this code from what's currently the state of the art in Razor. I also used this opportunity to prototype and approach for Razor's design time code that was suggested by @jasonmalinowski to solve our alignment woes in the debugger...

Example:
```
            __o = 
#line 6 "x:\dir\subdir\Test\TestComponent.cshtml"
                    new SomeType()

#line default
#line hidden
            ;
```
Traditionally we'd put all of this junk on one line. This is a problem whenever the design time code gen has a significant offset that differs from what the user typed.

This deffo comes up in Blazor with cases like the following - imagine the user's code is in the `...`.
```
#line 6 "x:\dir\subdir\Test\TestComponent.cshtml"
            __o =  new UIEventHandler(e => ...);
#line default
#line hidden
```

NTaylorMullen: When we initially re-did Razor for .net Core this is what I did. In the end it ended up breaking formatting and a few other pieces that I can't recall. Because of that we ended up reverting the multi-line code gen functionality 😢

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -0,0 +1,499 @@
...
+        }
+
+        public override void BeginWriteAttribute(CodeWriter codeWriter, string key)
+        {
+            codeWriter
+                .WriteStartMethodInvocation($"{_scopeStack.BuilderVarName}.{nameof(BlazorApi.RenderTreeBuilder.AddAttribute)}")
+                .Write("-1")
+                .WriteParameterSeparator()
+                .WriteStringLiteral(key)
+                .WriteParameterSeparator();
```
SteveSandersonMS: Out of interest, why do we need to emit the `AddAttribute` lines at design time?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
SteveSandersonMS: A further update would generalise this to arbitrary delegate types (not just `UIEventHandler`), would it?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
```diff
@@ -373,6 +363,173 @@ private void WriteElementAttributeToken(CodeRenderingContext context, HtmlTagTok
...
+                //
+                // Example: 
+                //      <MyComponent OnClick="Foo()"/> 
+                //      --> builder.AddAttribute(X, "OnClick", new UIEventHandler((e) => Foo()));
+                //
+                // The constructor is important because we want to put type inference into a state where
+                // we know the delegate's type should be UIEventHandler. AddAttribute has an overload that
+                // accepts object, so without the 'new UIEventHandler' things will get ugly.
```
SteveSandersonMS: I know this is a general C# question but after compilation, is `new UIEventHandler((e) => Foo())` equivalent to `((UIEventHandler)((e) => Foo()))`? The extra `new` doesn't cause an extra heap allocation, does it?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
```diff
@@ -373,6 +363,173 @@ private void WriteElementAttributeToken(CodeRenderingContext context, HtmlTagTok
...
+                // We don't expect this to happen, we just want to know if it can.
+                throw new InvalidOperationException("Attribute nodes should either be minimized or a single content node.");
+            }
+            else if (node.BoundAttribute.IsUIEventHandlerProperty())
+            {
+                // This is a UIEventHandler property. We do some special code generation for this
+                // case so that it's easier to write for common cases.
+                //
+                // Example: 
+                //      <MyComponent OnClick="Foo()"/>
```
SteveSandersonMS: This is great BTW

test/Microsoft.AspNetCore.Blazor.Build.Test/ComponentRenderingRazorIntegrationTest.cs
```diff
@@ -2,101 +2,319 @@
...
-            // Arrange/Act
-            var testComponentTypeName = FullTypeName<TestComponent>();
-            var component = CompileToComponent($"<c:{testComponentTypeName} />");
+            // Arrange
+            AdditionalSyntaxTrees.Add(CSharpSyntaxTree.ParseText(@"
+using Microsoft.AspNetCore.Blazor.Components;
+
+namespace Test
+{
+    public class MyComponent : BlazorComponent
```
SteveSandersonMS: Just to check, these tests that use `BlazorComponent`, they'd also work if it inherited `IComponent` directly would they? Or does it depend on something in the `BlazorComponent` base class?

test/Microsoft.AspNetCore.Blazor.Build.Test/RazorIntegrationTestBase.cs
```diff
@@ -56,21 +56,28 @@ static RazorIntegrationTestBase()
...
             AdditionalSyntaxTrees = new List<SyntaxTree>();
+            AdditionalRazorItems = new List<RazorProjectItem>();
+
             Configuration = BlazorExtensionInitializer.DefaultConfiguration;
             FileSystem = new VirtualRazorProjectFileSystem();
             WorkingDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ArbitraryWindowsPath : ArbitraryMacLinuxPath;
 
             DefaultBaseNamespace = "Test"; // Matches the default working directory
             DefaultFileName = "TestComponent.cshtml";
+
```
SteveSandersonMS: Typo

test/Microsoft.AspNetCore.Blazor.Build.Test/RazorIntegrationTestBase.cs
```diff
@@ -92,58 +102,83 @@ internal RazorProjectEngine CreateProjectEngine(RazorConfiguration configuration
...
-        {
-            return CompileToCSharp(WorkingDirectory, cshtmlRelativePath, cshtmlContent);
-        }
-
-        protected CompileToCSharpResult CompileToCSharp(string cshtmlRootPath, string cshtmlRelativePath, string cshtmlContent)
+        internal RazorProjectItem CreateProjectItem(string chtmlRelativePath, string cshtmlContent)
         {
             // FilePaths in Razor are **always** are of the form '/a/b/c.cshtml'
-            var filePath = cshtmlRelativePath.Replace('\\', '/');
+            var filePath = chtmlRelativePath.Replace('\\', '/');
```
SteveSandersonMS: Is this change a typo?

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
```diff
@@ -43,15 +43,14 @@ public void SupportsCSharpExpressions()
...
-                frame => AssertFrame.Text(frame, new object().ToString(), 7),
-                frame => AssertFrame.Whitespace(frame, 8));
+                frame => AssertFrame.Text(frame, "Hello", 0),
+                frame => AssertFrame.Whitespace(frame, 1),
+                frame => AssertFrame.Whitespace(frame, 2), // @((object)null)
+                frame => AssertFrame.Whitespace(frame, 3),
+                frame => AssertFrame.Text(frame, "123", 4),
+                frame => AssertFrame.Whitespace(frame, 5),
+                frame => AssertFrame.Text(frame, new object().ToString(), 6),
+                frame => AssertFrame.Whitespace(frame, 7));
```
SteveSandersonMS: Removing the leading whitespace frame is great. Pretty soon we'll probably want to remove trailing ones too (because I have an important runtime optimisation that is enabled only in the common case where a component renders exactly one top-level node, and trailing whitespace in source docs tends to interfere with that).

If you know an easy way to strip trailing whitespace frames in the Razor compiler please let me know. I was going to implement a compiler pass to do it.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
rynowak: yes, that's the idea

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
rynowak: I left them out because overloading can make things somewhat confusing while you're typing.

In general an expression like `@(foo)` in Razor will be assigned to a variable of type `object` at design time, which really simplifies the space we need to reason about compared to `AddAttribute(...)` which is overloaded.  If we come up with cases where this illusion breaks we can try to figure something out.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
rynowak: I believe they both allocate, but I'm not totally sure.

I know the former does the right thing for type inference on the lambda, which is important for design time but not runtime. I'll look into this when I revisit the delegate to make it better. We can make the runtime code as efficient as possible while keeping design time tailored to the intellisense experience.

test/Microsoft.AspNetCore.Blazor.Build.Test/ComponentRenderingRazorIntegrationTest.cs
rynowak: Yeah, I just did this brevity. The tests focused on component discovery include `IComponent` as well.

test/Microsoft.AspNetCore.Blazor.Build.Test/RazorIntegrationTestBase.cs
rynowak: yeah just a mistake

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
rynowak: Compiler pass is the right way to do it.

PR #224, opened 2018-03-09
Sets up the monolinker to optionally link the apps on build with MSBuild integration

javiercn:


Blazor.sln
```diff
@@ -12,16 +12,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "mono", "mono", "{7B5CAAB1-A
 		src\mono\mono.targets = src\mono\mono.targets
 	EndProjectSection
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Blazor.Mono", "src\Microsoft.AspNetCore.Blazor.Mono\Microsoft.AspNetCore.Blazor.Mono.csproj", "{39FEC72D-AF52-47A3-B63D-7BF0E4335248}"
```
rynowak: Is this destined for `dev` or `feature/0.1.0`?

src/Microsoft.AspNetCore.Blazor.Build/Core/IndexHtmlWriter.cs
```diff
@@ -1,39 +1,40 @@
...
 using AngleSharp.Html;
-using System;
+using AngleSharp.Parser.Html;
+using Mono.Cecil;
 
-namespace Microsoft.AspNetCore.Blazor.Build.Core.FileSystem
+namespace Microsoft.AspNetCore.Blazor.Build
 {
-    internal class IndexHtmlFileProvider : InMemoryFileProvider
+    public class IndexHtmlWriter
```
rynowak: Is there a reason why this isn't a `Task`? I'm not sure that I have the big picture totally about whether or not the tools needs to be standalone or integrated with MSBuild. If the latter, then just writing a task will simplify your life greatly.

src/Microsoft.AspNetCore.Blazor.Build/ReferenceFromSource.props
```diff
@@ -7,9 +8,23 @@
...
+    <MonoBaseClassLibraryPath>$(BlazorMonoRuntimeBasePath)dist/optimized/bcl/</MonoBaseClassLibraryPath>
+    <MonoBaseClassLibraryFacadesPath>$(BlazorMonoRuntimeBasePath)dist/optimized/bcl/Facades/</MonoBaseClassLibraryFacadesPath>
+    <MonoAsmjsRuntimePath>$(BlazorMonoRuntimeBasePath)dist/optimized/asmjs/</MonoAsmjsRuntimePath>
+    <MonoWasmRuntimePath>$(BlazorMonoRuntimeBasePath)dist/optimized/wasm/</MonoWasmRuntimePath>
+    <BlazorJsPath>$(MSBuildThisFileDirectory)../Microsoft.AspNetCore.Blazor.Browser.JS/dist/blazor.js</BlazorJsPath>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)targets\All.props" />
   <Import Project="$(MSBuildThisFileDirectory)targets\All.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)../mono/mono.targets" />
```
rynowak: nit, pick a kind of *slash* and stick to it. I think generally the MSBuild style is to the use backslash.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+      OverwriteReadOnlyFiles="$(OverwriteReadOnlyFiles)"
+      Retries="$(CopyRetryCount)"
+      RetryDelayMilliseconds="$(CopyRetryDelayMilliseconds)"
+      UseHardlinksIfPossible="$(CreateHardLinksForCopyFilesToOutputDirectoryIfPossible)"
+      UseSymboliclinksIfPossible="$(CreateSymbolicLinksForCopyFilesToOutputDirectoryIfPossible)"
+      Condition="'@(BlazorItemSpec)' != '' and '$(CopyBuildOutputToOutputDirectory)' == 'true' and '$(SkipCopyBuildProduct)' != 'true'">    
+    </Copy>
+
+    <ItemGroup>
+      <BlazorOutputs Include="$(ProjectDir)$(OutputPath)dist/**/*" />
```
rynowak: Can we leverage `CopiedFiles`? https://msdn.microsoft.com/en-us/library/3e54c37h.aspx

Generally listing files using `**/*` is for users not for tool authors. There are gotchas here.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+
+  <PropertyGroup>
+    <BlazorBuildDependsOn>
+      GetBlazorBuildConfiguration;
+      ResolveBinaryOutput;
+      BuildBlazorIndexHtml;
+    </BlazorBuildDependsOn>
+  </PropertyGroup>
+
+  <Target Name="BlazorBuild" DependsOnTargets="$(BlazorBuildDependsOn)" AfterTargets="Compile" />
```
rynowak: I'd suggesting finding another name, and dropping the `AfterTargets`. This will already be called  as a dependency of `_BlazorCopyFilesToOutputDirectory`

Blazor.sln
javiercn: Dev then 0.1.0? /cc: @SteveSandersonMS?

src/Microsoft.AspNetCore.Blazor.Build/Core/IndexHtmlWriter.cs
javiercn: @rynowak Steve had all these things in a CLI tool and I want to keep it that way for now as this dll runs in a separate process from MSBuild and we get to have random dependencies. If I were to build an msbuild task for this it would just wrap the executable.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+    <MakeDir Directories="$(BlazorIntermediateOutputPath)" />    
+    
+  </Target>
+
+  <Target Name="ResolveBinaryOutput" DependsOnTargets="CollectLinkerOutputs;GetBaseClassLibrary" />
+
+  <Target Name="CollectLinkerDescriptors"
+          Inputs="@(IntermediateAssembly);@(BlazorLinkerDescriptor);@(ProducedLinkerDescriptor)"
+          Outputs="@(ProducedLinkerDescriptor)"
+          Condition="$(BlazorShouldRunMonoLinker) == 'true' and '@(BlazorLinkerDescriptor)' == ''">
```
rynowak: What's the purpose of `'@(BlazorLinkerDescriptor)' == ''` in the condition? That's seems a little strange

javiercn: That you haven't provided one explicitly

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+      <GeneratedDescriptorLine Include="@(PrepareAssemblyLine->'&lt;assembly fullname=&quot;%(Identity)&quot; /&gt;')" />
+      <GeneratedDescriptorLine Include="&lt;/linker&gt;" />
+    </ItemGroup>
+
+    <WriteLinesToFile
+      Condition = "'@(BlazorLinkerDescriptor)' == ''"
+      Lines = "@(GeneratedDescriptorLine)"
+      File ="$(GeneratedBlazorLinkerDescriptor)" Overwrite="true" />
+    
+    <ItemGroup Condition = "'@(BlazorLinkerDescriptor)' == ''">
```
rynowak: IMO just embrace that condition attributes in MSBuild are going to look message and kill all the extra whitespace.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+  </ItemGroup>
+   
+    <MakeDir Directories="$(BlazorIntermediateOutputPath)" />    
+    
+  </Target>
+
+  <Target Name="ResolveBinaryOutput" DependsOnTargets="CollectLinkerOutputs;GetBaseClassLibrary" />
+
+  <Target Name="CollectLinkerDescriptors"
+          Inputs="@(IntermediateAssembly);@(BlazorLinkerDescriptor);@(ProducedLinkerDescriptor)"
```
rynowak: If you have a list of files of inputs, you may need to have them and store that in a file. This can matter when something is **removed**.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+    </ItemGroup>
+  
+  </Target>    
+    
+  <Target
+    Name="RunMonoLinker"
+    Condition="$(BlazorShouldRunMonoLinker) == 'true'"
+    DependsOnTargets="CollectLinkerDescriptors"
+    Inputs="@(IntermediateAssembly);
+      @(ReferenceCopyLocalPaths->WithMetadataValue('Extension','.dll'));
```
rynowak: This is another case where you may need to have the filenames of the inputs. If you remove a reference that should trigger a re-link

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+        <PrimaryOutput Condition="'%(FileName)' == @(IntermediateAssembly->'%(FileName)')">true</PrimaryOutput>
+      </BlazorItemSpec>    
+      <FileWrites Include="@(BlazorItemSpec->WithMetadataValue('Type','Assembly')->'%(TargetOutputPath)')" />
+    </ItemGroup>
+  
+  </Target>
+
+  <Target
+    Name="ResolveRuntimeDependencies"
+    Condition="'$(BlazorShouldRunMonoLinker)' == 'false'"
```
rynowak: Unless you can guarantee that this property always has the value `true` or `false` then this should be `==''`. There aren't booleans in MSBuild 😆

src/Microsoft.AspNetCore.Blazor.Build/Core/IndexHtmlWriter.cs
rynowak: > If I were to build an msbuild task for this it would just wrap the executable.

We made the decision to do this (wrap the executable) in Razor, and I don't regret it. I think it's fine for now.

Blazor.sln
SteveSandersonMS: I think we're ready to merge `0.1.0` back into `dev`, right @rynowak? I'm not aware we have any remaining reason to keep it separate.

If that's the case then @javiercn it's up to you: would you prefer to merge `0.1.0` into `dev` first and then you rebase onto that, or would you prefer to merge this PR into `dev` first and then you rebase `0.1.0` onto the new `dev`?

samples/HostedInAspNet.Client/HostedInAspNet.Client.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <BlazorLinkOnBuild>true</BlazorLinkOnBuild>
```
SteveSandersonMS: Later on if you don't mind, I'll remove the `<BlazorLinkOnBuild>` override from the samples (so they build quicker in development) and just make the E2E tests pass in an override somehow. Will that be OK?

samples/StandaloneApp/descriptors/StandaloneApp.xml
```diff
@@ -0,0 +1,3 @@
+<linker>
+	<assembly fullname="StandaloneApp" />
```
SteveSandersonMS: I'm curious about why this is necessary and doesn't get generated during the build like it does for other projects.

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/BuildIndexHtmlCommand.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.AspNetCore.Blazor.Build.Cli.Commands
+{
+    internal class BuildIndexHtmlCommand
+    {
+        public static void Command(CommandLineApplication command)
+        {
+            var clientPage = command.Option("-hp|--html-page",
+                "Path to the HTML Page containing the Blazor bootstrap script tag.",
+                CommandOptionType.SingleValue);
```
SteveSandersonMS: Since this tool isn't intended to be invoked by developers directly, we don't need to do nice things like have shorthand aliases for the options. For simplicity I'd prefer not to have extra ways of doing the same thing.

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/ResolveRuntimeDependenciesCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.AspNetCore.Blazor.Build.Cli.Commands
+{
+    class ResolveRuntimeDependenciesCommand
+    {
+        public static void Command(CommandLineApplication command)
+        {
+            var references = command.Option("--reference",
+                "Full path to the application dependencies that are not part of netstandard2.0",
```
SteveSandersonMS: That's a bit overly specific about `netstandard2.0` (which will change, and arguably no dependency except `netstandard.dll` is part of `netstandard2.0` anyway). Can this just say `Full path to a referenced assembly file`?

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/ResolveRuntimeDependenciesCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+                CommandOptionType.MultipleValue);
+
+            var baseClassLibrary = command.Option("--base-class-library",
+                "Full path to the directories with the dlls that make up the mono BCL",
```
SteveSandersonMS: Although we're using Mono, that's an implementation detail. Could this say `Full path to a directory in which BCL assemblies can be found`?

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/ResolveRuntimeDependenciesCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+                CommandOptionType.MultipleValue);
+
+            var resolutionOverrides = command.Option(
```
SteveSandersonMS: I can't see any indication that this is used anywhere (I mean, the value gets passed through the code, but it's never used to affect the output). Can this be removed? Or am I missing something?

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/ResolveRuntimeDependenciesCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+                "--resolution-override",
+                "Semi-colon separated value pair of assembly name, path",
+                CommandOptionType.MultipleValue);
+
+            var outputPath = command.Option("--output",
+                "Path to the output file that will contain the selected assemblies",
```
SteveSandersonMS: Output *file*? Should this be output directory?

src/Microsoft.AspNetCore.Blazor.Build/Core/IndexHtmlWriter.cs
SteveSandersonMS: Looks good to me.

src/Microsoft.AspNetCore.Blazor.Build/Core/RuntimeDependenciesResolver.cs
```diff
@@ -0,0 +1,140 @@
...
+                    if (Assembly.Definition.Name.Name == current.Name)
+                    {
+                        return Assembly;
+                    }
+
+                    var referencedAssemblyCandidate = FindCandidate(current, Dependencies);
+                    var bclAssemblyCandidate = FindCandidate(current, Bcl);
+
+                    // Resolution logic. For right now, we will prefer the mono BCL version of a given
+                    // assembly if there is a candidate assembly and an equivalent mono assembly.
```
SteveSandersonMS: Sounds good!

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+    dist/Index.html <- This is the optional Index.html file generated from wwwroot/Index.html in case it's present.
+    
+    This task also defines some intermediate paths that we will use:
+    /obj/debug/netstandard2.0/blazor/blazor/linker <- This will be used to create the output from the linker.
+    /obj/debug/netstandard2.0/blazor/blazor/linker.marker.txt <- This will be used to save the output files from
+    the linker and use that as marker to identify whether or not we need to run the linker.
+    -->
+
+    <PropertyGroup Label="Build properties">
+      <BlazorShouldRunMonoLinker>$(BlazorLinkOnBuild)</BlazorShouldRunMonoLinker>
```
SteveSandersonMS: Is there ever a scenario where BlazorShouldRunMonoLinker != BlazorLinkOnBuild? In other words, why have a separate property?

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -0,0 +1,305 @@
...
+  <Target
+    Name="GetBaseClassLibrary"
+    DependsOnTargets="ResolveRuntimeDependencies"
+    Condition="'$(BlazorShouldRunMonoLinker)' == 'false'">
+
+    <!--
+    At this point we have decided not to run the linker and instead to just copy the BCL from
+    the nuget package into the _framework/_bin folder. The only thing we need to do here is
+    collect the list of items that will go into _framework/_bin. This list is composed of:
+    * All the mono BCL files.
```
SteveSandersonMS: That's not the case any more, is it?

src/mono/mono.nuspec
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
SteveSandersonMS: I'm a bit confused about why we have this extra `nuspec` file. Are we publishing this as an extra package? Don't all the assets also get bundled into `Microsoft.AspNetCore.Blazor.Build` due to your other changes in this PR?

test/Microsoft.AspNetCore.Blazor.Build.Test/ReferencedAssemblyFileProviderTest.cs
```diff
@@ -1,128 +0,0 @@
...
-                item => { Assert.Equal("/Mono.Security.dll", item.PhysicalPath); },
-                item => { Assert.Equal("/mscorlib.dll", item.PhysicalPath); },
-                item => { Assert.Equal("/System.Core.dll", item.PhysicalPath); },
-                item => { Assert.Equal("/System.dll", item.PhysicalPath); },
-                item => { Assert.Equal("/System.Linq.Expressions.dll", item.PhysicalPath); },
-                item => { Assert.Equal("/System.Xml.dll", item.PhysicalPath); });
-        }
-
-        [Fact]
-        public void FindsReferencedAssemblyGraphRealistic()
```
SteveSandersonMS: What's the equivalent to this test for the new implementation of dependency graph walking?

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: Publish

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: Yeah, I need to do a bit of cleanup, but I'm trying to get this in and do that later

src/mono/mono.nuspec
javiercn: nah, I just need to kill this

test/Microsoft.AspNetCore.Blazor.Build.Test/ReferencedAssemblyFileProviderTest.cs
javiercn: Fair point, I'll bring the test back

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/BuildIndexHtmlCommand.cs
javiercn: Sure

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/ResolveRuntimeDependenciesCommand.cs
javiercn: This was before our talk about us mimicking the code the linker uses to pick the final list of assemblies. The idea was to allow specifying overrides to the algorithm. I think we won't need it, so I'll take it out.

src/Microsoft.AspNetCore.Blazor.Build/Cli/Commands/ResolveRuntimeDependenciesCommand.cs
javiercn: The language is not super clear. Is actually a file containing the list of paths resulting from resolving the dependencies. Later on the build reads it and uses it to copy things to the right places.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: This is just for verbose reporting only. I've changed it to use a transform. CopiedFiles wouldn't work due to target batching

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: I've renamed it to PrepareBlazorOutput and removed the after targets

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: Essentially if you give me a descriptor file I'll use it, if not I'll generate one to inform the linker to preserve all the code in the main app assembly.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: Not sure what that means if you mean the whitespace between Condition and = it'salready removed

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: I've implemented incremental compilation properly to fix this

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: All this is removed

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
javiercn: I do understand, I've changed it. But using '' and not '' also has its caveheats. Its easy enough to guarantee by just initializing '' to false

samples/HostedInAspNet.Client/HostedInAspNet.Client.csproj
javiercn: Yes. We can do that by passing /p:BlazorLinkOnBuild=true somewhere in the tests I guess. For MonoSanity we need to use TreatPropertyAsLocal to prevent /p:BlazorLinkOnBuild=true from ever overriding

javiercn:
Merged into 0.1.0 59635331601dc8496c7cd5958552dec24d1b0ffd

PR #227, opened 2018-03-09
Support non-property attributes for components

rynowak:


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -423,7 +423,7 @@ public override void WriteComponentAttribute(CodeRenderingContext context, Compo
                 // We don't expect this to happen, we just want to know if it can.
                 throw new InvalidOperationException("Attribute nodes should either be minimized or a single content node.");
             }
-            else if (node.BoundAttribute.IsUIEventHandlerProperty())
+            else if (node.BoundAttribute?.IsUIEventHandlerProperty() ?? false)
```
rynowak: The only change to the codegen is to ignore the case where we don't have a matching property. All of the complexity here was in the lowering part.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
```diff
@@ -69,23 +69,10 @@ public static RazorDiagnostic Create_MultipleComponents(SourceSpan? span, string
...
-
-        public static readonly RazorDiagnosticDescriptor UnboundComponentAttribute = new RazorDiagnosticDescriptor(
-            "BL9987",
-            () => "The component '{0}' does not have an attribute named '{1}'.",
-            RazorDiagnosticSeverity.Error);
-
-        public static RazorDiagnostic Create_UnboundComponentAttribute(SourceSpan? source, string componentType, TagHelperHtmlAttributeIntermediateNode node)
-        {
-            return RazorDiagnostic.Create(UnboundComponentAttribute, source ?? SourceSpan.Undefined, componentType, node.AttributeName);
+            var content = string.Join("", node.FindDescendantNodes<IntermediateToken>().Select(t => t.Content));
```
rynowak: Simplified and handles more cases

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentAttributeExtensionNode.cs
```diff
@@ -47,10 +69,6 @@ public ComponentAttributeExtensionNode(TagHelperPropertyIntermediateNode propert
 
         public BoundAttributeDescriptor BoundAttribute { get; set; }
 
-        public string FieldName { get; set; }
-
-        public bool IsIndexerNameMatch { get; set; }
```
rynowak: We don't use these in Blazor

PR #228, opened 2018-03-09
Add support general delegate types

rynowak:
This builds upon existing support for UIEventHandler-typed component
properties and applies the same principle to any delegate type.

We try to help by generating the LHS of the lambda `=>` allowing you to
write `OnClick="Foo()"` rather than `OnClick="(e) => Foo()"`. You can of
course use @ as an escape.

The only rough edge here is that if the parameter names aren't memorable
for the delgate type, it's not super helpful.


SteveSandersonMS:
Looks great!

> The only rough edge here is that if the parameter names aren't memorable for the delgate type, it's not super helpful.

We'll have to see how it ends up feeling in practice. As long as it's also possible to do `<ItemsEditor OnDelete=@((item) => DoSomething(item)) />` (which I understand it is) then people can give their own names to delegate parameters.

PR #230, opened 2018-03-09
Correct the handling of List<T> in PocoJsonSerializerStrategy

### LINKED ISSUE #225: Http GetJson does not support List<T> as the cast type
Example code:
```csharp
Results = await Http.GetJsonAsync<List<TweetResult>>($"/api/Tweets/{parameters["term"]}");
```

This crashes with the following error:
WASM: [System.Reflection.TargetParameterCountException] Number of parameters specified does not match the expected number.
MonoPlatform.ts:169:28

Changing it to:
```csharp
Results = await Http.GetJsonAsync<TweetResult[]>($"/api/Tweets/{parameters["term"]}");
```

Corrected the problem.

Cheers.

CXuesong:
`List<>` is used for all the `IList`-implementing types except for arrays, and in the following snippet, the caller expects the generated `List<T>` constructor should accept one parameter (`int` of capacity).
https://github.com/aspnet/Blazor/blob/e4b977bc9b5efad6ff13b3b898d805e005a43a30/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs#L1456-L1462
But in the 
`ContructorDelegateFactory`, such constructor is only offered for arrays (`int` of length), so I think this special case need to be extended to `List<T>`.

This PR should address this problem and fix #225.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
@@ -1272,7 +1272,9 @@ namespace SimpleJson
 
         internal virtual ReflectionUtils.ConstructorDelegate ContructorDelegateFactory(Type key)
         {
-            return ReflectionUtils.GetContructor(key, key.IsArray ? ArrayConstructorParameterTypes : EmptyTypes);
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
This is great - thanks so much @CXuesong! It's now merged.

PR #232, opened 2018-03-10
Add Simple onclickAsync for BlazorComponent

RyoukoKonpaku:
Added a simple ```@onclickAsync``` on BlazorComponent to support async Task methods for onclick events.

This should at least be enough for use cases on fetching data that is usually async in nature until something better is created.
I'm not sure if webassembly HTTP supports canceling of requests though, should cancellation token be added?

Should be inline with this tracker #1 on the Lifecycle section.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -202,6 +202,30 @@ namespace Microsoft.AspNetCore.Blazor.Components
             => RenderTreeFrame.Attribute(0, "onclick", _ => handler());
 
         /// <summary>
+        /// Handles async click events by invoking the async <paramref name="handler"/>
+        /// </summary>
+        /// <param name="handler">The async handler to be invoked when the event occurs.</param>
+        /// <returns></returns>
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks for contributing this, @RyoukoKonpaku!

The reason it's not yet merged is that I need to carefully review whether this approach is as performant at runtime as the alternative I have in mind. This approach might cause extra heap allocations due to the multi-layered lambda, so it might be necessary to have a native `UIEventAsyncHandler` concept instead, which involves a lot more implementation.

I'll look into it shortly after the 0.1.0 release. Hope that's OK!

RyoukoKonpaku:
@SteveSandersonMS that's fine xD thanks for the heads up. Been quite fun doing full stack on C# as a junior web dev.

Quite excited for what's to come. Cheers~!

RyoukoKonpaku:
@SteveSandersonMS I think this should be closed? Since blazor is now using the tag helper style ```onclick=``` and moving away from ```@onclick``` syntax.

SteveSandersonMS:
@RyoukoKonpaku Yes good point. I'll make a separate issue to track the need for async event handlers with the new syntax.

PR #244, opened 2018-03-12
Add @page directive

rynowak:
Adds the @page directive and support for specifying routes in components
at compile time.

For now the route is required and must begin with a leading /.


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/PageDirective.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    public class PageDirective
+    {
+        public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
+            "page",
+            DirectiveKind.SingleLine,
+            builder =>
+            {
+                builder.AddStringToken(Resources.PageDirective_RouteToken_Name, Resources.PageDirective_RouteToken_Description);
+                builder.Usage = DirectiveUsage.FileScopedSinglyOccurring;
```
SteveSandersonMS: Could we have `FileScopedMultipleOccurring`? There will be cases where people want to define more than one URL pattern that maps to the same component.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/PageDirectivePass.cs
```diff
@@ -0,0 +1,108 @@
...
+                var directive = directives[i];
+                if (directive.Node.IsImported())
+                {
+                    directive.Node.Diagnostics.Add(BlazorDiagnosticFactory.CreatePageDirective_CannotBeImported(directive.Node.Source.Value));
+                }
+            }
+
+            // Only process the first directive occurrence. The parser will have already flagged multiple
+            // usages as errors.
+            var pageDirective = (DirectiveIntermediateNode)directives[0].Node;
```
SteveSandersonMS: As previous comment, would be good if we allowed multiple.

src/Microsoft.AspNetCore.Blazor/Components/RouteAttribute.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Blazor.Components
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
```
SteveSandersonMS: Also here

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/PageDirective.cs
rynowak: Sure

PR #250, opened 2018-03-13
Fix for Dependency injection doesn't handle inheritance

### LINKED ISSUE #216: Dependency injection doesn't handle inheritance
Example scenario: if you subclass `Router`, e.g.,

    public class MyRouter : Router {}

... and then try to use `<MyRouter />`, it will fail because `Router` didn't get its `UriHelper` injected.

LunicLynx:
fix for #216

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/ComponentFactory.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/ComponentFactory.cs
@@ -56,7 +56,7 @@ namespace Microsoft.AspNetCore.Blazor.Components
         {
             // Do all the reflection up front
             var injectableProperties = type.GetTypeInfo()
-                .GetProperties(_injectablePropertyBindingFlags)
+                .GetPropertiesIncludingInherited(_injectablePropertyBindingFlags)
                 .Where(p => p.GetCustomAttribute<InjectAttribute>() != null);
             var injectables = injectableProperties.Select(property =>
             {
--- a/src/Microsoft.AspNetCore.Blazor/TypeInfoExtensions.cs
+++ b/src/Microsoft.AspNetCore.Blazor/TypeInfoExtensions.cs
@@ -1,0 +1,22 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Blazor
+{
+    public static class TypeInfoExtensions
+    {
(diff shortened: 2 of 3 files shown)


SteveSandersonMS:
Thanks @LunicLynx! This is now merged. Great to get this fixed :)

khellang:
Isn't this what `BindingFlags.FlattenHierarchy` does?

SteveSandersonMS:
@khellang I thought so too, but it only applies to `public` members. We want to include `private` and `protected` too.

khellang:
Ah. Well, it includes `protected`, but not private (from base class):

```csharp
using System;
using System.Reflection;

public class A
{
	public int PublicA { get; set; }

	protected int ProtectedA { get; set; }
	
	internal int InternalA { get; set; }

	private int PrivateA { get; set; }
}

public class B : A
{
	public int PublicB { get; set; }

	protected int ProtectedB { get; set; }
	
	internal int InternalB { get; set; }

	private int PrivateB { get; set; }
}
					
public class Program
{
	public static void Main()
	{
		const BindingFlags flags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.FlattenHierarchy;
		
		foreach (var property in typeof(B).GetProperties(flags))
		{
			Console.WriteLine(property.Name);	
		}
	}
}
```

Outputs:

```
PublicB
ProtectedB
InternalB
PrivateB
PublicA
ProtectedA
InternalA
```

Runnable example: https://dotnetfiddle.net/12Fg1k

LunicLynx:
@khellang I had the same idea. But as @SteveSandersonMS says that does not work.

I think that this isn't easily supported seems like an oversight in the api. Doing it this way returns the public properties several times, which is why the must be filtered with `.Where(prop => prop.ReflectedType == prop.DeclaringType)`. 

It would be nice to have a method `GetDeclaredProperties` with `BindingFlags`.

khellang:
There's also `typeof(Blah).GetTypeInfo().GetRuntimeProperties()`, which seems to return the same results 😄

SteveSandersonMS:
Yes, the docs for that seem wrong or at least very misleading:

> This method returns all properties defined on the specified type, including **inherited, non-public**, instance, and static properties - https://msdn.microsoft.com/en-us/library/system.reflection.runtimereflectionextensions.getruntimeproperties(v=vs.110).aspx

It returns properties that are inherited, and ones that are non-public, but **not** ones that are both inherited *and* nonpublic.

PR #251, opened 2018-03-13
Fix incremental builds

javiercn:
Adds additional inputs needed for incremental builds to work, but at the same time breaks them, because they are broken at an earlier point. What's happening is that the assembly info file is being generated every time even though the inputs don't change.

new
old

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
+++ b/src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
@@ -351,7 +351,10 @@
   <Target
       Name="_LinkBlazorApplication"
       Condition="$(_BlazorShouldLinkApplicationAssemblies) != ''"
-      Inputs="$(BlazorBuildLinkerInputsCache)"
+      Inputs="$(BlazorBuildLinkerInputsCache);
+              @(IntermediateAssembly);
+              @(ReferenceCopyLocalPaths->WithMetadataValue('Extension','.dll'));
+              @(BlazorLinkerDescriptor)"
       Outputs="$(BlazorIntermediateLinkerResultFilePath)"
     >
     <!--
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -351,7 +351,10 @@
   <Target
       Name="_LinkBlazorApplication"
       Condition="$(_BlazorShouldLinkApplicationAssemblies) != ''"
-      Inputs="$(BlazorBuildLinkerInputsCache)"
+      Inputs="$(BlazorBuildLinkerInputsCache);
+              @(IntermediateAssembly);
+              @(ReferenceCopyLocalPaths->WithMetadataValue('Extension','.dll'));
```
rynowak: The extra metadata check here probably isn't needed. Is there a particular problem you're trying to avoid?

javiercn: I don't care about the PDBs

rynowak: Ah interesting. Hadn't thought about that 👍 

The case where Razor uses assemblies we're OK with ref assemblies so we don't have to deal with this.

SteveSandersonMS:
> Adds additional inputs needed for incremental builds to work, but at the same time breaks them, because they are broken at an earlier point

Not sure how to interpret this. What is broken about it (as in, what user-visible problem surfaces)? Should the PR be merged or is it still a work-in-progress?

javiercn:
@SteveSandersonMS The part for producing the blazor output works as expected but there is something in the pipeline that triggers compilation all the time, so a new StandaloneApp assembly gets produced all the time and that triggers everything else (as it should).

So for example linking happens on every build due to the main assembly of the app being newer than the rest of the inputs. I'm going to follow up with Nate today to figure out if this is an MSBuild bug as I suspect.

PR #256, opened 2018-03-13
Fix path in props file

javiercn:


javiercn:
FYI, I'm merging as soon the tests pass. I tried it out by installing the generated templates, restoring the project (with the added nuget package source) and building with the linker on and off.

javiercn:
0384187b26dce8bef67f957c444e5e29bc2a1cb4

PR #257, opened 2018-03-14
Publishing

SteveSandersonMS:
This PR means you can just hit "publish" from VS, or run `dotnet publish -c Release` on the command line, and it will produce a working result for both standalone and ASP.NET-hosted app models.

Once we've published updated nupkgs, it should be possible to check whether this works with Kudu as well.

Most of the implementation comes down to changing which files are picked up during the `GetCopyToPublishDirectoryItems` target. I've set it up so we get the exact same set whether it's a standalone app or one being copied to the publish dir of a host app, but there's slightly different postprocessing in the standalone case (e.g., to generate a default `web.config` if you don't have one). A lot of it feels quite nonstandard, but then there's no existing standard for publishing a client-side .NET app.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
+++ b/src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
@@ -34,9 +34,17 @@
     </ItemGroup>
   </Target>
 
-  <Target Name="BlazorWriteStandaloneWebConfig" AfterTargets="CopyFilesToPublishDirectory">
-    <Copy SourceFiles="$(MSBuildThisFileDirectory)Standalone.Web.config"
-          DestinationFiles="$(PublishDir)web.config"
-          Condition="!Exists('$(PublishDir)web.config')" />
+  <Target Name="BlazorCompleteStandalonePublish" AfterTargets="CopyFilesToPublishDirectory">
+    <!-- Add a suitable web.config file if there isn't one already -->
+    <ItemGroup>
+      <_StandaloneWebConfigContent Include="$([System.IO.File]::ReadAllText('$(MSBuildThisFileDirectory)Standalone.Web.config'))"/>
+    </ItemGroup>
+    <WriteLinesToFile
+      Condition="!Exists('$(PublishDir)web.config')"
+      File="$(PublishDir)web.config"
+      Lines="@(_StandaloneWebConfigContent->Replace('[ServeSubdirectory]','$(AssemblyName)'))" />
+  
+    <!-- Remove the .blazor.config file, since it's irrelevant for standalone publishing -->
+    <Delete Files="$(PublishDir)$(AssemblyName).blazor.config" />
   </Target>
 </Project>
--- a/src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
+++ b/src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
@@ -4,6 +4,7 @@
     <RazorCompileOnPublish>false</RazorCompileOnPublish>
     <GenerateDependencyFile>false</GenerateDependencyFile>
     <IsWebConfigTransformDisabled>true</IsWebConfigTransformDisabled>
+    <BlazorLinkOnBuild Condition="'$(Configuration)'=='Release' AND '$(BlazorLinkOnBuild)'==''">true</BlazorLinkOnBuild>
   </PropertyGroup>
 
   <Target Name="SetBlazorFilesToPublish" AfterTargets="ComputeRefAssembliesToPublish" BeforeTargets="CopyFilesToPublishDirectory">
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
```diff
@@ -0,0 +1,59 @@
+﻿<Project>
+  <PropertyGroup>
+    <BlazorLinkOnBuild Condition="'$(Configuration)'=='Release' AND '$(BlazorLinkOnBuild)'==''">true</BlazorLinkOnBuild>
+    <BlazorPublishDistDir>$(AssemblyName)\dist\</BlazorPublishDistDir>
+
+    <!-- Disable unwanted parts of the default publish process -->
+    <CopyBuildOutputToPublishDirectory>false</CopyBuildOutputToPublishDirectory>
+    <CopyOutputSymbolsToPublishDirectory>false</CopyOutputSymbolsToPublishDirectory>
+    <PreserveCompilationContext>false</PreserveCompilationContext>
+    <RazorCompileOnPublish>false</RazorCompileOnPublish>
```
rynowak: I think we already turn `RazorCompileOnPublish` off in the Razor targets. The `RazorCompileOnBuild`/`RazorCompileOnPublish` properties refer to the way that MVC uses Razor and are thus never relevant for Blazor.

src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
```diff
@@ -0,0 +1,59 @@
...
+      <!-- Don't want to publish the assemblies from the regular 'bin' dir. Instead we publish ones from 'dist'. -->
+      <ResolvedAssembliesToPublish Remove="@(ResolvedAssembliesToPublish)" />
+
+      <!-- Move wwwroot files to output root -->
+      <ContentWithTargetPath Update="@(ContentWithTargetPath)" Condition="$([System.String]::new(%(TargetPath)).StartsWith('wwwroot\'))">
+        <TargetPath>$(BlazorPublishDistDir)$([System.String]::new(%(TargetPath)).Substring(8))</TargetPath>
+      </ContentWithTargetPath>
+      
+      <!-- Publish all the 'dist' files -->
+      <_BlazorGCTPDIDistFiles Include="$(OutDir)dist\**" />
```
rynowak: This kind of thing can get you in trouble with incrementalism and dirty-state when files get **deleted** or **removed**. Generally, speaking `**` is for users, not for SDK authors.

If possible you'll want to recompute this list of files from the actual inputs, and not rely on a wildcard.

src/Microsoft.AspNetCore.Blazor.Build/targets/RazorCompilation.targets
```diff
@@ -19,8 +19,10 @@
   <!-- Something quick for input/output tracking - the assumptions here should match what the CLI does -->
   <ItemGroup>
     <BlazorGenerate Include="**\*.cshtml" />
-    <Content Update="@(Content->WithMetadataValue('Extension', '.cshtml'))">
+    <_BlazorGenerateDeclarationContent Include="@(Content->WithMetadataValue('Extension', '.cshtml'))" />
+    <Content Update="@(_BlazorGenerateDeclarationContent)">
       <Generator>MSBuild:BlazorGenerateDeclaration</Generator>
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
```
rynowak: Its weird to me that this didn't do the right thing. I believe you that there's a bug here, but do you know the explanation?

Is the general idea that `Update="..."` doesn't do what you want unless the you're passing it an actual, unmodified itemgroup?

src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
SteveSandersonMS: Thanks for the info! I've updated it to use the itemgroup computed by our earlier build target rather than looking at the files on disk. Wildcard gone.

src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
SteveSandersonMS: It looks like we turn `RazorCompileOnPublish` **on** in the Razor targets: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets#L94

Also in practice there's a build error if we don't explicitly set it to `false` here (*The type or namespace name 'Razor' does not exist in the namespace 'Microsoft.AspNetCore' (are you missing an assembly reference?)*).

src/Microsoft.AspNetCore.Blazor.Build/targets/RazorCompilation.targets
SteveSandersonMS: It's weird to me too, but it didn't do the right thing before. At least, MSBuild Structured Log Viewer listed all the non-cshtml files with the `Generator: MSBuild:BlazorGenerateDeclaration` metadata as well as the cshtml files.

Since I can't find a single other example of anyone on the internet combining `Update` with `->WithMetadataValue`, it's hard to say whether this syntax is meant to be legit or not.

SteveSandersonMS:
The wildcard thing is fixed, so I'll merge. Thanks!

PR #262, opened 2018-03-14
Make things build with preview1 SDK without warnings

pranavkm:


pranavkm:
This makes it so that you can build a blazor project using the **preview1** sdk without issues. Using this with a preview2 sdk fails pretty miserably, but we can have a guide to update for users trying this with a newer Sdk.

SteveSandersonMS:
Thanks @pranavkm!

It looks good to me, but I think @rynowak is best placed to review it.

rynowak:
> We might still want to find some combination of configs that allow it to work on both preview1 and preview2, even if it also gives warnings on preview2. 

This is the current state right? 

I'm out of the loop on this one about what the problem is so I'm going to put off digging into this until I'm back in the office (tomorrow). Feel free to merge if this is blocking.

pranavkm:
Things work now, but they produce warnings. I spent about 5 mins trying to see if I can get rid of them for preview1. Let's chat when you're in, perhaps there's a little more we could do to make this work in both Sdks.

SteveSandersonMS:
Thanks again @pranavkm.

I've merged this into `dev`, but am leaving the issue opened (with a modified title) because now there are warnings with the preview2 SDK, and I understand you and @rynowak are still going to discuss whether that's something we can also work around.

With your change here it seems totally OK on preview1 which is enough for our 0.1.0 release.

SteveSandersonMS:
Actually it will make more sense if I close this since the PR was merged, and create a new issue for the preview2 warnings.

PR #265, opened 2018-03-14
Stop IL linker from stripping System.Threading.WasmRuntime. Fixes #239

### LINKED ISSUE #239: After IL linking, HttpClient logs warning to console on each request
Here's what it logs:

    * Assertion at /Users/builder/jenkins/workspace/test-mono-mainline-webassembly/label/highsierra/mono/mini/mini-wasm.c:167, condition `method' not met
    abort(). Build with -s ASSERTIONS=1 for more info.

Despite this, the request completes and works successfully. We should track down why it's logging this, and fix it.

SteveSandersonMS:
The issue was that, to implement a timer, some of Mono's C code was invoking .NET methods on this type (when the timer fires). The IL linker doesn't know about that, so it was stripping those .NET methods.

Having this new file `BuiltInBclLinkerDescriptor.xml` also gives us a convenient pattern for adding future exclusions.

@javiercn Your work on integrating the linker cleanly is paying off, because this new behavior was extremely easy to add.


rynowak:
Agree, it's really nice that this is so easy

javiercn:
BTW, there's a descriptor already embedded in mscorlib.dll do both of them get processed? (I would imagine so)

SteveSandersonMS:
> there's a descriptor already embedded in mscorlib.dll do both of them get processed? (I would imagine so)

As I understand it, yes - it logs something like "processing descriptor from mscorlib.dll" (and did so even before this PR).

PR #266, opened 2018-03-14
Add routing support

javiercn:
This is even before WIP.

Putting it out just to show it working. Feel free to look around, download the branch and play with it if you wan to, but restrain from making comments on it for now.


rynowak:
> restrain from making comments on it for now.

This is America, I have the right to comment if I want

javiercn:
This was merged as https://github.com/aspnet/Blazor/commit/841c777991b5a16453952379aaaad73a8d56206c. Leaving the PR open for comments in case there are further cleanups we want to take.

src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            var assembly = Assembly.Load(assemblyName);
+            if (visited.Contains(assembly))
+            {
+                // Avoid traversing visited assemblies.
+                yield break;
+            }
+            visited.Add(assembly);
+            var references = assembly.GetReferencedAssemblies();
+            if (!references.Any(r => string.Equals(r.FullName, blazorAssembly.FullName, StringComparison.Ordinal)))
```
SteveSandersonMS: I'm curious about this logic. What if there was the following dependency graph?

  MyApp -> SuperVendorComponents -> SuperVendorComponentsImpl -> Microsoft.AspNetCore.Blazor

Wouldn't this fail to fetch the components from `SuperVendorComponents` if it only references `SuperVendorComponentsImpl` and doesn't have a direct reference to `Microsoft.AspNetCore.Blazor`?

I'm guessing that `GetReferencedAssemblies` doesn't include transitive references, because if it did then you wouldn't need line 53.

src/Microsoft.AspNetCore.Blazor/Routing/RouteTable.cs
```diff
@@ -0,0 +1,130 @@
...
+namespace Microsoft.AspNetCore.Blazor.Routing
+{
+    internal class RouteTable
+    {
+        public RouteTable(RouteEntry[] routes)
+        {
+            Routes = routes;
+        }
+
+        public RouteEntry[] Routes { get; set; }
```
SteveSandersonMS: Could be `public RouteEntry[] Routes { get; }`?

src/Microsoft.AspNetCore.Blazor/Routing/RouteTable.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        public static RouteTable Create(IEnumerable<Type> types)
+        {
+            var routes = new List<RouteEntry>();
+            foreach (var type in types)
+            {
+                var routeAttributes = type.GetCustomAttributes<RouteAttribute>(); // Inherit: true?
```
SteveSandersonMS: Yes I think we should be allow inherited attributes. It's less confusing than not doing.

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
```diff
@@ -56,6 +46,8 @@ public void Init(RenderHandle renderHandle)
         public void SetParameters(ParameterCollection parameters)
         {
             parameters.AssignToProperties(this);
+            var types = ComponentResolver.ResolveComponents(AppAssembly);
+            Routes = RouteTable.Create(types);
```
SteveSandersonMS: Instead of re-scanning the assemblies and building a new `RouteTable` for each `Router` instance or when `SetParameters` gets called, could we cache the `RouteTable` for any given `AppAssembly`? For example in a `static IDictionary<Assembly, RouteTable>`.

I know it would be uncommon for `Router.SetParameters` to be called more than once, but it is possible depending on how people structure their apps. Also when we get onto support for nested `Router` instances it will become common.

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
```diff
@@ -96,31 +65,32 @@ private string StringUntilAny(string str, char[] chars)
...
         }
 
         private void Refresh()
         {
             var locationPath = UriHelper.ToBaseRelativePath(_baseUriPrefix, _locationAbsolute);
-            var matchedComponentType = GetComponentTypeForPath(locationPath);
-            if (!typeof(IComponent).IsAssignableFrom(matchedComponentType))
+            locationPath = StringUntilAny(locationPath, _queryOrHashStartChar);
+            var context = new RouteContext(locationPath);
+            Routes.Route(context);
```
SteveSandersonMS: Could lines 80 and 81 be moved into a `protected virtual` method called `GetComponentTypeForPath` that by default returns `context.Handler`? Then people can still subclass `Router` if they want, and put in completely custom route matching logic if they need it. I don't think we need to pass the `RouteTable` into this method, as that's internal - we should only pass the original `locationPath` (i.e., before truncating it at `_queryOrHashStartChar`).

Or do you think there's an argument that people *shouldn't* implement custom route matching logic?

test/testapps/BasicTestApp/RouterTest/Links.cshtml
```diff
@@ -1,5 +1,5 @@
-﻿@inject Microsoft.AspNetCore.Blazor.Services.IUriHelper uriHelper
-
+@page "/Links"
```
SteveSandersonMS: Do we need the `@page` directive here? I thought this was only used as a component, not as a routable page in itself.

src/Microsoft.AspNetCore.Blazor/Routing/RouteTable.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            var xTemplate = x.Template;
+            var yTemplate = y.Template;
+            if (xTemplate.Segments.Length != y.Template.Segments.Length)
+            {
+                return xTemplate.Segments.Length < y.Template.Segments.Length ? -1 : 1;
+            }
+            else
+            {
+                for (int i = 0; i < xTemplate.Segments.Length; i++)
```
SteveSandersonMS: Is it an accident that this whole `for` block is repeated below? It looks like lines 74-86 should be removed because their logic is repeated (with some additional stuff) directly below.

src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
SteveSandersonMS: On further consideration, I've realised it's probably better if we *don't* scan any referenced assemblies. We should only include components in `AppAssembly`. This is so that, in the future, we can add a feature where you explicitly opt in to mounting other assemblies at specified points in the URI space. For example, you might choose to attach "MyCompany.MyProduct.Administration" with the URI prefix `/admin`.

So for now I think we can avoid the complexity here entirely :)

javiercn: How would you split the implementation between supervendorcomponents and supervendorcomponentsimpl. Isn't supervendorcomponentsimpl the only thing that matters?

That said, we do a similar thing on MVC, we only take into account libraries that reference at least one of the MVC libraries directly.

src/Microsoft.AspNetCore.Blazor/Routing/RouteTable.cs
javiercn: Yes

test/testapps/BasicTestApp/RouterTest/Links.cshtml
javiercn: Maybe not. I added it just by inertia

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
javiercn: That's fair. How do you feel about statics vs DI for framework components?

src/Microsoft.AspNetCore.Blazor/Routing/RouteTable.cs
javiercn: It's an accident. I don't know how I missed this 😖

src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
SteveSandersonMS: Let's not worry about those details. For now we just don't consider any assembly other than `AppAssembly`.

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
SteveSandersonMS: I don't necessarily have a preference. What do you think would work better? We are using statics for caching stuff in other places already. For any case where we do use a `static`, consider SSR: it's important that it never leaks any state that would be different from one user to another. In this example I think that's OK because the computed route table would always be identical for any given `AppAssembly`.

rynowak:
This was merged right? Should this PR be closed?

SteveSandersonMS:
@rynowak It was merged by @javiercn in this comment: https://github.com/aspnet/Blazor/pull/266#issuecomment-373887925, but there are still some outstanding CR items to be resolved once Javier gets back.

rynowak:
OK Cool beans

javiercn:
Yes. I’ll address the feedback when I’m back

src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
SteveSandersonMS: **Update**: Changed my mind. I think this is fine as-is. Let me know if you want to discuss reasoning.

javiercn:
Closing this as there's a separate PR for addressing the feedback.

PR #270, opened 2018-03-15
Workaround for VS publishing failures for hosted scenario

SteveSandersonMS:
Although publishing worked correctly from the command line (for both standalone and hosted-style solutions), and it worked in VS for standalone Blazor projects, I discovered it does **not** work in VS for hosted Blazor projects.

### The problem

The original reason was:

 1. `BlazorGetCopyToPublishDirectoryItems` did not declare a dependency on `PrepareBlazorOutputs`, so it wasn't guaranteeing that `@(BlazorItemOutput)` would be populated yet.
 2. When the publish runs in VS, a different set of targets gets run. It seems that VS does a regular build first, then runs publish in some kind of "don't build first" mode. The result was that `PrepareBlazorOutputs` did not run, hence `@(BlazorItemOutput)` was empty, hence we didn't know what files to publish.

I thought I could fix it by making `BlazorGetCopyToPublishDirectoryItems` depend on `PrepareBlazorOutputs`, but this opens up a whole different set of problems. Although `PrepareBlazorOutputs` relies on `@(ReferenceCopyLocalPaths)` being populated, it doesn't explicitly take a dependency on `ResolveReferences`. Again, in the VS publish scenario, no regular build is occurring, so `@(ReferenceCopyLocalPaths)` is empty and `PrepareBlazorOutputs` fails.

I thought I could fix it by making `PrepareBlazorOutputs` depend on `ResolveReferences` (or even `Build`), but this opens up a whole different set of problems. The `ResolveReferences` target knows that no regular build is taking place, so it only collects a subset of the regular references list. This means that it doesn't populate `@(ReferenceCopyLocalPaths)` with what we need. And if we try making `PrepareBlazorOutputs` depend on `Build`, then the build fails because it's not referencing the full set of assemblies.

### Workaround

In practice, a build of the client app always *does* get run prior to the publish phase (both for VS and command-line publishes). Therefore the client app `dist` directory is always up to date. As a cheap workaround, then, I've just reverted the "don't use wildcards" change from before so now it publishes whatever's in the `dist` directory on disk without needing to declare a dependency on `PrepareBlazorOutputs`.

### Longer-term proper fix

@javiercn, do you think the `PrepareBlazorOutputs` target *should* be able to run during a VS publish? If so, would you be able to look into wiring up the correct dependencies so that it does succeed? Not this week of course!

If not, we might have to review more broadly how the build/publish works.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
+++ b/src/Microsoft.AspNetCore.Blazor.Build/targets/Publish.targets
@@ -12,6 +12,19 @@
     <IsWebConfigTransformDisabled>true</IsWebConfigTransformDisabled>
   </PropertyGroup>
 
+  <!-- Ideally, this should have DependsOnTargets="PrepareBlazorOutputs" and then
+       populate _BlazorGCTPDIDistFiles using @(BlazorItemOutput) instead of using
+       the wildcard that publishes whatever's currently on disk. However that
+       doesn't currently work because PrepareBlazorOutputs assumes the build will
+       already have run and hence @(ReferenceCopyLocalPaths) will already be
+       populated. The assumption is correct for regular builds and command-line
+       publishes, but it's *not* correct for VS publishes. Somehow we'd need
+       PrepareBlazorOutputs to also depend on Build or at least ResolveReferences,
+       but that alone doesn't cause @(ReferenceCopyLocalPaths) to be populated,
+       because ResolveReferences specifically behaves differently when
+       '$(BuildingProject)'!='true'. Hackily overriding that flag to true still
+       somehow isn't sufficient, but even if it was it's not a good solution as
+       it might have other side-effects. -->
   <Target Name="BlazorGetCopyToPublishDirectoryItems" BeforeTargets="GetCopyToPublishDirectoryItems">
     <ItemGroup>
       <!-- Don't want to publish the assemblies from the regular 'bin' dir. Instead we publish ones from 'dist'. -->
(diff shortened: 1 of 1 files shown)


javiercn:
@rynowak Is the expert on all things publishing (as I haven't ever implemented it myself) but this is more or less what I've talked about with him in the past about the topic (Ryan feel free to correct me).
1) Publish should typically pretty much copy items, build should do most of the work to create the files necessary for getting something runable.
2) If you are publishing and the project was already built with your configuration and linking options, is just a matter of copying files.
3) If the project was already built but without linking, then we need to invoke the linker to produce the optimized version (hence _ShouldLinkerRun).
4) If the project wasn't build and we need to publish we can error out and indicate that a build needs to happen first.

SteveSandersonMS:
@javiercn Those are all good principles but I'm unsure how we'd apply them to this situation. Do you think that `PrepareBlazorOutputs` should be able to run during a publish? If not, how do we know which Blazor outputs to publish?

javiercn:
@SteveSandersonMS Not exactly prepare blazor outputs but something very similar. Something like this:
```
PrepareBlazorPublishOutputs
  GetBlazorPublishConfiguration
    ReadBlazorBuildOutputs
  ResolvePublishBinaries
    Run the linker (BlazorLinkOnPublish=true)
      Do nothing if the build already linked the assemblies.
      Run the linker against the build outputs and copy the newly linked assemblies.
    Copy assemblies (BlazorLinkOnPublish=false)
      If we didn't link on build, do nothing. (we just need to copy the assemblies)
      If we linked on build and we don't want to link on publish (resolve reference assemblies)
  CopyFilesToPublishDirectory
```
Does that make sense?

SteveSandersonMS:
Maybe. Will all those steps be able to run correctly if `@(ReferenceCopyLocalPaths)` is not populated? Are you somehow storing the information that was in `@(ReferenceCopyLocalPaths)` on the previous build and reading it back?

javiercn:
@SteveSandersonMS Take a look at this to see if it makes sense https://github.com/aspnet/Blazor/commit/690a43b44218f6c35f005feba332df29b9999b86

javiercn:
> I thought I could fix it by making BlazorGetCopyToPublishDirectoryItems depend on PrepareBlazorOutputs, but this opens up a whole different set of problems. Although PrepareBlazorOutputs relies on @(ReferenceCopyLocalPaths) being populated, it doesn't explicitly take a dependency on ResolveReferences. Again, in the VS publish scenario, no regular build is occurring, so @(ReferenceCopyLocalPaths) is empty and PrepareBlazorOutputs fails.

You should be able to make the above work with the changes mentioned above + a few tweaks to make the configuration suitable for publish.

The idea apparently is that you won't copy stuff from the bin folder but from the obj folder instead, you just depend on PrepareBlazorOutputs and then define the items to copy to the publish directory

SteveSandersonMS:
Thanks @javiercn, this is much better! Closing this in favour of #274.

PR #274, opened 2018-03-15
A better fix for the VS publishing hosted projects issue

SteveSandersonMS:


SteveSandersonMS:
This is more or less the same as in Javier's suggestion at 690a43b. Updates:

* Didn't change how `@(IntermediateAssembly)` is managed, because it already works fine in all the scenarios
* The rule for choosing whether to read the inputs back from `$(BlazorLocalReferencesOutputPath)` is now "did `ResolveReferences` already run or not". Because if it did already run, then we always want the freshest data and have no reason to assume that `$(BlazorLocalReferencesOutputPath)` is up-to-date. We only read the data back from the file when we're in the VS publishing scenario, where VS already takes care of having run a build immediately before it invokes the publish target.

PR #277, opened 2018-03-16
Enable Razor build server by default

SteveSandersonMS:


SteveSandersonMS:
@pranavkm @rynowak @javiercn From our discussions it sounds like there's no reason not to do this, so I'll merge now, but please let me know if you disagree and we can change it.

src/Microsoft.AspNetCore.Blazor.Build/targets/All.props
```diff
@@ -3,5 +3,8 @@
 
   <PropertyGroup>
     <DefaultWebContentItemExcludes>$(DefaultWebContentItemExcludes);wwwroot\**</DefaultWebContentItemExcludes>
+
+    <!-- We can remove this after updating to newer Razor tooling, where it's enabled by default -->
+    <UseRazorBuildServer>true</UseRazorBuildServer>
```
pranavkm: We set this to true in newer builds of the Razor Sdk. Might not need this to be explicitly set if you update the other

PR #282, opened 2018-03-16
[WIP] Add `activeClass` attribute to NavLink component

austincummings:
For #162.

Currently the "active" class is set when a route link becomes active. Adds a simple `activeClass` attribute to the `NavLink` component to control this. Will be set to `string.Empty` if no value was supplied.

Also updated the samples and templates to use the new attribute.


SteveSandersonMS:
Thanks @austincummings for contributing this!

We'd be happy to merge it, though there are a couple of tweaks I'd like to request:

* Can we please retain the default active class value of `"active"` if no other value was specified? It's preferable not to have to override this for developers who are happy with the default CSS class name.
* Could you rename `activeClass` to `ActiveClass`? That's because it's a property of the `NavLink` component which is a .NET type, so like the `Match` property, it follows C# naming conventions. This differs from (e.g.) `href` which is merely a DOM attribute.

Again, we appreciate the contribution!

austincummings:
@SteveSandersonMS Sure thing. I'll make those changes and push tonight. Thanks!

austincummings:
Hey @SteveSandersonMS, I made those changes. I did run into some weirdness though.

I wanted to only apply the `class` attribute if the value string was not null or empty in order to keep the DOM clear of empty class attributes. So I had this code in the render method...

```c#
// Set class attribute
string classAttrValue = CombineWithSpace(_cssClass, _isActive ? ActiveClass : null);
if (!string.IsNullOrEmpty(classAttrValue)) 
{
    builder.AddAttribute(0, "class", classAttrValue);
}

// Pass through all other attributes unchanged
foreach (var kvp in _allAttributes.Where(kvp => kvp.Key != "class"))
{
    builder.AddAttribute(0, kvp.Key, kvp.Value);
}
```

Now after clicking onto and then away from any of the `NavLink`s they would lose all their attributes and they'd be unclickable.

If I moved the class attribute code to after the "other" attributes code like this...

```c#
// Pass through all other attributes unchanged
foreach (var kvp in _allAttributes.Where(kvp => kvp.Key != "class"))
{
    builder.AddAttribute(0, kvp.Key, kvp.Value);
}

// Set class attribute
string classAttrValue = CombineWithSpace(_cssClass, _isActive ? ActiveClass : null);
if (!string.IsNullOrEmpty(classAttrValue)) 
{
    builder.AddAttribute(0, "class", classAttrValue);
}
```

then it would work just fine. I'm not sure what's going on here or if I made a mistake that I'm just not seeing. The code right now is applying the class attribute whether it has a value or not. It just has an empty class attribute like this `<a class href="...">`.


Also not sure why the last build failed when I only updated a comment. Looks like a fluke.

Thanks!

SteveSandersonMS:
Thanks for the updates @austincummings!

> I made those changes. I did run into some weirdness though.

Yes, we realised in #444 that there was a bug with that, and it will be fixed when #444 is merged.

Once that's gone in, we should be able to merge this PR.

SteveSandersonMS:
Thanks @austincummings - this is now merged!

thepra:
Is there any way to override that default "active" class in a global way? My case is to adapt the NavLink to the css framework in use, which active class differs from that one.

Issue #219, opened 2018-03-08
C# expressions in attributes with hyphens don't compile


SteveSandersonMS:
In a cshtml file, try `<div data-something="@someValue"></div>`

rynowak:
I imagine this is a bug in the node writer logic. If you think we're close to removing support for the temporary `<c:Foo>` syntax, I'd prefer to do that first.

I have a PR out for https://github.com/aspnet/Blazor/issues/214 now.

rynowak:
I can confirm that this does reproduce easily though

SteveSandersonMS:
@rynowak I know this is assigned to you, and maybe you're addressing it as part of the `bind` work, but would it be OK if I put in a fix for this? It would help with some sample stuff I'm doing.

rynowak:
Go ahead, I don't think I can get to this before monday

davidnmbond:
Has this one been closed because there is a temporary workaround?  If so, it should be re-opened.

This still doesn't compile:

	<span class="input-group-text" data-container="body" data-toggle="popover" data-placement="left" data-content="@HelpText">?</span>

rynowak:
Hi @davidnmbond - we know of an issue specifically with `data-` attributes that we're tracking at https://github.com/aspnet/Blazor/issues/297

You can work around that for now by using `data_`. 

If that doesn't resolve your problem please open a new issue and provide details so we can investigate.

PR #292, opened 2018-03-19
Temporary workaround for #219 ("data-" attributes don't compile)

SteveSandersonMS:
The problem with compiling `data-` attributes comes down to this in Razor's `HtmlMarkupParser`: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs#L934

That is, it specifically detects `data-` as an attribute prefix, and when found outputs different IR that unfortunately doesn't retain information about the fact that it's even an attribute (it just looks like arbitrary HTML markup in the IR). As such, the Blazor node writer doesn't know we're dealing with an attribute and can't produce the correct code.

### Short-term workaround

Since Blazor developers may need to output `data-*` attributes into the DOM (e.g., to control certain Bootstrap features), there has to be a way of doing it.

As a short-term workaround, this PR causes `data_*` attributes to be rendered as `data-*` in the DOM. This at least gives developers a way of producing `data-*`, even it's quite awkward. This workaround is probably preferable over delaying the 0.1.0 release.

We can write it up as a "known issue" in the 0.1.0 release notes.

### Longer-term solution

I understand there will be good reasons the `data-*` special case in `HtmlMarkupParser` for server-side use (based on the HTML semantics of `data-*` attributes), but it's not applicable for client-side execution. So, we'll want some way of disabling that special case in `HtmlMarkupParser` when compiling Blazor components.

Right now I don't see any good way of disabling that special case from the outside. I think we'll need to modify `HtmlMarkupParser.cs` in the Razor repo, giving it some config option or other method to signal we're building client-side code.

@rynowak Do you agree? Can we tell `HtmlMarkupParser` not to treat `data-*` as special if it's doing Blazor compilation? Do you have a preference on how the configuration for that should work?

If it's totally trivial to add and to update Blazor to depend on a newer version of Razor that supports this, we should consider it for 0.1.0. But if it's nontrivial in any way I think we should do it post-0.1.0. We can discuss.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
@@ -543,6 +543,15 @@ namespace Microsoft.AspNetCore.Blazor.Razor
 
         public override void BeginWriteAttribute(CodeWriter codeWriter, string key)
         {
+            // Temporary workaround for https://github.com/aspnet/Blazor/issues/219
+            // Remove this logic once the underlying HTML parsing issue is fixed,
+            // as we don't really want special cases like this.
+            const string dataUnderscore = "data_";
+            if (key.StartsWith(dataUnderscore, StringComparison.Ordinal))
+            {
+                key = "data-" + key.Substring(dataUnderscore.Length);
+            }
+
             codeWriter
                 .WriteStartMethodInvocation($"{_scopeStack.BuilderVarName}.{nameof(BlazorApi.RenderTreeBuilder.AddAttribute)}")
                 .Write((_sourceSequence++).ToString())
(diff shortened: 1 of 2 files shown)


rynowak:
That that end, what I've had in mind the first time something like comes up is to add a `Version_Experimental` to the set of language versions recognized by Razor. https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs#L22

For now version=experimental means that we can add/tweak things for Blazor with no long term support contract (experimental explicitly allows breaking changes between releases).

Then we add a feature toggle for this behavior and toggle it off for version=experimental.

----

The keys from my point of view is that:
- this isn't a big investment from Razor (it's just an if)
- we can do this without expanding the API surface of Razor in a permanent way

SteveSandersonMS:
@rynowak That sounds good to me. What do you think we should do for 0.1.0, given the goal of shipping tomorrow or maybe Wednesday at a push? This PR as-is, or the new experimental target for Razor?

rynowak:
We should proceed with a workaround for now. I won't be able to deliver anything additional by then.

rynowak:
https://github.com/aspnet/Razor/issues/2183 FYI

Issue #278, opened 2018-03-16
Proposal: make autogenerated BlazorComponent views partial class


uazo:
I would like to split the Razor c # + html template from the purely functional code of the view and I tried to make every generated class contained in "BlazorRazorComponents.g.cs" with the partial modifier (simply adding it in "Microsoft.AspNetCore.Blazor.Razor. Extensions \ BlazorRazorEngine.cs ").

So my views are composed in

nomeview.cshtml       <- html layout
nomeview.cshtml.cs   <- functional code

The compilation works without problems.
The problem is Visual Studio is unable to "read" the contents of the code in cshtml.cs, showing the error CS1061 'Class' does not contain a definition for 'Method'.

Now, adding the directive

`@inherits <name of the same class>`

instead Visual Studio works correctly, while, obviously, the compilation is not successful because the autogenerated code becomes

`public partial class <classname>: <classname> <- the same`

But if the code remain

`public partial class <classname>: BlazorComponent`

both the compilation and Visual Studio would work.

Is it possible to make that if implements the same class, BlazorComponent is kept as a base class?

SteveSandersonMS:
You can use the `@inherits` mechanism today in the way you've described if you make your "code behind" class have a different name. For example, have `MyPage.cshtml` inherit from `MyPageBase.cs`.

However I agree with you that `partial` would be more convenient. I'm totally open to doing that. @rynowak we discussed this before - do you think we should? I can easily add the `partial` modifier to our compiler if we want that.

uazo:
Yes, I know, in fact for now I'm proceeding this way.

But to simplify the navigation of the project (for example through intellisense), I would prefer not to have basic classes that for now that are not useful for the project.
Among other things, it would also be useful for the future to have an "abstract" directive, so that the component can be defined as abstract, although in that case it is necessary to review the way in which the sequence number is constructed in the BuildRenderTree builder.

rynowak:
Yeah, I agree with @uazo - I think partial makes a lot more sense for the Blazor use case. There's no reason not to just do this.

vibeeshan025:
Is there any possiblity to expect support for the View class to have the ability to be used as **Generic** class, WPF, Silverlight, UWP none of them support this, however this helps us to swap the view or the model depending on the scenario.

`public partial class MyView<T>: BlazorComponent
{
        public T ViewModel;
}` 

An extenral constructor can decide, what the view and viewmodel will be depending on the scenario.

[...]

PR #294, opened 2018-03-19
[WIP] Codegen partial classes. Implements #278

SteveSandersonMS:
Although this works fine at compile time and runtime, it doesn't work correctly at design time in VS:

image

I'm unsure why partial classes from the same project aren't included in the design-time compilation that VS is getting these errors from, especially given that things like `@inherits SomeClass` works fine.

@rynowak, do you know why this is, and if there's any straightforward thing we can do in this repo to fix it?

**Not urgent!** This doesn't need to be included in 0.1.0.


rynowak:
Ah, I see.

The problem with this is that the editor code is **Fake**. It mangles the class name so that it doesn't conflict with the file on disk. I'm sad that we can't do this triviall 😢 

I have a few ideas for how to fix it, but they will be more involved.

danroth27:
#278

SteveSandersonMS:
@rynowak Since this PR is relatively simple, and the work to really enable `partial` might be a while off, I think I'll just close this and then we can reimplement/rediscover it later. Please reopen if you disagree.

PR #295, opened 2018-03-19
Basic route constraints (:int, etc.)

SteveSandersonMS:
I think the first thing people will do with routing is navigation to entities by ID, e.g., `/products/{productId}`. Since most IDs are `int` or `Guid`, that would be inconvenient because previously we only supplied parameters of type `string`.

This PR adds support for `{productId:int}`, `{productId:guid}`, etc.

I know it's a little different to ASP.NET server-side routing, in that the type coercion is happening inside routing, whereas for MVC it happens inside model binding. Since Blazor doesn't have an equivalent to model binding (and doesn't need or want one), it makes sense to do the coercion here. If we decide we absolutely must do the coercion later in the flow (e.g., during application of parameters to a component) we can consider that but it will be trickier because we definitely don't want *all* parameters to be auto-coerced by default.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Layouts/LayoutDisplay.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Layouts/LayoutDisplay.cs
@@ -26,7 +26,7 @@ namespace Microsoft.AspNetCore.Blazor.Layouts
         /// <summary>
         /// Gets or sets the parameters to pass to the page.
         /// </summary>
-        public IDictionary<string, string> PageParameters { get; set; }
+        public IDictionary<string, object> PageParameters { get; set; }
 
         /// <inheritdoc />
         public void Init(RenderHandle renderHandle)
--- a/src/Microsoft.AspNetCore.Blazor/Routing/RouteConstraint.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Routing/RouteConstraint.cs
@@ -45,26 +45,26 @@ namespace Microsoft.AspNetCore.Blazor.Routing
             switch (constraint)
             {
                 case "bool":
-                    return new RouteConstraintType<bool>(bool.TryParse);
+                    return new TypeRouteConstraint<bool>(bool.TryParse);
                 case "datetime":
-                    return new RouteConstraintType<DateTime>((string str, out DateTime result)
+                    return new TypeRouteConstraint<DateTime>((string str, out DateTime result)
                         => DateTime.TryParse(str, CultureInfo.InvariantCulture, DateTimeStyles.None, out result));
                 case "decimal":
-                    return new RouteConstraintType<decimal>((string str, out decimal result)
+                    return new TypeRouteConstraint<decimal>((string str, out decimal result)
                         => decimal.TryParse(str, NumberStyles.Number, CultureInfo.InvariantCulture, out result));
                 case "double":
-                    return new RouteConstraintType<double>((string str, out double result)
+                    return new TypeRouteConstraint<double>((string str, out double result)
                         => double.TryParse(str, NumberStyles.Number, CultureInfo.InvariantCulture, out result));
                 case "float":
-                    return new RouteConstraintType<float>((string str, out float result)
+                    return new TypeRouteConstraint<float>((string str, out float result)
                         => float.TryParse(str, NumberStyles.Number, CultureInfo.InvariantCulture, out result));
                 case "guid":
-                    return new RouteConstraintType<Guid>(Guid.TryParse);
+                    return new TypeRouteConstraint<Guid>(Guid.TryParse);
                 case "int":
-                    return new RouteConstraintType<int>((string str, out int result)
+                    return new TypeRouteConstraint<int>((string str, out int result)
                         => int.TryParse(str, NumberStyles.None, CultureInfo.InvariantCulture, out result));
                 case "long":
-                    return new RouteConstraintType<long>((string str, out long result)
+                    return new TypeRouteConstraint<long>((string str, out long result)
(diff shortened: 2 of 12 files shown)


campersau:
It would be cool if this is expandable via options similar to https://github.com/aspnet/Routing/blob/dev/src/Microsoft.AspNetCore.Routing/RouteOptions.cs

rynowak:
> I know it's a little different to ASP.NET server-side routing, in that the type coercion is happening inside routing, whereas for MVC it happens inside model binding.

I wouldn't recommend that we try to make it perfectly MVC-like in this. What you have here matches users expectations. 

We get lots of confusion about what MVC does that would be solved by doing what you do here.

rynowak:
I think this looks great overall, but I question the utility of route constraints beyond int, guid, date and decimal. I think in MVC they are there 'for completeness' and are just a thing that exists that people never use.

Some thoughts about a few specific constraint types - I don't think these things are important for a 0.1 but food for thought.

numeric types: We do special work in MVC's model binders to parse numeric types with the thousands separator, and I think to avoid some formats that folks don't frequently type https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/FloatingPointTypeModelBinderProvider.cs

dates: my expectation is that folks will want to parse formats like MM-DD-YY or DD-MM-YYYY and not do too much more than that. It might make more sense long term to just define exactly what we mean than just call it `datetime` and parse everything that DateTime accepts in InvariantCulture.

javiercn:
Overall looks good. A few thoughts:
I think we can get away without having to add additional syntax if we are willing to accept the following:
* We don't need to have an extensible system for constraints (for now).
* We know the type of the property we are being bound to.

The idea would be to just do this automatically for you based on the property on the component is going to be bound to. The benefit is that you only need to specify the information of what you expect in one place (the type of the property).

We could do it somewhere around here:
https://github.com/aspnet/Blazor/blob/dev/src/Microsoft.AspNetCore.Blazor/Routing/RouteTable.cs#L29

* We would create the route template.
* We would inspect the public properties of the component and add a constraint for the right type automatically for you.
* When we are (parsing the template/binding page parameters) we convert the value to the right type.

SteveSandersonMS:
> We know the type of the property we are being bound to.

That's the thing we can't guarantee. The `IComponent` contract doesn't require a component to declare its set of parameters up front. This makes components more flexible as a concept. For example, `NavLink` doesn't need to know about all possible HTML attributes - it just passes through arbitrary incoming parameters onto rendered HTML attributes.

Given that we probably want to support these simple route constraints anyway, even if we didn't use them for type coercion, I think we're not doing any bad thing by also using them for type coercion right now. It doesn't prevent us from also adding some kind of auto-coercion in the future if we wanted to.

SteveSandersonMS:
@rynowak 

> I question the utility of route constraints beyond int, guid, date and decimal

I agree others are unlikely (though wouldn't be surprised to see `/map/{latitude:double}/{longitude:double}` in the wild).

Do you think we should actually remove the less common ones? I thought that there wasn't really any extra cost in supporting them, and it's one fewer thing to have to remember as a difference vs potential expectations.

> parse numeric types with the thousands separator

I'll make sure "thousands separator" is covered and will add a test case.

>  It might make more sense long term to just define exactly what we mean than just call it datetime and parse everything that DateTime accepts in InvariantCulture.

Isn't that what we're already doing in https://github.com/aspnet/Blazor/pull/295/files#diff-347a2f8dde39261e6a0e1c33423e1e7bR51 ? Should this be changed in some way?

rynowak:
>Isn't that what we're already doing in https://github.com/aspnet/Blazor/pull/295/files#diff-347a2f8dde39261e6a0e1c33423e1e7bR51 ? Should this be changed in some way?

From docs
> The DateTime.TryParse(String, IFormatProvider, DateTimeStyles, DateTime) method parses a string that can contain date, time, and time zone information. It is similar to the DateTime.Parse(String, IFormatProvider, DateTimeStyles) method, except that the DateTime.TryParse(String, DateTime) method does not throw an exception if the conversion fails.

`DateTime.TryParse` accepts a lot of stuff. If the goal is to serialize a .NET DateTime in the URL, it will work for that. If your goal is to support the more common usage for a date in the URL path - this brings a lot more power than is needed for that.  Because it will 'just work', you'll find people using this to serialize a .NET DateTime in the URL, and then they'll have questions and feedback about that usage. We see these questions expressed in MVC usually phrased as "why is the URL not culture/language sensitive"


I think this is a fine thing to do for 0.1. We might want to think about making it more limited in the future.


>  I thought that there wasn't really any extra cost in supporting them, and it's one fewer thing to have to remember as a difference vs potential expectations.

I think the key is, would we **tell someone** to use these features. This is a factor of the goodness of its design (in context) and whether or not we have confidence that they will be satisfied with the outcome. The extra cost is a big metaphysical, but it's the cost of someone finding and using a thing, where it isn't the thing they want. 


------------

I don't feel really strongly about any of this feedback. My experience with the MVC version of these is that users are very underwhelmed and I think the Blazor version goes a long way towards addressing that.

javiercn:
> That's the thing we can't guarantee. The IComponent contract doesn't require a component to declare its set of parameters up front.

But when you are doing the routing piece you are not dealing with IComponent but with the actual type, so you get a chance to inspect its properties and bound them however you want. Am I missing something here?

SteveSandersonMS:
Yes, it doesn't have to be a property at all. The `IComponent` contract doesn't require parameters to match any property. An `IComponent` can implement whatever logic it wants in `SetParameters` to do things with incoming parameters, whether or not there exist any matching properties.

PR #321, opened 2018-03-21
lambda fixes

rynowak:


rynowak:
Tested this out with Flightfinder and different syntaxes and it all seems pretty good.

Also noticed that `<Search OnSearch="async (e) => await Task.Delay(50)" /> works pretty well also.

rynowak:
Will cherry pick to release asap

PR #323, opened 2018-03-21
Fix the Blazor about dialog

rynowak:
Have to do some gritty work to be able to show a proper version in the
about dialog. The generation of the proper assembly attributes is part
of the new-style csproj, so we don't get to use it for free.


SteveSandersonMS:
@rynowak @danroth27 I wasn't sure whether this should be in `release/0.1.0`, but it looks like it, so I have cherry-picked it over there for inclusion in 0.1.0. If this was incorrect please let me know and we can rebuild.

danroth27:
Yes, this should be in 0.1.0.

rynowak:
I thought I had already done this 😲

PR #325, opened 2018-03-22
Move standalone template into subdirectory with .sln

SteveSandersonMS:
Workaround for https://developercommunity.visualstudio.com/content/problem/217493/visual-studio-nuget-package-manager-ignores-nugetc-1.html

Otherwise, VS can't restore packages if you create a "standalone" app on a machine that's never restored the Blazor packages before


SteveSandersonMS:
Merged

PR #326, opened 2018-03-22
Update third party notices

danroth27:


SteveSandersonMS:
I don't exactly know what the requirements are for this file, but on the surface it looks good to me.

PR #343, opened 2018-03-23
Improve dependencies of the Blazor VSIX

rynowak:
This should now correctly require VS 15.7-preview1 or newer with the
ASP.NET and web development tools to installed.

Removed the Razor Language Servcies dependency since it's prompting an
install of the wrong tools from the VS gallery.


rynowak:
FYI

PR #346, opened 2018-03-24
Remove workaround for old Razor tooling

rynowak:
This isn't needed anymore to support the Blazor design-time experience.

Now that it's gone, it will no longer cause conflicts with MVC's types
so we can remove the other workaround (privateassets).


src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Client/BlazorHosted.CSharp.Client.csproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.1.0-preview2-30230" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.AspNetCore.Blazor.Browser" Version="0.1.0-rc1-10117" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.AspNetCore.Blazor.Browser" Version="0.1.0-rc1-10117" />
```
rynowak: @SteveSandersonMS do you remember why we have PrivateAssets on the Razor build package?

Issue #334, opened 2018-03-23
Move to Bootstrap 4


danroth27:

FabriDamazio:
@danroth27 i'm doing it already. If all works i'll do a PR

danroth27:
Done https://github.com/aspnet/Blazor/pull/619

Issue #329, opened 2018-03-23
 Bootstrap hamburger menu is not working


SIkebe:
Bootstrap hamburger menu is not working.
Adding `jquery.min.js` and `bootstrap.min.js` resolves it.

### How to repro
1. Select File -> New Project -> Web -> ASP.NET Core Web Application
2. Pick the Blazor template
3. Press Ctrl + F5
4. Reduce the browser width and show hamburger menu

camou23:
Can i ask you how you added them?
I tried to add them in the idex.html but it seems im doing something wrong

SIkebe:
@camou23 Nothing special 😄 
```
<body>
    <app>Loading...</app>

    <!-- add these js files-->
    <script src="js/jquery.min.js"></script>
    <script src="js/bootstrap.min.js"></script>

    <script src="css/bootstrap/bootstrap-native.min.js"></script>
    <script type="blazor-boot">
    </script>
</body>
```

FabriDamazio:
I'm investigating it and appears be an issue with the `bootstrap-native.min.js`
Maybe something related with [this](https://github.com/thednp/bootstrap.native/issues/136)

SteveSandersonMS:
Adding them to `index.html` should work fine. Just remember to rebuild your application after (because the `index.html` file is *built* so we can perform the replacement on the `blazor-boot` script tag).

I expect we won't fix this issue in isolation, because instead we'll soon move to Bootstrap 4 in the templates, so the issue will no longer apply. Hope that's OK!

PR #363, opened 2018-03-24
Move to bootstrap v4

FabriDamazio:
Move to Bootstrap to v4. Addresses #334 
Remove fonts folders because Bootstrap v4 doesn’t include an icon library by default.


FabriDamazio:
@AWulkan you are right. I removed the  redundant MIME types. Tkx!

Eilon:
@danroth27 - please note that if we take this, we need to update the ThirdPartyNotices.

FabriDamazio:
@Eilon i can update the ThirdPartyNotices doc if you wish.

Eilon:
@FabriDamazio you can give it a shot, but we'll need @danroth27 to review it 😄

FabriDamazio:
@Eilon haha no problem with that! I'll do it.

FabriDamazio:
@AWulkan changed `.flex-column` for ` .navbar-nav` as indicated in bootstrap docs.

@Eilon @danroth27 ThirdPartyNotices ready for review

SteveSandersonMS:
@FabriDamazio Thanks so much for this!

The technical aspects of which files you've updated, code style, etc., all look perfect to me. I understand also it was necessary to drop the glyphicons.

The one bit I think could use a bit of further tweaking is the visual styling itself. Examples:

image
image

Rather than making small patches to the exact styling and positioning of elements within the existing structure, do you think we should consider changing a bit more radically?

It looks like the following is a more idiomatic dashboard style for Bootstrap 4: https://getbootstrap.com/docs/4.0/examples/dashboard/ To me it looks much cleaner and more modern than what we already have, with the exception of looking a bit colorless (which we could easily address by making the header bar some color other than gray).

@FabriDamazio If you're interested in switching the templates over to the Bootstrap 4-styled dashboard, that would be great. If not, no worries - we can take your PR here as a starting point and one of us would do the further style updates.

FabriDamazio:
@SteveSandersonMS I agree with your points and I'll improve the templates styles following the given example and show the results in one day or two.

danroth27:
@FabriDamazio Looks really nice! I know this is an asthetic comment, but personally I'd prefer something a bit more muted than the magenta coloring. To me red is pretty strongly associated with errors.

FabriDamazio:
@danroth27 what about something like: 
image

danroth27:
Maybe a teal green?

FabriDamazio:
image

danroth27:
@FabriDamazio That looks good to me.

grahamehorner:
Q: should the templates/samples include the work that has been done on the site templates for constent/GDPR ? update is looking cool 😎

FabriDamazio:
Bootstrap hamburger menu isnt working in v4 either. (see #329 ). 
If i add `jquery.min.js` and `bootstrap.min.js` and remove `bootstrap-native`, everything works. I tried many ways but i can manage it to work with bootstrap-native. Someone try make this works using bootstrap-native or should we drop bootstrap native and use jquery instead?

SteveSandersonMS:
Thanks for trying. We definitely won’t add a jQuery dependency to the default template, but we could remove the hamburger button for now.

FabriDamazio:
@SteveSandersonMS  OK, I'll remove the button and update the PR

AWulkan:
@FabriDamazio Can it be because the JS script for bootstrap loads before the <app></app> content is loaded? Therefore it's not hooking up the html elements with event listeners correctly.

FabriDamazio:
Visual updated to TEAL GREEN example. (Tkx @danroth27, nice color)

@SteveSandersonMS the hamburger button remain untouched (collapsed menu not working) We need more investigation about the causes. I understand we need use bootstrap-native due size and performance.

@AWulkan i tried that solutions but i think you are in the right way to solve it.

I can add icons after SVG support. Just tell me when.

panesofglass:
This looks really nice! Out of curiosity, is there a reason to use bootstrap vs something like [bulma](https://bulma.io/), which doesn't include a JS library? Is bootstrap used b/c of popularity or to also show off JS interop? If the latter, are there any examples in the template showing interop? I don't recall seeing them, but it would be a nice way to show off interop and provide an additional reason for using bootstrap.

AWulkan:
@panesofglass I think the only reason for using Bootstrap is because its popular. Bulma would be more optimal in terms of file size and lack of JS dependencies, but it is not as well known to people.

SteveSandersonMS:
Thanks again @FabriDamazio!

In #619, parts of this PR were merged, along with some alternate styling work needed to account for mobile scenarios and make the hamburger menu work.

PR #397, opened 2018-03-27
Mono Platform MethodHandle Cache

grahamehorner:
added a simple caching of assembly, class & methods to findMethod


src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
```diff
@@ -26,23 +30,32 @@ export const monoPlatform: Platform = {
...
-    }
-
-    const methodHandle = find_method(typeHandle, methodName, -1);
-    if (!methodHandle) {
-      throw new Error(`Could not find method "${methodName}" on type "${namespace}.${className}"`);
-    }
-
-    return methodHandle;
+        var assemblyHandle: number | undefined = registeredAssembly[assemblyName];
+        if (assemblyHandle) {
```
campersau: Shouldn't this be `!assemblyHandle`, `!typeHandle` and `!methodHandle`?

grahamehorner: @campersau yes, your correct this was spotted, but the commit failed as our Internet is currently down :'( and I'm having to use 4G connection at present

SteveSandersonMS:
Thanks for contributing this, @grahamehorner!

There are a few issues in the code here that would have to be resolved before we could merge this. Hope that's OK.

* Some of the tests are broken by this change: https://ci.appveyor.com/project/aspnetci/blazor/build/1.0.431/tests. I suspect the reason is that the caching logic isn't quite correct. It assumes (for example) that all class names are uniquely identified by `"${namespace}.${className}"`, which is not the case (there could be two different assemblies containing different classes with the same namespace and name), and that all methods are uniquely identified by `"${namespace}.${className}.${methodHandle}"`, which is also not the case for the same reason.
* The type annotations on the `registeredAssembly`/`registeredClass`/`registeredMethod` lookups are set to allow entries of type `number | undefined` or `any | undefined`. I think the value types should just be `number` for all three, because there isn't any legal case where we'd store the value `undefined` in the cache.
* The indentation style is inconsistent with the rest of the file, which uses two spaces for indents rather than four (because 2 is becoming more common in the JS/TS world).

If you're able to resolve these issues we'd definitely be happy to merge this PR. Thanks again for contributing!

grahamehorner:
@SteveSandersonMS thanks for your feedback; I will adjust the implementation accordingly

grahamehorner:
@SteveSandersonMS I've changed VS2017 Text Editor defaults for JavaScript/TypeScript, these are now 2 spaces; wondering if its worth letting VS2017 team no that the defaults should be aligned to the common practices used by the JavaScript/TypeScript communities ?

grahamehorner:
closed request as fork had build issues; and was deleted by a dev (not me) :'(
so recreated based from a new fork and updated code AGAIN ;-)

Issue #357, opened 2018-03-24
Question: Send cookies


aguacongas:
Hi Guys,

I try to port a sample application I wrote to Blazor (it's just a chat actualy). 
The application is not hosted by the chat server, I use CORS. The server is hosted on [heroku](https://chatle-server.herokuapp.com).  
In this application, user are able to login throught third parties providers so I need to send authentication to the server on the callback page.

How can I do that ?

The application code is [there](https://github.com/aguacongas/Chatle.Blazor).
The server code is [there](https://github.com/aguacongas/chatle).

floreseken:
looks like this is not possible yet, just as sending extra headers..

floreseken:
Oh wait.. headers are possible:

```
            HttpRequestMessage req = new HttpRequestMessage(HttpMethod.Get, "\api\test");
            req.Headers.Add("auth", "jdfhkjhfd");

            http.SendAsync(req);
```

So you could fix it yourself that way..

aguacongas:
I tried but it doesn't work, I can't read cookies I received.  
But I finaly found a solution, I need to override sendAsync function to set credentials in `fetch` function :  

``` ts
async function sendAsync(id: number, method: string, requestUri: string, body: string | null, headersJson: string | null, credentials: boolean | null) {
    let response: Response;
    let responseText: string;
    try {
        response = await fetch(requestUri, {
            method: method,
            body: body || undefined,
            headers: headersJson ? (JSON.parse(headersJson) as string[][]) : undefined,
            credentials: credentials ? "include" : undefined
        });
        responseText = await response.text();
    } catch (ex) {
        dispatchErrorResponse(id, ex.toString());
        return;
    }

    dispatchSuccessResponse(id, response, responseText);
}
```

Do you agree if I submit a PR with this feature ?  
I plan to add a default option to send credentials in BrowserHttpMessageHandler and add HTTP extensions methods overriding this setting

aguacongas:
You can review the code of [this commit](https://github.com/aguacongas/Blazor/commit/578fa332860575ee9ca916d790d37081f21fe2a9)

aguacongas:
@danroth27 Did you have the time to review that ?

danroth27:
Not yet - will take a look later this week.

SteveSandersonMS:
There's more discussion in #399 so I'll close this issue in favour of that PR.

PR #399, opened 2018-03-27
Implement Request credentials 

aguacongas:
#357 Implements fetch request credentials


SteveSandersonMS:
Thanks for contributing this, @aguacongas! I appreciate the thorough PR with tests.

On consideration, I think I would propose a slightly different way to implement this feature. My concern about the approach here is that it's very specialised:

 * It only covers the case where you're calling one of the `*JsonAsync` methods, and not if you're calling say `GetStringAsync` or just `SendAsync`
 * It only covers setting the `credentials` arg on `fetch`, and only to a set of predefined values. It's likely that people will want to configure other aspects of `fetch` too, which would end up creating an explosion of method overload combinations on the `*JsonAsync` methods.

If I'm understanding the design of `HttpClient` correctly, it appears that `HttpRequestMessage.Properties` is designed as an arbitrary bag of additional `HttpMessageHandler`-specific metadata. As such the approach I'd prefer would be:

* Declare some static readonly `string` on `BrowserHttpMessageHandler` such as `public readonly static string FetchArgs = "BrowserHttpMessageHandler.FetchArgs";`
* Update the interop code so that `BrowserHttpMessageHandler` supplies the value `request.Properties[FetchArgs]` to the JS-side code
* Update the JS-side code so that if the supplied "fetchArgs" value is nonnull, its properties are all added to the options object passed when we call `fetch`

Then, developers can construct a `HttpRequestMessage` and set:

```csharp
myMessage.Properties[BrowserHttpMessageHandler.FetchArgs] = new
{ 
    credentials = "include",
    anyOtherJsProperty = new { myValue = 123 }
};
```

... and then the JS-side code will call:

```js
fetch(url, { ... existing args ..., credentials: 'include', anyOtherJsProperty: { myValue: 123 } })...
```

... which allows a way to control all aspects of `fetch`, including future APIs that we don't even know about yet. It also avoids any arbitrary distinction between `*JsonAsync` and the regular `*Async` methods on `HttpClient`.

Drawback: I know this means you'd need to call `SendAsync` and do your own JSON-serialization, but I don't know of a way to avoid that while remaining consistent with the broader design of `HttpClient`. Developers who commonly need to specify particular custom arguments with their HTTP calls can create their own wrapper around `SendAsync` that takes whatever custom arguments (e.g., `Credentials`) that they need. We can't really put `credentials` directly onto `GetJsonAsync` in the core library here for the reasons described above, i.e., it's not consistent with `HttpClient` design and wouldn't scale to other flags and options that you might want to pass through to `fetch`.

What do you think?

aguacongas:
Hi Steve,

It's look really better actualy, I agree. Can I implement it ?

SteveSandersonMS:
@aguacongas 

Re `const` vs `static` - either is fine here really.

> And does it means the user cannot use GetJsonAsync methods if it need to use some fetch options then

In the general case, it means the user needs to construct a `HttpRequestMessage` and pass it to `SendAsync`. This allows for any HTTP method, any combination of headers, any payload (JSON or otherwise), etc.

I know there will be times where it's not as convenient as using a higher-level method like `GetJsonAsync` but as mentioned above, developers can create their own wrapper methods around this if they have specific needs (such as setting the `credentials` option, or `integrity`, or anything else).

Basically this is a relatively safe low-level design (not going to trap us in the future) that others can build higher-level methods on top of.

aguacongas:
@SteveSandersonMS 

I uptated according to the discussion. I didn't update the test layout but if works for one options it should for others.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
```diff
@@ -7,77 +7,80 @@ const httpClientTypeName = 'BrowserHttpMessageHandler';
 const httpClientFullTypeName = `${httpClientNamespace}.${httpClientTypeName}`;
 let receiveResponseMethod: MethodHandle;
 
-registerFunction(`${httpClientFullTypeName}.Send`, (id: number, method: string, requestUri: string, body: string | null, headersJson: string | null) => {
-  sendAsync(id, method, requestUri, body, headersJson);
+registerFunction(`${httpClientFullTypeName}.Send`, (id: number, method: string, requestUri: string, body: string | null, headersJson: string | null, fetchArgs: RequestInit | null) => {
+    sendAsync(id, method, requestUri, body, headersJson, fetchArgs);
```
SteveSandersonMS: Could you please fix the indentation to be consistent with project conventions for `.ts` files? (e.g., 2 spaces per indentation level)

src/Microsoft.AspNetCore.Blazor.Browser/Http/BrowserHttpMessageHandler.cs
```diff
@@ -22,6 +23,11 @@ public class BrowserHttpMessageHandler : HttpMessageHandler
         static IDictionary<int, TaskCompletionSource<HttpResponseMessage>> _pendingRequests
             = new Dictionary<int, TaskCompletionSource<HttpResponseMessage>>();
 
+        public const string FetchArgs = "BrowserHttpMessageHandler.FetchArgs";
+
+        public BrowserHttpMessageHandler()
+        { }
```
SteveSandersonMS: Redundant empty constructor can be removed

src/Microsoft.AspNetCore.Blazor.Browser/Http/BrowserHttpMessageHandler.cs
```diff
@@ -36,17 +42,31 @@ public class BrowserHttpMessageHandler : HttpMessageHandler
                 _pendingRequests.Add(id, tcs);
             }
 
+            request.Properties.TryGetValue(FetchArgs, out object fetchArgs);
```
SteveSandersonMS: Please use `out var fetchArgs` for consistency with project conventions

src/Microsoft.AspNetCore.Blazor.Browser/Http/BrowserHttpMessageHandler.cs
```diff
@@ -36,17 +42,31 @@ public class BrowserHttpMessageHandler : HttpMessageHandler
...
                 request.RequestUri,
                 request.Content == null ? null : await GetContentAsString(request.Content),
-                SerializeHeadersAsJson(request));
+                SerializeHeadersAsJson(request),
+                fetchArgs);
 
             return await tcs.Task;
         }
 
+        private static string GetDescription(Enum value)
```
SteveSandersonMS: I don't think we need this method any more

test/testapps/TestServer/Controllers/PersonController.cs
```diff
@@ -32,6 +33,23 @@ public IEnumerable<string> Get()
             }
         }
 
+        [HttpPost("xhrf")]
```
SteveSandersonMS: Are these methods `GetAntiForgery` and `PostAntiForgery` used anywhere? Could you clarify their purpose?

test/testapps/TestServer/Startup.cs
```diff
@@ -17,6 +19,13 @@ public Startup(IConfiguration configuration)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddLogging(builder =>
```
SteveSandersonMS: Doesn't seem related to the rest of the PR - please remove

test/testapps/TestServer/Startup.cs
```diff
@@ -26,9 +35,20 @@ public void ConfigureServices(IServiceCollection services)
                         .AllowAnyOrigin()
                         .AllowAnyHeader()
                         .AllowAnyMethod()
+                        .AllowCredentials()
                         .WithExposedHeaders("MyCustomHeader");
                 });
             });
+
+            services.AddAntiforgery(options =>
```
SteveSandersonMS: I'm guessing this is meant to be used from an E2E test somewhere, but there doesn't appear to be one included in the PR. Is there still something to be finished here?

test/testapps/TestServer/Controllers/PersonController.cs
aguacongas: I added these methods to test antiforgery cookies submition, the e2e test is missing.  I'm going to add it

test/testapps/TestServer/Startup.cs
aguacongas: yes the e2e test is missing.

aguacongas: I tried to add an e2e test but cookies aren't recieved when selenium runs the test. If I run the test manualy, it works, but not with selenium. The code is :  

``` csharp
        [Fact]
        public void CanPerformPostRequestWithFetchCredentials()
        {
            new SelectElement(Browser.FindElement(By.Id("request-credentials")))
                    .SelectByValue("include");
            IssueRequest("GET", "/api/person/xhrf");
            Assert.Equal("OK", _responseStatus.Text);
            var token = _responseBody.Text;
            AddRequestHeader("X-XSRF-TOKEN", token);

            IssueRequest("POST", "/api/person/xhrf", token);
            Assert.Equal("OK", _responseStatus.Text);
            Assert.Equal($"You posted: {token}", _responseBody.Text);
        }
``` 

Any idea ?

SteveSandersonMS: I'm not sure about that. Maybe Selenium requires additional config to enable cookies.

Instead of doing an E2E test about XSRF tokens, would it be simpler for the E2E test to cover a simpler `fetch` option such as [`referrer`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch)?

aguacongas: yes it is

SteveSandersonMS:
Thanks @aguacongas for finishing this off! It's now merged.

aguacongas:
@SteveSandersonMS You're welcome. When do you plan to released 0.2.0 ?

danroth27:
> When do you plan to released 0.2.0 ?

We're planning to ship sometime next week.

PR #410, opened 2018-03-28
Json to struct error 388

MatthewLazarow:
Summary of the changes
 - Added 2 new unit tests to prove out [de]serialization to a struct
 - Updated code as indicated in the bug details to resolve issue.

Addresses bug #388

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
@@ -1428,7 +1428,7 @@ namespace SimpleJson
                             obj = value;
                         else
                         {
-                            obj = ConstructorCache[type]();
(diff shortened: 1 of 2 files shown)


MatthewLazarow:
I seem to keep having CI issues caused by npm not installing correctly. I'm not seeing anything obvious with my code changes that could cause this...is it possible something with my visual studio account could cause an issue like this?

SteveSandersonMS:
Thanks @MatthewLazarow - merged!

I did change the implementation slightly though, to ensure it only uses `Activator.CreateInstance` when we know for definite that the target type is a struct and we're looking for a zero-arg constructor: https://github.com/aspnet/Blazor/commit/7fc383ccf9ec13639200d38de506e0a451520304#diff-b074a84659130e6e3102450892b52562R1855

It still passes your tests. As far as I know this should still be suitable for your needs, but please let me know if you have any concerns that either this is insufficient or incorrect in any case. Thanks!

PR #426, opened 2018-03-29
Mono Platform Method Caching

grahamehorner:
simple caching of assembly, class, methods with changes as recommended by @SteveSandersonMS


src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
```diff
@@ -26,22 +29,31 @@ export const monoPlatform: Platform = {
   },
 
   findMethod: function findMethod(assemblyName: string, namespace: string, className: string, methodName: string): MethodHandle {
-    // TODO: Cache the assembly_load outputs?
-    const assemblyHandle = assembly_load(assemblyName);
+
+    var assemblyHandle: number | undefined = registeredAssemblies[assemblyName];
```
SteveSandersonMS: Could we use `let` rather than `var`? In general `var` can always be replaced by `let` or `const`, depending on whether the value needs to change.

SteveSandersonMS: Also below

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
```diff
@@ -26,22 +29,31 @@ export const monoPlatform: Platform = {
...
-      throw new Error(`Could not find assembly "${assemblyName}"`);
+      assemblyHandle = assembly_load(assemblyName);
+      if (!assemblyHandle) {
+        throw new Error(`Could not find assembly "${assemblyName}"`);
+      }
+      registeredAssemblies[assemblyName] = assemblyHandle;
     }
-
-    const typeHandle = find_class(assemblyHandle, namespace, className);
+    var typeHandle: number | undefined = registeredClasses["[${assemblyName}]${namespace}.${className}"];
```
SteveSandersonMS: This is broken because it uses the wrong quote style. Should be backtick, not double-quote, so as to enable string interpolation.

SteveSandersonMS: Also below.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
grahamehorner: const failed as the value changes; sorry I'm a noob to typescript/javascript and learning as I implement ;) I will change the var to let

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
grahamehorner: oops my bag; noob error

grahamehorner:
@SteveSandersonMS I'm also still having issues with the E2E tests failing when using the source repo; even though I have managed to get selenium-standalone installed and running after the update/re-installation of

RyoukoKonpaku:
@grahamehorner try checking the chrome driver version, I got errors running the E2E last time as well due to my selenium version and it's chrome driver being incompatible with my chome version installed.

grahamehorner:
@RyoukoKonpaku sorry I'm a complete noob to this client side javascript/typescript stuff, learning all the time, where would I find what versions etc. out work together; I've simply just installed latest of chrome and selenium as shown in the perquisites

RyoukoKonpaku:
@grahamehorner try opening this on file explorer ```%appdata%\npm\node_modules\selenium-standalone\.selenium\chromedriver``` it should show what chrome driver you are using. If you're using the latest version of chrome right now it should be chrome driver 2.36 not 2.33 since 2.33 is incompatible with the latest chrome version.

grahamehorner:
@SteveSandersonMS idea for a better caching solution; I feel the caching of methods should be at the platform abstraction layer and not no the platform layer it's self; also I'm thinking about possible pre-population of the cache using a promise at start-up. what would your thoughts be? I understand this is possibly away of into the future as the only WASM platform currently supported in mono; just ideas and thinking aloud.

grahamehorner:
@RyoukoKonpaku thanks

29/03/2018  15:30    <DIR>          .
29/03/2018  15:30    <DIR>          ..
29/03/2018  15:30         6,406,144 2.36-x64-chromedriver
29/03/2018  15:30         3,357,751 2.36-x64-chromedriver.zip

Google Chrome is up to date
Version 65.0.3325.181 (Official Build) (64-bit)

RyoukoKonpaku:
@grahamehorner hmm seems chrome driver and selenium is up to date, mind if what does the cmd show when the E2E test fails?

grahamehorner:
@RyoukoKonpaku I've captured the console out
[e2e-build.log](https://github.com/aspnet/Blazor/files/1863869/e2e-build.log)
put of the build with E2E tests in the attached file.

RyoukoKonpaku:
@grahamehorner That's odd seems Selenium is timing out. particularly here in this method ```WaitUntilMonoRunningInBrowser``` on ```MonoSanityTest.cs``` which waits around 30 seconds to let mono load up and in other related places where it waits for Blazor to setup. 
I'm not too sure what's causing this though, but I think Mono or Blazor isn't starting up properly where selenium just times out due to waiting for the app to startup.

SteveSandersonMS:
This is merged. Thanks, @grahamehorner!

I also made some further refactorings to reduce the number of cache lookups needed in typical cases (e.g., when finding a method that was already cached earlier, it's not necessary to locate the assembly or type).

PR #443, opened 2018-03-30
Double-quote attribute values

guardrex:
Singles are perfectly valid under HTML5 ... but inconsistent with our doc samples.

Why do I have the sneaky feeling this was done on purpose to solve a problem?? :smile:

If so, this is a throw-away PR. 🗑 Sorry for the trouble.


SteveSandersonMS:
Thanks - that is more consistent!

The single quotes are just a hangover from the original copy-paste from our other SPA templates.

PR #444, opened 2018-03-30
Add conditional attributes and hash-join for attributes

rynowak:


rynowak:
/cc @SteveSandersonMS - here's a possible approach for implementing conditional attributes.

Interested in your feedback about whether this is the optimal way to proceed. I haven't added unit tests for the changes yes, waiting on understanding of the design implications.

This approach implements the 'conditionality' inside the render tree builder, to try and keep the presence/absence of attributes reflected in the tree.

**pro:** keeps knowledge of .NET's types in .NET code
**pro:** the tree reflects more the presence/absence of attributes (it just feels right to me)
**con:** look at the changes required to NavLink. I'm not sure if this sort of thing is intended but it seems apparent to me that if we're going to put any significant logic in `RenderTreeBuilder` then component implementors that override rendering will have to *do the right thing*. This didn't seem to be an issue in the past, a naively written component would function pretty well.

Your thoughts?

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
```diff
@@ -97,45 +97,90 @@ public void AddContent(int sequence, RenderFragment fragment)
...
+                }
+                else if (value is UIEventHandler eventHandler)
+                {
+                    Append(RenderTreeFrame.Attribute(sequence, name, eventHandler));
+                }
+                else
+                {
+                    // Element attribute values can only be strings or UIEventHandler
+                    Append(RenderTreeFrame.Attribute(sequence, name, value.ToString()));
+                }
```
rynowak: These cases weren't handled before. If you had a weakly typed event handler, it would all go wrong for you.

src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
```diff
@@ -100,19 +100,20 @@ private bool ShouldMatch(string currentUriAbsolute)
...
-            foreach (var kvp in _allAttributes.Where(kvp => kvp.Key != "class"))
+            foreach (var kvp in _allAttributes.Where(kvp => kvp.Key != "class" && kvp.Key != RenderTreeBuilder.ChildContent))
             {
-                builder.AddAttribute(0, kvp.Key, kvp.Value);
+                builder.AddAttribute(i++, kvp.Key, kvp.Value);
             }
 
             // Pass through any child content unchanged
-            builder.AddContent(1, _childContent);
+            builder.AddContent(i++, _childContent);
```
rynowak: I'm not 100% certain if this is intended to work or not 😆 - I think that's the most important factor for how we proceed.

src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
```diff
@@ -100,19 +100,20 @@ private bool ShouldMatch(string currentUriAbsolute)
...
+            var i = 0;
+            builder.OpenElement(i++, "a");
 
             // Set "active" class dynamically
-            builder.AddAttribute(0, "class", CombineWithSpace(_cssClass, _isActive ? "active" : null));
+            builder.AddAttribute(i++, "class", CombineWithSpace(_cssClass, _isActive ? "active" : null));
 
             // Pass through all other attributes unchanged
-            foreach (var kvp in _allAttributes.Where(kvp => kvp.Key != "class"))
+            foreach (var kvp in _allAttributes.Where(kvp => kvp.Key != "class" && kvp.Key != RenderTreeBuilder.ChildContent))
```
rynowak: This was just a bug, the dom would contain an attribute for the child content with its type name as the value.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
SteveSandersonMS: Ah, great - well spotted!

src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
SteveSandersonMS: It's not necessary (or shouldn't be) to auto-generate the sequence numbers here. You can keep the sequence number as `0` for the element, and `1` for all the attributes.

Reason:

* The diff algorithm uses the sequence numbers as a basis for a merge-join on sibling nodes. It's slightly more specialised than a normal SQL-style merge join because using the sequence numbers, it can also identify if/else branches and loops. The sequence numbers let us reduce the diff to a linear O(N) operation (versus O(N^3) for general tree diff), and cause the edit script to match very closely the groupings in the original source code (e.g., blocks surrounded by `if`/`else` are atomic). 
* However, the handling for attributes on a given node is different. It's a simpler hash join instead of the loop-aware merge join, because we don't care about the final output order (attributes are independent of order), plus attributes inherently come with string keys (attribute name) and the total number of attributes tends to be small (<10, say). The hash join doesn't compare relative sequence numbers at all. So for attributes, it's fine just to give all of them on a given node the same value. Also it's not harmful to give them distinct values, so we just do whatever's most convenient.

I'm totally aware that the rules here are complex, and we can't expect most users to know about them. People will definitely try to auto-generate sequence numbers at runtime, even though AFAIK there's no scenario where runtime execution order is relevant - it's only source code instruction order that's relevant.

What I'm hoping is that:

1. Most devs will write render methods in the form of Razor files exclusively, so they don't even need to know that sequence numbers exist. I'd hope that even `NavLink` could be written purely in Razor once we get further with #1.
2. For devs building more advanced components who really have to bypass Razor, if they get the sequence number generation wrong, it will still produce a correct result at runtime, even if the diff algorithm produces suboptimal edit scripts.
3. For devs building really advanced components who also need the edit scripts to be optimal, they will have to understand the rules.

What do you think?

src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
SteveSandersonMS: Ah yes, I think I saw that at some point. Glad you squashed that one!

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
```diff
@@ -97,45 +97,90 @@ public void AddContent(int sequence, RenderFragment fragment)
...
+        /// </summary>
+        /// <param name="sequence">An integer that represents the position of the instruction in the source code.</param>
+        /// <param name="name">The name of the attribute.</param>
+        /// <param name="value">The value of the attribute.</param>
+        public void AddAttribute(int sequence, string name, bool value)
+        {
+            AssertCanAddAttribute();
+            if (value || _lastNonAttributeFrameType == RenderTreeFrameType.Component)
+            {
+                Append(RenderTreeFrame.Attribute(sequence, name, true));
```
SteveSandersonMS: Is this going to create a separate boxed `bool` value on every invocation? If so, would we be able to replace the `true` literal here with something like `"true"` so it's the same reference-type instance on every invocation?

SteveSandersonMS:
This looks great to me!

> keeps knowledge of .NET's types in .NET code

Excellent. That's important to avoid any chatty back-and-forth communication between .NET and JS (e.g., to query the type of something). We really want the render batch to contain all the necessary information up front.

> the tree reflects more the presence/absence of attributes (it just feels right to me)

Agreed that's good.

> it seems apparent to me that if we're going to put any significant logic in RenderTreeBuilder then component implementors that override rendering will have to do the right thing.

Once we revert the auto-generated sequence numbers, is that still a concern? It looks to me like it will be no more complex than it was before. I'm assuming you didn't find it was strictly necessary to put in the auto-generated sequence numbers, though if you did, that suggests there might be some other issue we should investigate.

src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
rynowak: Thanks for the explanation. If I understand correctly, then there's a bug that I need to investigate then with that you're describing.

Is this expected to work when the element, attributes and content **all** share the sequence number `0`?

Or the numbering then look like:
```
            var i = 0;
            builder.OpenElement(i++, "a");

            // Set "active" class dynamically
            builder.AddAttribute(i, "class", CombineWithSpace(_cssClass, _isActive ? "active" : null));

            // Pass through all other attributes unchanged
            foreach (var kvp in _allAttributes.Where(kvp => kvp.Key != "class" && kvp.Key != RenderTreeBuilder.ChildContent))
            {
                builder.AddAttribute(i, kvp.Key, kvp.Value);
            }

            // Pass through any child content unchanged
            builder.AddContent(++i, _childContent);

            builder.CloseElement();
```

What I was seeing without any code changes to this class is that *all** of the attributes would removed on the second render if the link was not active.

That looks like this.... Initial render
```
<a class="active" href="/Home" childcontent"Microsoft....">
  ...child content...
</a>
```

And then, I navigate to counter, now link is no longer active, and it renders as:
```
<a>
  ...child content...
</a>
```

So once the render encountered the 'remove' attribute edit for `class`, it was also convinced that the other attributes should be removed as well.

We can discuss at the sync up if its not obvious what I'm talking about 😆

SteveSandersonMS: Ah wait, I was misremembering and mixing up how the original prototype diffing worked compared with how the current one does.

In the current implementation, we *do* use the source sequences for attributes. It's the same function (`AppendDiffEntriesForRange`) that processes all ranges of sibling nodes, where siblings are "nodes with the same parent" and "attributes of the same node".

However I'm still surprised that giving them all the same sequence number causes them to disappear. I'd expect it to produce a valid but not necessarily optimal sequence number. Let's discuss later.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
rynowak: Fixed this by using cached statics of the boxed values.

This removes the boxing as well for components, since we'd always box in those cases.

rynowak:
Updated and added unit tests. Took at stab at fixing the issues around attributes. Looking for feedback on my assumptions about how the attribute diff is intended. I took what I thought was a pretty simple path to solving the problem, but I might be missing something.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -138,6 +139,57 @@ public static void DisposeFrames(RenderBatchBuilder batchBuilder, ArrayRange<Ren
...
+                {
+                    // No match in the new tree, remove old attribute
+                    RemoveOldFrame(ref diffContext, i);
+                }
+            }
+
+            foreach (var kvp in diffContext.AttributeDiffSet)
+            {
+                // No match in the old tree
+                InsertNewFrame(ref diffContext, kvp.Value);
```
SteveSandersonMS: Don't we also need to remove the entry from `AttributeDiffSet` here to ensure it's empty at the start of the next pass? Or would it be easier to reason about if we just cleared it at the start of this method?

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -219,7 +271,7 @@ private static int NextSiblingIndex(RenderTreeFrame frame, int frameIndex)
                             var newFrameAttributesEndIndexExcl = GetAttributesEndIndexExclusive(newTree, newFrameIndex);
 
                             // Diff the attributes
-                            AppendDiffEntriesForRange(
+                            AppendAttributeDiffEntriesForRange(
```
SteveSandersonMS: Regarding perf, I haven't measured it, but do you have a sense of how the old and new mechanisms would compare in the case where the attributes are unchanged? I'd think that 95%+ of the time, the attribute names are at least unchanged (same set in the same order). Conditional or procedurally-generated attribute names are relatively rare.

For the old method, I expected (but didn't verify) that in virtually all cases the attribute name comparisons would reduce down to reference equality checks without needing to iterate through the characters of the strings, since they'd come from the same compile-time constants. Whereas for the new method it will need to compute a hash code for each attribute name, which at least involves iterating through all the characters.

I know there's a risk of over-thinking the optimisation here, but given that in the vast majority of cases the set of attribute names will be unchanged, do you think it might be worth doing a first-pass check here where we establish whether the set of attribute names are unchanged (by looping through them, expecting the string comparisons to be reference-equality checks in most cases), and if they are, falling back on the earlier `AppendDiffEntriesForRange` logic to avoid the hashcode calculations?

src/Microsoft.AspNetCore.Blazor/Rendering/RenderBatchBuilder.cs
```diff
@@ -27,6 +27,12 @@ internal class RenderBatchBuilder
         public Queue<RenderQueueEntry> ComponentRenderQueue { get; } = new Queue<RenderQueueEntry>();
         public Queue<int> ComponentDisposalQueue { get; } = new Queue<int>();
 
+        // Scratch data structure for understanding attribute diffs.
+        //
+        // One of these can be used for the whole tree diff because attributes cannot be nested.
+        // Does not need to be cleared because it is always emptied at the end of each attribute pass.
```
SteveSandersonMS: What do you think about the risk that on an earlier pass, it didn't get emptied due to being halted by an exception? Should we say that behavior is undefined from that point on, or would it be preferable to empty the dict at the start, since it should be nearly free anyway if it's already empty?

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
SteveSandersonMS: Or a hybrid like where we start with the old approach, but switch to the new one when we find the first changed attribute name:

    for(var index = start; index < endExcl; index++)
    {
        if (oldAttributeName[index] == newAttributeName[index])
        {
            AppendDiffEntriesForFramesWithSameSequence(... index ...);
        }
        else
        {
            // Something's changed. Process the remaining attributes using the hash lookups
            AppendAttributeDiffEntriesForRange(index, endExcl, ...);
            break;
        }    
    }

SteveSandersonMS: Actually that approach might get more messy when the two sets have different numbers of entries. I'll stop speculating now as I'm sure you will have a view on how you'd prefer to handle this, if at all.

rynowak: I'll take a look at this and see if it makes a significant difference. 

One idea that I had is to do a fast path for single attributes since a single entry can't ever be in a different order 👍 

If your main concern is the cost of hashing, we could build a really simple equality comparer that just uses the first character or length. I'm interested in trying this out because I want to see if this is useful for routing 😆

Note, we still need to diff the attribute values in all cases.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
rynowak: Oh oopsies. This is a bug.

src/Microsoft.AspNetCore.Blazor/Rendering/RenderBatchBuilder.cs
rynowak: That's an interesting thought. Since we don't expect anything during the diff pass to *handle* the exception, I propose that we put a 'reset' at the top level in a try/finally. Then it's a once-per-pass cost.

rynowak:
If fixed the 2 problems we discussed, and added a benchmark project so we have a real perf test. Let me know if you think this is a reasonable scenario for analysis and I'll follow up on that.

rynowak:
> added a benchmark project so we have a real perf test

Actually I should say more. I started by adding perf test (can add more if we deem it necessary) so we can measure this in a standardized, repeatable way. Let me know if you have feedback on the structure of the test, I kept it simple to avoid solving too many problems at once.

Next, I plan to profile the test and see what the contribution of the new logic is to the overall perf, and make a judgement from there.

I think we'll want to expand the set of tests at some point to include some compiled Razor code as well, which hopefully won't be too hard. Once we know have some good tests for the areas that are important we can get help from some of tuning whizzes that have been busy in the common repo.

rynowak:
So the answer here is that the overhead of the dictionary is super significant compared to the existing code:
```
![image](https://user-images.githubusercontent.com/1430011/38227491-5007d068-36b3-11e8-8619-735c779638f9.png)
```

rynowak:
So, it turns out that your intuition was correct. The dictionary has a really significant effect on the overall perf.

After seeing this I added a fast path for merge join for attributes based on sequence numbers + equality of the names. 

Results:
```
``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.309)
Intel Core i7 CPU 950 3.07GHz (Nehalem), 1 CPU, 8 logical cores and 4 physical cores
Frequency=3030835 Hz, Resolution=329.9421 ns, Timer=TSC
.NET Core SDK=2.1.300-preview2-008322
  [Host]     : .NET Core 2.0.6 (CoreCLR 4.6.26212.01, CoreFX 4.6.26212.01), 64bit RyuJIT
  Job-COYIZC : .NET Core 2.1.0-preview2-26308-01 (CoreCLR 4.6.26307.01, CoreFX 4.6.26308.02), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 2.1  
RunStrategy=Throughput  

```
|                                                                Method |     Mean |     Error |    StdDev |      Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|---------------------------------------------------------------------- |---------:|----------:|----------:|----------:|-------:|---------:|-------:|----------:|
| &#39;RenderTreeDiffBuilder: Input and validation on a single form field.&#39; | 3.270 us | 0.0640 us | 0.1051 us | 305,803.4 |   1.00 |     0.00 | 0.0191 |    1.8 KB |
|                              &#39;RenderTreeDiffBuilder: With Fast Path.&#39; | 1.527 us | 0.0178 us | 0.0139 us | 654,685.8 |   0.47 |     0.02 | 0.0153 |   1.42 KB |

```

One issue with this is that the original case I hit - where `NavLink` has conditional attributes without unique sequence numbers will go down the slow path. I don't have any great ideas about how to fix that, but if we think it's going to be common it might be worth figuring out.

Thoughts?

I would be great if you could give the new code a look over as well.

**Comment by rynowak** [comment]
In case you're curious, this is what a profile of this test looks like - not sure if you're measured this code in the past, but there's no obvious low-hanging fruit.

image

**Comment by rynowak** [comment]
Here's more data with outlining of the attribute case from `AppendDiffEntriesForFramesWithSameSequence` - I think it's a small improvement, but I think it's a good change in general since attributes mostly have their own code path

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.309)
Intel Core i7 CPU 950 3.07GHz (Nehalem), 1 CPU, 8 logical cores and 4 physical cores
Frequency=3030835 Hz, Resolution=329.9421 ns, Timer=TSC
.NET Core SDK=2.1.300-preview2-008322
  [Host]     : .NET Core 2.0.6 (CoreCLR 4.6.26212.01, CoreFX 4.6.26212.01), 64bit RyuJIT
  Job-KJRAYN : .NET Core 2.1.0-preview2-26308-01 (CoreCLR 4.6.26307.01, CoreFX 4.6.26308.02), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 2.1  
RunStrategy=Throughput  

```
|                                                                Method |     Mean |     Error |    StdDev |      Op/s | Scaled | ScaledSD | Allocated |
|---------------------------------------------------------------------- |---------:|----------:|----------:|----------:|-------:|---------:|----------:|
| &#39;RenderTreeDiffBuilder: Input and validation on a single form field.&#39; | 2.980 us | 0.0090 us | 0.0071 us | 335,582.8 |   1.00 |     0.00 |      40 B |
|                              &#39;RenderTreeDiffBuilder: With Fast Path.&#39; | 1.546 us | 0.0413 us | 0.0387 us | 647,028.7 |   0.52 |     0.01 |      40 B |
|                &#39;RenderTreeDiffBuilder: With Fast Path and outlining.&#39; | 1.475 us | 0.0421 us | 0.0591 us | 678,114.6 |   0.49 |     0.02 |      40 B |

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -284,46 +404,39 @@ private static int NextSiblingIndex(RenderTreeFrame frame, int frameIndex)
...
-                                }
-                                InitializeNewAttributeFrame(ref diffContext, ref newFrame);
-                                var referenceFrameIndex = diffContext.ReferenceFrames.Append(newFrame);
-                                diffContext.Edits.Append(RenderTreeEdit.SetAttribute(diffContext.SiblingIndex, referenceFrameIndex));
-                            }
-                            else if (oldFrame.AttributeEventHandlerId > 0)
-                            {
-                                // Retain the event handler ID
-                                newFrame = oldFrame;
-                            }
```
rynowak: @SteveSandersonMS  - this block:
```
else if (oldFrame.AttributeEventHandlerId > 0)
{
...
}
```

I'm confused by this, what's it supposed to be doing? It looks to me like it's just here to prevent falling into the `else`. Now that I've pulled this out into its own method, nether the `else if` nor the `else` seems necessary.

let me know if I'm missing something here.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -149,19 +149,84 @@ public static void DisposeFrames(RenderBatchBuilder batchBuilder, ArrayRange<Ren
...
+                    AppendDiffEntriesForAttributeFrame(ref diffContext, oldStartIndex, newStartIndex);
+
+                    oldStartIndex = NextSiblingIndex(oldTree[oldStartIndex], oldStartIndex);
+                    newStartIndex = NextSiblingIndex(newTree[newStartIndex], newStartIndex);
+                    hasMoreOld = oldEndIndexExcl > oldStartIndex;
+                    hasMoreNew = newEndIndexExcl > newStartIndex;
+                }
+                else if (oldSeq < newSeq)
+                {
+                    // An attribute was removed from the new sequence.
```
SteveSandersonMS: Typo? I think this should say removed from the *old* sequence

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -339,46 +404,39 @@ private static int NextSiblingIndex(RenderTreeFrame frame, int frameIndex)
...
-                            }
-                            else if (oldFrame.AttributeEventHandlerId > 0)
-                            {
-                                // Retain the event handler ID
-                                newFrame = oldFrame;
-                            }
-                        }
-                        else
-                        {
-                            // Since this code path is never reachable for Razor components (because you
```
SteveSandersonMS: I agree that this `else` block can be removed

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -149,19 +149,84 @@ public static void DisposeFrames(RenderBatchBuilder batchBuilder, ArrayRange<Ren
...
+            // in the old and new sequences. It will be handled correct, but will generate a suboptimal
+            // diff.
+            var hasMoreOld = oldEndIndexExcl > oldStartIndex;
+            var hasMoreNew = newEndIndexExcl > newStartIndex;
+            var oldTree = diffContext.OldTree;
+            var newTree = diffContext.NewTree;
+
+            while (hasMoreOld || hasMoreNew)
+            {
+                var oldSeq = hasMoreOld ? oldTree[oldStartIndex].Sequence : int.MaxValue;
```
SteveSandersonMS: I know it's only a tiny improvement, but to avoid the duplicated array lookups, we could have at the start of the loop:

```csharp
ref var oldItem = ref oldTree[oldStartIndex];
ref var newItem = ref newTree[newStartIndex];
```

It probably makes no difference for JIT-enabled runtimes, but may do for non-JIT ones that we're targeting.

SteveSandersonMS: No, forget that, it makes it more complex given that we might have reached the end of one of the arrays already.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
SteveSandersonMS: Ah, I see. It is needed, and apparently we're missing unit test coverage for it.

The reason is that for each .NET-side event handler, there's a corresponding entry in a JS-side data structure. This keeps track of how to route incoming events to the correct .NET delegate. We have logic for "disposing" the JS-side entries when the .NET-side ones are removed (they are tracked in `diffContext.BatchBuilder.DisposedEventHandlerIds`).

So, if you're preserving an event handler (i.e., same delegate instance) then we don't want the overhead of notifying the JS side that it should dispose the old entry and initialise a new one. We can just not mention it to the JS side at all. That's what this `else if` block was achieving. Since `oldFrame` and `newFrame` are `ref var`, the assignment on line 363 is writing to the underlying array.

Here's an example of a unit test we could add that does pass with the old code but would fail if this assignment was removed:

```csharp
[Fact]
public void PreservesEventHandlerIdsForRetainedEventHandlers()
{
    // Arrange
    UIEventHandler retainedHandler = _ => { };
    oldTree.OpenElement(0, "My element");
    oldTree.AddAttribute(1, "will remain", retainedHandler);
    oldTree.CloseElement();
    newTree.OpenElement(0, "My element");
    newTree.AddAttribute(1, "will remain", retainedHandler);
    newTree.CloseElement();

    // Act
    var (result, referenceFrames) = GetSingleUpdatedComponent(initializeFromFrames: true);
    var oldAttributeFrame = oldTree.GetFrames().Array[1];
    var newAttributeFrame = newTree.GetFrames().Array[1];

    // Assert
    Assert.Empty(result.Edits);
    AssertFrame.Attribute(oldAttributeFrame, "will remain", retainedHandler);
    AssertFrame.Attribute(newAttributeFrame, "will remain", retainedHandler);
    Assert.NotEqual(0, oldAttributeFrame.AttributeEventHandlerId);
    Assert.Equal(oldAttributeFrame.AttributeEventHandlerId, newAttributeFrame.AttributeEventHandlerId);
}
```

Would you be able to add this unit test or something along these lines to `RenderTreeDiffBuilderTest.cs` (and of course updating the new implementation to retain that behavior)?

SteveSandersonMS:
Thanks for the updates - this is looking great!

> One issue with this is that the original case I hit - where NavLink has conditional attributes without unique sequence numbers will go down the slow path.

I'm not personally concerned by that. It would only affect the `NavLink` instances that are changing from "active" to "inactive" (or vice-versa), and in the most typical case there would only be a couple of those in the app at any time (one becoming active, another becoming inactive). People aren't generally going to have thousands of `NavLink` instances all becoming active or inactive at the same time. And if this really did bother us, we could change `NavLink` to always have the same number of attributes, but just toggle the `class` value between `active` and `inactive`. I'd expect it to be totally fine as-is already.

The only thing I think that remains to be done here is preserving the event handler IDs (see comment above). Other than that 👍

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
rynowak: Fixed

Issue #441, opened 2018-03-29
Add Blazor templates for ASP.NET Core 2.1


danroth27:
Looks like there is quite a bit of drop off in the number of people that successfully create a Blazor template after getting the VSIX. One hypothesis is that they can't find the templates. Currently the Blazor templates only show up if you select .NET Core and ASP.NET Core 2.0 in the "New ASP.NET Core Web Application" dialog. It would help if the Blazor templates showed up when the .NET Framework is selected and for ASP.NET Core 2.1.

danroth27:
Add additional `framework` option for 2.0, and the associated conditions to pull in the right package versions in your csproj.

For full framework, the template needs to:
1. Define `TargetFrameworkOverride`: https://github.com/dotnet/templating/blob/d18c42fe62ca500bfd1afbaa7a6382180dff52ef/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-CSharp/.template.config/template.json#L33
1. Hide it from the CLI: https://github.com/dotnet/templating/blob/d18c42fe62ca500bfd1afbaa7a6382180dff52ef/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-CSharp/.template.config/dotnetcli.host.json#L4
1. Support it in the template content: https://github.com/dotnet/templating/blob/d18c42fe62ca500bfd1afbaa7a6382180dff52ef/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-CSharp/Company.ClassLibrary1.csproj#L4-L5.

chrissainty:
@danroth27 I'm happy to have a go at this one. Is it just a case of updating the various json files in the .template.config folders for each of the two templates and the .csproj files?

danroth27:
We think so, but we'll have to test to make sure. Feel free to have a go at it!

chrissainty:
I'm about there with the changes. I've just got one question. In the [BlazorHosted.CSharp.Server.csproj](https://github.com/aspnet/Blazor/blob/512a15eb74514d87c3b84abb1a5d77c2e45b4389/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Server/BlazorHosted.CSharp.Server.csproj#L4) file the project is targeting netstandard2.0 and not netcoreapp2.0 is there a specific reason for this?

[...]

PR #465, opened 2018-04-02
Updated Blazor project templates to show in additional places

chrissainty:
Add Blazor templates for ASP.NET Core 2.1. Addresses [#441](https://github.com/aspnet/Blazor/issues/441)

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Server/BlazorHosted.CSharp.Server.csproj
+++ b/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Server/BlazorHosted.CSharp.Server.csproj
@@ -1,10 +1,16 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework Condition="'$(TargetFrameworkOverride)' == ''">netcoreapp2.1</TargetFramework>
+    <TargetFramework Condition="'$(TargetFrameworkOverride)' != ''">TargetFrameworkOverride</TargetFramework>
   </PropertyGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkOverride)' == ''">
+    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.0-preview1-final" />
+    <PackageReference Include="Microsoft.AspNetCore.Blazor.Server" Version="0.2.0-preview1-10152" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkOverride)' != ''">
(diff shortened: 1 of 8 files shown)


SteveSandersonMS:
@phenning Would you mind taking a look? Does this look right to you?

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Server/BlazorHosted.CSharp.Server.csproj
```diff
@@ -1,10 +1,16 @@
...
   </PropertyGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkOverride)' == ''">
+    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.0-preview1-final" />
+    <PackageReference Include="Microsoft.AspNetCore.Blazor.Server" Version="0.2.0-preview1-10152" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkOverride)' != ''">
     <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
```
phenning: Shouldn't this also be 2.1.0-preview1-final instead of 2.0.0?

chrissainty: Yes, but there is now a preview 2 final should I update to that instead?

phenning: Depends on what the team wants to target, I'll leave that to @danroth27  and @SteveSandersonMS

Also, this PR would have Blazor showing when 2.1 is selected in the dialog, but not 2.0.
I think you would want to add 2.0 as an additional choice in template.json and add the associated conditions in the csproj file.

Note that all the other templates in the aspnet/templating repo are a bit different because they ship with a given CLI, so they only have one framework choice.

danroth27: We'd like the Blazor templates to be available for .NET Core & .NET Framework when using ASP.NET Core 2.0 or ASP.NET Core 2.1 (latest preview)

SteveSandersonMS: @phenning Thanks for the feedback! Based on your comments, I've implemented a different version of this in #545.

SteveSandersonMS:
It turns out that the changes we need are quite a bit simpler than this, so closing this PR in favour of #545. Thanks for contributing though, @chrissainty!

PR #467, opened 2018-04-02
Trigger live reload feature after successful builds in VS

SteveSandersonMS:
This is all very unfamiliar to me, so I'm mainly looking for feedback from @rynowak about whether this is the right sort of approach. Especially all the stuff with `JoinableTaskFactory` and `ProjectLockService`.

### DIFF
--- a/tooling/Microsoft.VisualStudio.BlazorExtension/BlazorPackage.cs
+++ b/tooling/Microsoft.VisualStudio.BlazorExtension/BlazorPackage.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Runtime.InteropServices;
 using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
 
 namespace Microsoft.VisualStudio.BlazorExtension
 {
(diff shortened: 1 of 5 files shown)


tooling/Microsoft.VisualStudio.BlazorExtension/BlazorPackage.cs
```diff
@@ -4,15 +4,32 @@
...
+        }
+
+        private void RegisterLiveReloadBuildWatcher()
+        {
+            // No need to unadvise, as this only happens once anyway
+            ThreadHelper.ThrowIfNotOnUIThread();
+            var buildManager = (IVsSolutionBuildManager)GetService(typeof(SVsSolutionBuildManager));
+            var hr = buildManager.AdviseUpdateSolutionEvents(new LiveReloadBuildWatcher(), out var cookie);
+            Marshal.ThrowExceptionForHR(hr);
+        }
```
rynowak: This all seems right so far.

tooling/Microsoft.VisualStudio.BlazorExtension/LiveReloadBuildWatcher.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public int UpdateSolution_Done(int fSucceeded, int fModified, int fCancelCommand)
+        {
+            _isListeningForProjectBuilds = false;
+
+            foreach (var fullPath in _signalFilePathsToNotify)
+            {
+                File.WriteAllText(fullPath, string.Empty);
+                File.Delete(fullPath);
+            }
```
rynowak: You should probably add some error handling and/or retries here. If you have any sort of failures here like a long path, permissions issue, file locking, etc I think this will crash VS.

SteveSandersonMS:
Thanks - merged.

Issue #376, opened 2018-03-26
On linux UseBlazor extension method fails with "The path must be absolute"


paulvanbladel:
I'm using the FlightFinder sample for reproducing the problem.
https://github.com/aspnet/samples
or my fork: https://github.com/paulvanbladel/samples

steps to reproduce:
1. from the rootfolder of the solution: 
```
dotnet publish FlightFinder.Server/FlightFinder.Server.csproj -c release -o publish
```
2. from the publish folder inside server project:
```
dotnet FlightFinder.Server.dll
```
Unfortunately, this bug makes that blazor can't be used with docker (didn't try with windows images).
```
paul@ubuntu-desktop:~/github/samples/samples/aspnetcore/blazor/FlightFinder/FlightFinder.Server/publish$ dotnet FlightFinder.Server.dll 
info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[0]
      User profile is available. Using '/home/paul/.aspnet/DataProtection-Keys' as key repository; keys will not be encrypted at rest.

Unhandled Exception: System.ArgumentException: The path must be absolute.
Parameter name: root
   at Microsoft.Extensions.FileProviders.PhysicalFileProvider..ctor(String root, PhysicalFilesWatcher physicalFilesWatcher)
   at Microsoft.AspNetCore.Builder.BlazorAppBuilderExtensions.UseBlazor(IApplicationBuilder applicationBuilder, BlazorOptions options)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.AspNetCore.Hosting.ConventionBasedStartup.Configure(IApplicationBuilder app)
   at Microsoft.AspNetCore.Hosting.Internal.AutoRequestServicesStartupFilter.<>c__DisplayClass0_0.<Configure>b__0(IApplicationBuilder builder)
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.Build()
   at FlightFinder.Server.Program.Main(String[] args) in /home/paul/github/samples/samples/aspnetcore/blazor/FlightFinder/FlightFinder.Server/Program.cs:line 14
Aborted (core dumped)

``

urvzh:
[...]

PR #473, opened 2018-04-03
Resolution for "The path must be absolute" on published projects

Daddoon:
Trying to replace the internal "." folder with the local assembly path in order to have an absolute path for the PhysicalFileStorage class at startup ("The path must be absolute" error at startup for published projects)

May resolve the issue of #376  on dev

_Also apology if (maybe) you see something related to me in commit history on aspnet/Blazor (internally ?), i'm pretty new to Github and Git._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ launchSettings.json
 artifacts/
 msbuild.binlog
 .vscode/
-src/Microsoft.AspNetCore.Blazor.Browser.JS/package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ launchSettings.json
 artifacts/
 msbuild.binlog
 .vscode/
+src/Microsoft.AspNetCore.Blazor.Browser.JS/package-lock.json
--- a/src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
@@ -29,6 +29,9 @@ namespace Microsoft.AspNetCore.Blazor.Server
             var configLines = File.ReadLines(configFilePath).ToList();
             SourceMSBuildPath = configLines[0];
 
+            if (string.IsNullOrEmpty(SourceMSBuildPath) || SourceMSBuildPath == ".")
+                SourceMSBuildPath = assemblyPath;
+
             var sourceMsBuildDir = Path.GetDirectoryName(SourceMSBuildPath);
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -5,4 +5,5 @@ obj/
 launchSettings.json
 artifacts/
 msbuild.binlog
-.vscode/
\ No newline at end of file
+.vscode/
+src/Microsoft.AspNetCore.Blazor.Browser.JS/package-lock.json
```
SteveSandersonMS: Could you revert the changes to this file? We don't want to exclude `package-lock.json` from source control.

src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
```diff
@@ -29,6 +29,9 @@ private BlazorConfig(string assemblyPath)
             var configLines = File.ReadLines(configFilePath).ToList();
             SourceMSBuildPath = configLines[0];
 
+            if (string.IsNullOrEmpty(SourceMSBuildPath) || SourceMSBuildPath == ".")
```
SteveSandersonMS: Is there any scenario where `SourceMSBuildPath` is null or empty? I'm not aware that there is. If I'm correct, could you revert that part of the `if` clause?

src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
```diff
@@ -29,6 +29,9 @@ private BlazorConfig(string assemblyPath)
...
+                SourceMSBuildPath = assemblyPath;
```
SteveSandersonMS: Could you put braces on separate lines around this one? The code style here is always to write:

```
if (condition)
{
    action();
}
```

... and not just

```
if (condition)
    action();
```

src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
Daddoon: Actualy no, i have only seen "." instead, maybe i have overdone this. We can assume that "." case is the only alternate scenario possible actually.

I will modify this part, thanks !

src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
Daddoon: Thanks for the information, i will do it like this in futur contribution on the project.

SteveSandersonMS:
Thanks @Daddoon! This is now merged into `dev`.

PR #480, opened 2018-04-04
Add Razor baseline test infrastructure

rynowak:
Ports somee infrastructure and converts Razor code generation tests to use
it. This makes it much easier to make cross cutting changes to code
generation and see the effect.

Use build /p:GenerateBaselines=true to update all of the generated code
in place or when adding new tests. Generally if tests are failing, the
easiest thing to do is to update the baselines and do a git diff to see
what the deltas are.

The changes to the tests here are to use the new baseline infrastructure
and to rename classes/methods to result in shorter file paths.


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentBodyExtensionNode.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.AspNetCore.Blazor.Razor
 {
-    public sealed class ComponentBodyExtensionNode : ExtensionIntermediateNode
+    internal sealed class ComponentBodyExtensionNode : ExtensionIntermediateNode
```
rynowak: This was always intended to be internal, it was just missed.

test/Microsoft.AspNetCore.Blazor.Build.Test/Razor/InitializeTestFileAttribute.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+using Microsoft.AspNetCore.Blazor.Build.Test;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNetCore.Razor.Language.IntegrationTests
```
rynowak: Most of the rest of this (other than generated files) has been copied verbatim from Razor. it's not really in a state in the razor repo where it's designed to be shared, and it was only like 4-5 files.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/DesignTimeCodeGenerationTest/ChildComponent_WithChildContent/TestComponent.ir.txt
```diff
@@ -0,0 +1,36 @@
+Document -
```
rynowak: This format ends up being the most useful for understanding bugs in Razor. This is the main reason why I wanted to switch to this style. It ends up being faster and easier to just add a new test and run it than to debug.

test/Microsoft.AspNetCore.Blazor.Build.Test/RazorIntegrationTestBase.cs
```diff
@@ -60,11 +61,11 @@ public RazorIntegrationTestBase()
 
             Configuration = BlazorExtensionInitializer.DefaultConfiguration;
             FileSystem = new VirtualRazorProjectFileSystem();
+            PathSeparator = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "\\" : "/";
```
SteveSandersonMS: Incredibly minor point, but for my own education, how is this different from `Path.PathSeparator`?

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/DesignTimeCodeGenerationTest/ChildComponent_WithChildContent/TestComponent.ir.txt
SteveSandersonMS: Sounds good!

test/testapps/LiveReloadTestApp/Home.cshtml
```diff
@@ -1 +1 @@
-<h1>Hello, world!</h1>
+<h1>Goodbye, world!</h1>
```
SteveSandersonMS: This is why the tests failed - the live reloading test should have reverted it back to `Hello` after the test completed (there's logic inside a `finally` block to do that). Not sure why the live reloading test failed to clean up after itself, but if you can repro that please let me know!

Anyway you'll just need to revert this change and hopefully then the build will succeed.

test/Microsoft.AspNetCore.Blazor.Build.Test/RazorIntegrationTestBase.cs
rynowak: We need the path logic to work on the same way on Windows and Linux for the baseline based tests because the filenames are backed into the tests themselves. Methods like `Path.GetFilename` and `Path.Combine` won't work with windows paths on linux.

SteveSandersonMS: I get that, I just wondered whether there was any reason to evaluate `RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "\\" : "/"` instead of using `Path.DirectorySeparatorChar` which AFAIK would return the same platform-sensitive character. Not at all important though!

rynowak: Ah, I missed what you were getting at. Sure that's right. I was doing this on my mac without intellisense.

PR #484, opened 2018-04-04
Revert live reloading implementation

SteveSandersonMS:
We're not going to include live reloading in 0.2.0 because, on further testing, there turn out to be VS+IISExpress cases that mean it would cause problems for people in certain cases.

The basic difficulty is that the client-side app can't reload before IISExpress has restarted, or after it has restarted. And unfortunately that kind of excludes all the possible times.

 1. If it tries to reload **before** IISExpress is fully restarted, then the requests can either fail as 502s, or the responses can be truncated part-way through. Worse, in multi-project builds, it can actually cause the build to fail because it tries to restart IISExpress too early and locks the `.dll` files, preventing the post-build file copy step from completing.
 2. If it tries to reload **after** IISExpress is fully restarted, then whatever socket/eventsource connection was being used to notify the client will already have been killed, so we have no way to notify the client.

We did try a modified version of (2) whereby the client would detect disconnection, and then start polling the server to find out when it was back up. But the polling itself triggers the same problems as in (1), i.e., it can break the build by forcing the server to restart too early.

Longer term, we can implement this robustly by *not* using the normal client-server connection as the notification channel, but instead having either the IDE or some auto-launched child process as a persistent side channel that doesn't get restarted (and hence disconnected) after each build. In dev mode, the client-side code can connect to this side channel (e.g., as a Websocket connection) to receive notifications of when it should reload. Those notifications should be given *after* IISExpress is fully restarted.


Mun1z:
Any advice on when live reload will be implemented?

PR #489, opened 2018-04-05
Fix mistake in array index

flash2048:
In the loop, only 0 array index was checked

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BindLoweringPass.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BindLoweringPass.cs
@@ -281,7 +281,7 @@ namespace Microsoft.AspNetCore.Blazor.Razor
             var segments = attributeName.Split('-');
             for (var i = 0; i < segments.Length; i++)
             {
-                if (string.IsNullOrEmpty(segments[0]))
+                if (string.IsNullOrEmpty(segments[i]))
                 {
                     return false;
                 }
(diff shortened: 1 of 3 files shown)


flash2048:
@SteveSandersonMS ok, I'll add the test tonight

flash2048:
@SteveSandersonMS I added test
But TryParseBindAttribute method is private
I used reflection to obtain a method. 
It may not be necessary to use unit tests for private methods, but most of the possible errors it will cover.

SteveSandersonMS:
@flash2048 Sorry if I was unclear. Yes that method is `private`, and therefore the kind of test we'd be looking for is one that exercises it using a valid `public` code path from the outside rather than using reflection to access the private method directly.

For this method, I'd hope we could add another test to `BindRazorIntegrationTest.cs`, adding a case where the Razor code it's trying to compile contains a malformed `bind` attribute that triggers the change of behavior. For example, I expect that `bind-=` or `bind--=` might do it.

Do you think that's viable?

flash2048:
I think it will be more effective to test this method directly as it is now.
In the `BindRazorIntegrationTest.cs` there are already tests covering the correct use.

SteveSandersonMS:
> In the BindRazorIntegrationTest.cs there are already tests covering the correct use.

But there aren't any tests covering the specific behavior you're fixing, which is what we need.

SteveSandersonMS:
That's excellent @flash2048 - thanks so much for this! It's now merged into `dev`.

I hope you don't mind but I excluded the reflection-based test of the `private` method, as that doesn't fit into the pattern of our tests (wouldn't be refactorable, etc.). The integration test you added covers this instead.

Issue #340, opened 2018-03-23
Component libraries


danroth27:

muqeet-khan:
Prior to 0.1.0, I could create a simple Blazor component library as NetStandard 2.0 with the following:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Blazor" Version="0.1.0" />
    <PackageReference Include="Microsoft.AspNetCore.Blazor.Build" Version="0.1.0" />
    <PackageReference Include="Microsoft.AspNetCore.Blazor.Razor.Extensions" Version="0.1.0" />
    <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.1.0-preview2-30230" />
    <PackageReference Include="Microsoft.NET.Sdk.Razor" Version="2.1.0-preview2-30230" />
  </ItemGroup>
</Project>
```

Currently cannot compile unless I do not link on build by: 

`<BlazorLinkOnBuild>False</BlazorLinkOnBuild>`

Building without the flag throws error as following:

[...]

PR #492, opened 2018-04-05
Reference static content from referenced assemblies. Implements #340

SteveSandersonMS:
This replaces the previous NuGet `contentFiles` solution. The new approach here uses embedded resources, which is much easier for the content package author, and behaves identically whether you're using a project reference or a package reference to the content package.

Previously I thought the `contentFiles` solution was necessary to avoid bloating the assemblies by including arbitrarily large static content embedded in them. But now we're linking by default on all builds, static content can always be stripped so it never reaches the client except when loaded deliberately (via `<link>` or `<script>` tags).

Note that the existing E2E test for package content files still applies and passes as before, because this PR also updates the format of `TestContentPackage` to the new style.


src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -588,26 +589,30 @@
...
     <PropertyGroup>
       <_LinkerEnabledFlag Condition="'$(_BlazorShouldLinkApplicationAssemblies)' != ''">--linker-enabled</_LinkerEnabledFlag>
     </PropertyGroup>
 
-    <Exec Command="$(BlazorBuildExe) build @(IntermediateAssembly) --html-page &quot;$(BlazorIndexHtml)&quot; @(_AppReferences->'--reference &quot;%(Identity)&quot;', ' ') @(_JsReferences->'--js &quot;%(Identity)&quot;', ' ') @(_CssReferences->'--css &quot;%(Identity)&quot;', ' ') $(_LinkerEnabledFlag) --output &quot;$(BlazorIndexHtmlOutputPath)&quot;" />
+    <Exec Command="$(BlazorBuildExe) build @(IntermediateAssembly) --html-page &quot;$(BlazorIndexHtml)&quot; @(_AppReferences->'--reference &quot;%(Identity)&quot;', ' ') @(_UnlinkedAppReferencesPaths->'--embedded-resources-source &quot;%(Identity)&quot;', ' ') $(_LinkerEnabledFlag) --output &quot;$(BlazorIndexHtmlOutputPath)&quot;" />
 
     <ItemGroup Condition="Exists('$(BlazorIndexHtmlOutputPath)')">
       <_BlazorIndex Include="$(BlazorIndexHtmlOutputPath)" />
+      <_BlazorIndexEmbeddedContentFile Include="$(BlazorIndexHtmlOutputDir)_content\**\*.*" />
```
SteveSandersonMS: I'm aware of the general guidance not to make decisions based on the files that happen to be on disk (i.e., wildcards), however in this specific case it's exactly equivalent to returning a response file from the `<Exec>` command above.

If we're running this target at all, then we know for sure that the contents of `_content` have been cleared and regenerated entirely.

rynowak: Yeah, I agree that's only an issue if you're relying on incremental clean. If you delete the directory and regenerate it then there's no problem with a wildcard.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
```diff
@@ -588,26 +589,30 @@
...
       <BlazorItemOutput Include="@(_BlazorIndex)">
         <TargetOutputPath>$(ProjectDir)$(OutputPath)dist/%(FileName)%(Extension)</TargetOutputPath>
         <Type>EntryPoint</Type>
       </BlazorItemOutput>
+      <BlazorItemOutput Include="@(_BlazorIndexEmbeddedContentFile)">
+        <TargetOutputPath>$(ProjectDir)$(OutputPath)dist/_content/%(RecursiveDir)%(FileName)%(Extension)</TargetOutputPath>
+      </BlazorItemOutput>
       <FileWrites Include="$(BlazorIndexHtmlOutputPath)" />
+      <FileWrites Include="@(_BlazorIndexEmbeddedContentFile)" />
```
rynowak: It will probably be worth converting this whole thing to a task at some point, I don't think there's any reason to change it now, but it's IMO a more flexible way to solve these things since you can return data to MSBuild.

src/Microsoft.AspNetCore.Blazor.Build/Core/EmbeddedResources/EmbeddedResourcesProcessor.cs
```diff
@@ -0,0 +1,137 @@
...
+            if (resource is EmbeddedResource embeddedResource)
+            {
+                if (TryInterpretLogicalName(resource.Name, out var kind, out var name))
+                {
+                    // Prefix the output path with the assembly name to ensure no clashes
+                    // Also be invariant to the OS on which the package was built
+                    name = Path.Combine(ContentSubdirName, assemblyName, EnsureHasPathSeparators(name, Path.DirectorySeparatorChar));
+
+                    // Write the file content to disk, ensuring we don't try to write outside the output root
+                    var outputPath = Path.GetFullPath(Path.Combine(outputDirPath, name));
```
rynowak: In what cases could this happen? Is this just checking for the possiblity that a resource name has an absolute path (or path navigation operators) in it?

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorContent.CSharp/.template.config/template.json
```diff
@@ -0,0 +1,70 @@
...
+      "manualInstructions": [
+        {
+          "text": "Run 'dotnet restore'"
+        }
+      ],
+      "actionId": "210D431B-A78B-4D2F-B762-4ED3E3EA9025",
+      "continueOnError": true
+    }
+  ]
+}
```
rynowak: I don't really know what's going on with these files so not commenting on them 😆

src/Microsoft.AspNetCore.Blazor.Build/Core/EmbeddedResources/EmbeddedResourcesProcessor.cs
SteveSandersonMS: Yes exactly. With NuGet packages (or NPM, etc) in general there’s a level of trust that they aren’t malicious, but it would be easy to envision someone accidentally making a package with an absolute pathname here and it might seem to work OK to them, but would utterly freak out the package consumer if files appeared at unrelated locations on disk.

src/Microsoft.AspNetCore.Blazor.Build/Core/EmbeddedResources/EmbeddedResourceKind.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Blazor.Build
+{
+    internal enum EmbeddedResourceKind
+    {
+        JavaScript,
+        Css,
+        Static
```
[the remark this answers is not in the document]
SteveSandersonMS: Yes. I didn’t have a better name for them.

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorContent.CSharp/BlazorContent.CSharp.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <!-- This custom package feed is required only when using nightly builds of Blazor -->
+    <RestoreSources>https://dotnet.myget.org/F/blazor-dev/api/v3/index.json;$(RestoreSources)</RestoreSources>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- .js/.css files will be referenced via <script>/<link> tags; other content files will just be included in the app's 'dist' directory without any tags referencing them -->
+    <EmbeddedResource Include="content\**\*.js" LogicalName="blazor:js:%(RecursiveDir)%(Filename)%(Extension)" />
+    <EmbeddedResource Include="content\**\*.css" LogicalName="blazor:css:%(RecursiveDir)%(Filename)%(Extension)" />
+    <EmbeddedResource Include="content\**" Exclude="**\*.js;**\*.css" LogicalName="blazor:file:%(RecursiveDir)%(Filename)%(Extension)" />
+  </ItemGroup>
```
rynowak: We could figure out a way to hide this if you don't want it to show up in the project. I don't think we should block on it for now.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.targets
SteveSandersonMS: Maybe. It’s very difficult to build tasks with .NET Core and ship them via NuGet today (ask Nate about all the stuff that goes wrong), but maybe it will get better in the future.

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorContent.CSharp/BlazorContent.CSharp.csproj
SteveSandersonMS: Yeah, it would be nice to hide them in a lot of cases, but there will be cases where people want to customise it. For example the might not want all CSS files to get top-level link tags, but rather just some subset that references the others. It would be hard to discover how to customise if you don’t see this source code here. 

Open to changing this in the future though!

PR #494, opened 2018-04-05
Add DI extension method namespace to Program.cs in templates

danroth27:
Adding a `using` for `Microsoft.Extensions.DependencyInjection` so that you can use common DI extension methods like `AddSingleton`.


danroth27:
Also rename `configure` parameter to `services` like it is in ASP.NET Core.

PR #500, opened 2018-04-06
Correct the use of a parameter 'autoSpace' in the constructor

flash2048:
Constructor parameter '**autoSpace**' is not used. It has default **true** value too.


SteveSandersonMS:
Thanks!

PR #516, opened 2018-04-09
Improvements for delegate types

rynowak:
See commits for details


src/Microsoft.AspNetCore.Blazor/Components/BindMethods.cs
```diff
@@ -35,10 +35,10 @@ public static string GetEventHandlerValue<T>(string value)
         /// <summary>
         /// Not intended to be used directly.
         /// </summary>
-        public static UIEventHandler GetEventHandlerValue<T>(Action<T> value)
+        public static MulticastDelegate GetEventHandlerValue<T>(Action<T> value)
             where T : UIEventArgs
         {
-            return e => value((T)e);
+            return value;
```
rynowak: This is the payoff. This will now 'just work', and doesn't cause a diff each time the component renders (except in the case of a capturing lambda).

The advantage of having the tag helper at all (vs no language support) is that because we know what `T` is at compile time, is that you get a much better experience using a lambda. Without the language support, lambdas would have to be `UIEventHandler` **only**

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
```diff
@@ -141,14 +149,44 @@ public void AddAttribute(int sequence, string name, string value)
...
+        /// <param name="value">The value of the attribute.</param>
+        /// <remarks>
+        /// This method is provided for infrastructure purposes, and is used to be
+        /// <see cref="UIEventHandlerRenderTreeBuilderExtensions"/> to provide support for delegates of specific
+        /// types. For a good programming experience when using a custom delegate type, define an
+        /// extension method similar to 
+        /// <see cref="UIEventHandlerRenderTreeBuilderExtensions.AddAttribute(RenderTreeBuilder, int, string, UIChangeEventHandler)"/>
+        /// that calls this method.
+        /// </remarks>
+        public void AddAttribute(int sequence, string name, MulticastDelegate value)
```
rynowak: See the commit message for details. Tuning the overload set of of `AddAttribute` to make it do everything we want was a little tricky.

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
```diff
@@ -78,8 +78,13 @@ protected void DispatchEvent(int componentId, int eventHandlerId, UIEventArgs ev
...
                 try
                 {
+                    if (!(handler is UIEventHandler uiEventHandler))
+                    {
+                        uiEventHandler = (UIEventArgs e) => handler.DynamicInvoke(e);
+                    }
+
                     _isBatchInProgress = true;
-                    GetRequiredComponentState(componentId).DispatchEvent(handler, eventArgs);
+                    GetRequiredComponentState(componentId).DispatchEvent(uiEventHandler, eventArgs);
```
rynowak: This was the trick. We only generate a 'wrapper' after we've processed the diff and decided to register the delegate for events. This allows the diff phase to leverage the delegate equality rules.

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
```diff
@@ -113,7 +118,28 @@ internal void InstantiateChildComponentOnFrame(ref RenderTreeFrame frame)
...
+            {
+                // IMPORTANT: we're creating an additional delegate when necessary. This is
+                // going to get cached in _eventHandlersById, but the render tree diff
+                // will operate on 'AttributeValue' which means that we'll only create a new
+                // wrapper delegate when the underlying delegate changes.
+                //
+                // TLDR: If the component uses a method group or a non-capturing lambda
+                // we don't allocate much.
+               _eventHandlersById.Add(id, (UIEventArgs e) => @delegate.DynamicInvoke(e));
+            }
```
rynowak: This was the trick. We only generate a 'wrapper' after we've processed the diff and decided to register the delegate for events. This allows the diff phase to leverage the delegate equality rules

src/Microsoft.AspNetCore.Blazor/UIEventHandlerRenderTreeBuilderExtensions.cs
```diff
@@ -0,0 +1,80 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Blazor.RenderTree;
+
+namespace Microsoft.AspNetCore.Blazor
+{
+    public static class UIEventHandlerRenderTreeBuilderExtensions
```
SteveSandersonMS: It would be great to include a comment block at the top of this file explaining the scenarios where these overloads are needed. I was finding it hard to figure out without actually commenting them out to see what broke, given that there's also an overload that takes `UIEventHandler` (the base type) directly and has exactly the same logic inside it.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -597,9 +597,15 @@ private static void InitializeNewComponentFrame(ref DiffContext diffContext, int
 
         private static void InitializeNewAttributeFrame(ref DiffContext diffContext, ref RenderTreeFrame newFrame)
         {
-            if (newFrame.AttributeValue is UIEventHandler)
+            if (newFrame.AttributeValue is MulticastDelegate @delegate &&
+                @delegate.Method != null)
             {
-                diffContext.Renderer.AssignEventHandlerId(ref newFrame);
+                var parameters = @delegate.Method.GetParameters();
```
SteveSandersonMS: I think `GetParameters` allocates (i.e., for the returned array). I'd strongly value not allocating here on every render, even at the cost of less fidelity over which delegate types we assign event handler IDs to.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -597,9 +597,15 @@ private static void InitializeNewComponentFrame(ref DiffContext diffContext, int
...
+                if (parameters.Length == 1 && 
+                    typeof(UIEventArgs).IsAssignableFrom(parameters[0].ParameterType))
+                {
+                    diffContext.Renderer.AssignEventHandlerId(ref newFrame);
+                }
```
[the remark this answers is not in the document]
SteveSandersonMS: > I assume we only want to 'register' delegates that can receive dom events

I'm not sure that's so important. Are there many likely scenarios where someone tries to use `onsomething=@myDelegate` where `myDelegate` isn't a `UIEventHandler`? Even if someone does that and we assign an event handler ID that never gets used, that doesn't seem like a big problem.

I'd favor optimizing for the common case (where the delegate is a `UIEventHandler`) even if in the uncommon case we do something a little less inefficient.

src/Microsoft.AspNetCore.Blazor/UIEventHandlerRenderTreeBuilderExtensions.cs
SteveSandersonMS: In fact I think it would be better still if it could document the overall flow, in terms of how different Razor syntaxes lead to different `AddAttribute` calls, what the different method overload resolution possibilities are, and what this means for the different code paths at runtime. It will be tricky to keep track of all this otherwise, as the implementation is spread out over quite a lot of files.

I know there always a risk that documentation like that tends to rot over time, but on this occasion it's hard to keep track of it at all otherwise. What do you think?

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
```diff
@@ -113,7 +113,28 @@ internal void InstantiateChildComponentOnFrame(ref RenderTreeFrame frame)
...
         {
             var id = ++_lastEventHandlerId;
-            _eventHandlersById.Add(id, (UIEventHandler)frame.AttributeValue);
+
+            // The attribute value might be a more specialized type like UIKeyboardEventHandler.
+            //
+            // In order to dispatch the event, we need a UIEventHandler, so we're going weakly
+            // typed here. The user will get a cast exception if they map the wrong type of
+            // delegate to the event.
+            if (frame.AttributeValue is UIEventHandler wrapper)
```
SteveSandersonMS: I will go and read the code until I understand it properly, but at first glance I'm struggling to tell why it's sometimes a `UIEventHandler` and other times a `MulticastDelegate`. Might be useful to have a comment specifying that explicitly.

SteveSandersonMS:
From the commit message:

> You will still get some inefficient behavior if your component uses a
> capturing lambda in an event handler, so don't do that.

While I understand there's nothing we can do about the fact that you're creating a different delegate instance on each render, it should be noted that capturing lambdas are legitimate and necessary in many cases. For example, if you're rendering a list of items and want a "delete" button next to each one, you're probably going to use a capturing lambda for that. It is technically possible to avoid the capturing lambda by having a separate component instance for each row, but that's quite a bit more work for the developer in some cases, so the capturing lambda is still probably preferable there.

SteveSandersonMS:
We probably should have discussed this earlier, but what would you consider to be a good way for us to deal with async event handlers? Previously we were going to have `@onclick` and `@onclickAsync` so that you could declare either `void MyHandler()` or `async Task MyHandler()`.

Do you think we'll end up extending the syntax to `onclick-async=@MyMethod`, or is there some way that we could make it work at compile-time if `MyMethod` returned either `void` or `Task`?

I'm not suggesting we should block this PR for that, but it would be good to be sure we have a likely plan in mind.

src/Microsoft.AspNetCore.Blazor/Components/BindMethods.cs
SteveSandersonMS: Excellent!

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
SteveSandersonMS: Makes sense. Sounds like a good way to get the perf benefits.

rynowak:
> it should be noted that capturing lambdas are legitimate and necessary in many cases. For example, if you're rendering a list of items and want a "delete" button next to each one, you're probably going to use a capturing lambda for that. It is technically possible to avoid the capturing lambda by having a separate component instance for each row, but that's quite a bit more work for the developer in some cases, so the capturing lambda is still probably preferable there.

The old school workaround for this would be to pass an additional `state` parameter. I don't want to do that though, I think this is good enough and that pattern is familiar.

rynowak:
> Is it the case that compile-time resolved UIEventHandler-compatible lambdas avoid it, and it's only used if you pass some other delegate type (and then we rely on it being compatible at runtime)?

The resolution is runtime based, since it's relies on casts. And yes, if you pass us something like like a `UIKeyboardEventHandler` or `Action<UIEventHandler>` it will be checked at runtime by Mono.

> Finally I hope we can avoid the need to inspect the delegate parameter types at runtime just by not worrying if we assign event handler IDs to things that wouldn't use them (unless you have reason to believe this will be extremely common).

That would streamline this, as well as make it possible to use things like `Action<object>`. I'll try this out.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
```diff
@@ -185,7 +223,7 @@ public void AddAttribute(int sequence, string name, object value)
 
                     // Don't add anything for false bool value.
                 }
-                else if (value is UIEventHandler eventHandler)
+                else if (value is MulticastDelegate eventHandler)
```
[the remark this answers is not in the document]
rynowak: Sure

src/Microsoft.AspNetCore.Blazor/UIEventHandlerRenderTreeBuilderExtensions.cs
rynowak: Sure

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
rynowak: Sure

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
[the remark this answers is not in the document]
SteveSandersonMS: We could end up doing something like "if it's on a component, leave it alone; if it's on an HTML element, then assign IDs to all delegate values" since there's no use case for delegate-typed HTML attribute values other than event handlers. I know we don't have the "component or not" info directly available here but could find some way to make it available (possibly as an extra param from the caller, or possible as extra info on the `RenderTreeFrame` itself).

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
rynowak: OK great. Sounds like we'll be able to avoid this.

rynowak:
@SteveSandersonMS - I think the changes to this worked out well.

We no longer look at the type of the delegate and instead register it if it matches the `"onfoo"` naming pattern.

PR #518, opened 2018-04-09
Enable same-origin credentials by default. Add E2E test to show they …

SteveSandersonMS:
…can be sent to different-origin domains too.


SteveSandersonMS:
Merged.

PR #521, opened 2018-04-09
Improve JS-side event handling code

SteveSandersonMS:
Previously the JS-side event-handling code was very rough and temporary. This new version:

 * Supports arbitrary event names
 * Has a cleaner system for us adding event-type-specific data extraction logic corresponding to the .NET side code (see `EventForDotNet.ts`)
 * No longer has to register separate DOM event handlers for every (element,eventName) pair. Now it contains an event delegation system so that each `BrowserRenderer` only has at most one real DOM event handler for each eventName, and dispatches the events based on expando info stored on the DOM elements. It also takes care of registering/unregistering the DOM event handlers according to whether there are any active listeners for that event name. Overall this is much less resource-heavy. It also gives us an easy way to plug in our own event bubbling/cancellation logic later if we want.

**Not yet handled**: We still don't properly handle nested event handlers. For example, if element A contains element B in the DOM, and both A and B have `click` handlers, and you click on `B`, then you'd expect both handlers to run followed by a single render. What actually happens is that the handler for `B` runs, then it does a first render, then it tries to run `A`'s handler which *might* run (if it hasn't been disposed) or it might crash (if it has been disposed). What we should do to to fix this is change the event dispatching logic so that we send an array of all the event handler IDs to be invoked, and then the .NET-side code should run all of them, followed by a single re-render.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
@@ -2,18 +2,24 @@
 import { getRenderTreeEditPtr, renderTreeEdit, RenderTreeEditPointer, EditType } from './RenderTreeEdit';
 import { getTreeFramePtr, renderTreeFrame, FrameType, RenderTreeFramePointer } from './RenderTreeFrame';
 import { platform } from '../Environment';
+import { EventDelegator } from './EventDelegator';
+import { EventForDotNet, UIEventArgs } from './EventForDotNet';
 const selectValuePropname = '_blazorSelectValue';
 let raiseEventMethod: MethodHandle;
 let renderComponentMethod: MethodHandle;
 
 export class BrowserRenderer {
+  private eventDelegator: EventDelegator;
   private childComponentLocations: { [componentId: number]: Element } = {};
 
   constructor(private browserRendererId: number) {
+    this.eventDelegator = new EventDelegator((event, componentId, eventHandlerId, eventArgs) => {
+      raiseEvent(event, this.browserRendererId, componentId, eventHandlerId, eventArgs);
+    });
   }
 
-  public attachComponentToElement(componentId: number, element: Element) {
-    this.childComponentLocations[componentId] = element;
+  public attachRootComponentToElement(componentId: number, element: Element) {
+    this.attachComponentToElement(componentId, element);
   }
 
   public updateComponent(componentId: number, edits: System_Array<RenderTreeEditPointer>, editsOffset: number, editsLength: number, referenceFrames: System_Array<RenderTreeFramePointer>) {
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -213,7 +213,7 @@ protected RenderTreeFrame bind(DateTime value, string format)
         /// <returns>A <see cref="RenderTreeFrame"/> that represents the event handler.</returns>
         protected RenderTreeFrame onclick(Action handler)
             // Note that the 'sequence' value is updated later when inserted into the tree
-            => RenderTreeFrame.Attribute(0, "onclick", _ => handler());
+            => RenderTreeFrame.Attribute(0, "onclick", handler != null ? (_ => handler()) : (UIEventHandler)null);
```
SteveSandersonMS: I know this code will be redundant once #516 is merged, but it's needed until then for the E2E scenario to be valid

grahamehorner:
EventForDotNet seems to suggest that .net is only interested in UIEvents ? Developers may be interested in other non dom events Window events and or other including Worker events

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
```diff
@@ -162,68 +178,41 @@ export class BrowserRenderer {
...
-  applyAttribute(componentId: number, toDomElement: Element, attributeFrame: RenderTreeFramePointer) {
+  private applyAttribute(componentId: number, toDomElement: Element, attributeFrame: RenderTreeFramePointer) {
     const attributeName = renderTreeFrame.attributeName(attributeFrame)!;
     const browserRendererId = this.browserRendererId;
     const eventHandlerId = renderTreeFrame.attributeEventHandlerId(attributeFrame);
 
+    if (eventHandlerId) {
+      const firstTwoChars = attributeName.substring(0, 2);
+      const eventName = attributeName.substring(2);
+      if (firstTwoChars !== 'on' || !eventName) {
```
rynowak: Do you think we should use the same logic in the `Renderer`? We were discussing whether or to 'track'  all delegates regardless of type, it makes sense that we'd use a check like this also since this is the common pattern for DOM events.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -0,0 +1,46 @@
...
+        const newValue = targetIsCheckbox ? !!element['checked'] : element['value'];
+        return new EventForDotNet<UIChangeEventArgs>('change', { Type: event.type, Value: newValue });
+      }
+      case 'keypress':
+        return new EventForDotNet<UIKeyboardEventArgs>('keyboard', { Type: event.type, Key: (event as any).key });
+      default:
+        return new EventForDotNet<UIEventArgs>('unknown', { Type: event.type });
+    }
+  }
+}
```
rynowak: Do you think that we'll ever turn the set of payloads here into an open set?

Put another way, would we ever make it possible to write your own `UIEventArgs`-derived class, and wire it up with strongly-typed data? I can't think of a reason why we need it now, but I was wondering if this would become important at some point.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
SteveSandersonMS: Yes, that does seem reasonable. Certainly I agree that whatever rules we have should be consistent across compilation, .NET side, and JS side.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
SteveSandersonMS: I could totally imagine us doing this as some point, either by having some API for plugging in your own event mappers, or perhaps more generally just serializing as much of the JS-side event object as we can and letting developers plug in a `UIEventArgs` subclass on the .NET side to deserialize any matching properties into.

But for now I think we're OK with controlling the set of event mappings ourselves.

Issue #503, opened 2018-04-06
Add first class support for onclick and other event handlers


rynowak:
## What is it?

We're doing some tweaks to the ways that event handlers get wired to DOM events in Blazor, and adding a lot more features and power.

Hooking up a C# function to a handle a DOM event will look like:
```
<button onclick="@OnClick" />
@functions {
    void OnClick(UIMouseEventArgs e)
    {
        Console.WriteLine("hello, world");
    }
} 
```

or if you'd prefer a lambda:
```
<button onclick="@(e => Console.WriteLine("hello, world"))"
```

This has first class integration with the editor in Visual Studio and will include tooltips and completion.

Passing in a string like `<button onclick="foo" />` will not do any magic, and will work as if it were normal HTML.

## What's changing

This kind of thing has been possible today using a few temporary syntaxes that will be removed as part of 0.2. Namely `<button @onclick(Foo) />`  and `<button onclick=@{ Foo(); } />. This change makes the event handler wire up look and function the same way as other Razor features.

## What's added

Along with this change we'll be defining more of the standard DOM events with first class completion and eventargs types.

We'll also be adding support for *arbitrary event names*. So `<button onfoo="@Foo" />` will bind the Foo method to be called when the `foo` DOM event is fired.

## TODO
- [ ] Add language support for `onclick` and `onchange`
- [ ] Remove temporary workarounds
- [ ] Define more standard DOM events and eventargs
- [ ] Add support for arbitrary events
- [ ] Remove allocations for common cases

dlr1:
how would passing parameters look like?
for ex; <div onclick="@OnClick(item)"/>

yowl:
I guess if you use the lambda variant you can do that
`<button onclick="@(e => Console.WriteLine("hello, world " + item))"`

dlr1:
I should have been more specific. How would the eventhandler be looking like in code/@functions, are the arguments going to be added the handler, is it params argument type, etc

AWulkan:
Is there any specific syntax for passing the event? Like `<div onclick="@OnClick($event)"/>`?

Suchiman:
@AWulkan just look at the first post, the example shows how to pass event (namely you don't explicitly, you always do). You currently cannot have events not receiving the event argument

SteveSandersonMS:
@AWulkan 

[...]

Issue #409, opened 2018-03-28
Add 'bind-...' for two-way binding


rynowak:
## Summary
We're adding a feature that replaces `@bind(...)` with something more first class for tooling. 

Introducing `bind-...` for two-way binding:
```
<input bind="@CurrentValue" />
@functions {
    public string CurrentValue { get; set; }
}
```

`bind-...` attribute implements two-way binding to components and dom elements via a tag helper with special compile-time behavior. This means that intellisense, colorization, and completion know about `bind-...` and can provide contextual documentation and contextual completion.

![image](https://user-images.githubusercontent.com/1430011/38001672-e272c8d4-31e2-11e8-9cda-28ae4dfdcb98.png)

**note: the correct date format is actually `yyyy-MM-dd`. Do as I say, not as I do 😆**

## What is two-way binding?

If you're not familiar with `@bind` (the thing being upgraded), this does **two-way** binding to both components and dom elements. This is like a macro, it doesn't do anything you couldn't do normally, but it should help you out 👍 

Example:
```
<input bind="@CurrentValue" />
@functions {
    public string CurrentValue { get; set; }
}
```

Generates something like:
```
<input value="@CurrentValue" onchange="@((__value) => CurrentValue = __value)/>
@functions {
    public string CurrentValue { get; set; }
}
```

This means than when the component first runs, the `value` of the input element will come from the `CurrentValue` property. When the user types in the textbox, the `CurrentValue` property will be set to the changed value. This is why we call it *two-way binding*.

In reality the code generation is a little more complex because bind deals with a few cases of type conversions. But in principle, `bind-...` will associate the current value of an expression with a *value* attribute, and will handle changes via a *change handler* attribute. We expect to invest more in the runtime support in this area in the future, including better conversion, error handler, etc.

One additional point, the expression provided to `bind-...` should be an **LValue**, meaning that it's something that can be assigned. Since this is code-generation, it works in places that the C# ref keyword doesn't work.

React can do something similar: https://reactjs.org/docs/two-way-binding-helpers.html

### DOM elements

The most obvious usage of `bind-...` is for input elements of various types. You're going to have lots of these 😆 

That looks like this:
```
<input bind="@CurrentValue" />
@functions {
    public string CurrentValue { get; set; }
}
```

Or this:
```
<input type="checkbox" bind="@IsSelected" />
@functions {
    public bool IsSelected { get; set; }
}
```

[...]

PR #525, opened 2018-04-10
Remove old workaround @onclick and @bind

rynowak:
This change removes support for the old syntax used for event handlers
and two-way binding.

See the relevant issues for details on the new features and
improvements:

bind https://github.com/aspnet/Blazor/issues/409
event handlers https://github.com/aspnet/Blazor/issues/503

Along with this change we've removed a few additional things Blazor
could do that aren't part of Razor's usual syntax.

----

The features that was used to make something like:
```
<button @onclick(...) />
```

is an expression that's embedded in a an element's attribute. This
feature might be useful in the future if we want to support 'splatting'
arbitrary attributes into a tag, but the runtime support for this isn't
accessible outside the Blazor core.

----

The features that implement:
```
<button onclick=@{ } />
```

have been removed in favor of a better design for lambdas, method group
conversions and other things for event handler attributes.

use `<button onclick=@(x => ...} />` instead.

We think is a better approach in general, because we want the app
developer to write and see the parameter list.

----

Both syntactic features that have been removed have dedicated error
messages in the compiler. If you're porting old code it should help you
figure out what to do.

### DIFF
--- a/samples/StandaloneApp/Pages/Counter.cshtml
+++ b/samples/StandaloneApp/Pages/Counter.cshtml
@@ -1,14 +1,15 @@
 ﻿@page "/counter"
+@using Microsoft.AspNetCore.Blazor
 <h1>Counter</h1>
 
 <p>Current count: @currentCount</p>
 
-<button @onclick(IncrementCount)>Click me</button>
+<button onclick="@IncrementCount">Click me</button>
 
 @functions {
     int currentCount = 0;
 
-    void IncrementCount()
+    void IncrementCount(UIMouseEventArgs e)
     {
         currentCount++;
     }
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -178,50 +178,5 @@ namespace Microsoft.AspNetCore.Blazor.Components
             // at the end of every event callback.
             StateHasChanged();
         }
-
-        // At present, if you have a .cshtml file in a project with <Project Sdk="Microsoft.NET.Sdk.Web">,
-        // Visual Studio will run design-time builds for it, codegenning a class that attempts to override
-        // this method. Therefore the virtual method must be defined, even though it won't be used at runtime,
-        // because otherwise VS will display a design-time error in its 'Error List' pane.
-        // TODO: Track down what triggers the design-time build for .cshtml files and how to stop it, then
-        // this method can be removed.
-        /// <summary>
-        /// Not used. Do not invoke this method.
-        /// </summary>
-        /// <returns>Always throws an exception.</returns>
-        public virtual Task ExecuteAsync()
-            => throw new NotImplementedException($"Blazor components do not implement {nameof(ExecuteAsync)}.");
-
-        /// <summary>
-        /// Applies two-way data binding between the element and the property.
-        /// </summary>
-        /// <param name="value">The model property to be bound to the element.</param>
-        protected RenderTreeFrame bind(object value)
-            => throw new NotImplementedException($"{nameof(bind)} is a compile-time symbol only and should not be invoked.");
-
-        /// <summary>
-        /// Applies two-way data binding between the element and the property.
-        /// </summary>
-        /// <param name="value">The model property to be bound to the element.</param>
-        protected RenderTreeFrame bind(DateTime value, string format)
-            => throw new NotImplementedException($"{nameof(bind)} is a compile-time symbol only and should not be invoked.");
-
-        /// <summary>
-        /// Handles click events by invoking <paramref name="handler"/>.
-        /// </summary>
-        /// <param name="handler">The handler to be invoked when the event occurs.</param>
-        /// <returns>A <see cref="RenderTreeFrame"/> that represents the event handler.</returns>
-        protected RenderTreeFrame onclick(Action handler)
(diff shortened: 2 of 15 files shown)


samples/StandaloneApp/Pages/Counter.cshtml
```diff
@@ -1,14 +1,15 @@
...
 <p>Current count: @currentCount</p>
 
-<button @onclick(IncrementCount)>Click me</button>
+<button onclick="@IncrementCount">Click me</button>
 
 @functions {
     int currentCount = 0;
 
-    void IncrementCount()
+    void IncrementCount(UIMouseEventArgs e)
```
rynowak: I didn't mess with it, because I wanted to get some feedback first, but I had two main thoughts while doing this porting that might improve usability....

1. We could add a special case for `Action` (no arg) for the case where you don't need the args, if we think it will be common. This would be pretty easy for us to do, but we would need to push knowledge of this case through a few levels of the stack.

2. We might want to consider adding `@using Microsoft.AspNetCore.Blazor` to the default usings. Now that we've got more useful stuff you'll want, we might want to do this as a practical matter. This will help keep docs and articles concise.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -178,50 +178,5 @@ void IHandleEvent.HandleEvent(UIEventHandler handler, UIEventArgs args)
...
-        // Visual Studio will run design-time builds for it, codegenning a class that attempts to override
-        // this method. Therefore the virtual method must be defined, even though it won't be used at runtime,
-        // because otherwise VS will display a design-time error in its 'Error List' pane.
-        // TODO: Track down what triggers the design-time build for .cshtml files and how to stop it, then
-        // this method can be removed.
-        /// <summary>
-        /// Not used. Do not invoke this method.
-        /// </summary>
-        /// <returns>Always throws an exception.</returns>
-        public virtual Task ExecuteAsync()
```
rynowak: This was missed from my cleanup a while back

test/Microsoft.AspNetCore.Blazor.Build.Test/DiagnosticRazorIntegrationTest.cs
```diff
@@ -47,5 +47,45 @@ public void RejectsEndTagWithDifferentNameToStartTag()
                     Assert.Equal(20, item.Span.CharacterIndex);
                 });
         }
+
+        // This is the old syntax used by @bind and @onclick, it's explicitly unsupported
+        // and has its own diagnostic.
+        [Fact]
+        public void OldEventHandlerSyntax_ReportsError()
```
rynowak: Moved these tests from the rendering tests since they are now error cases.

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
```diff
@@ -403,7 +374,7 @@ public void SupportsTwoWayBindingForTextboxes()
             // Assert
             var frames = GetRenderTree(component);
             Assert.Collection(frames,
-                frame => AssertFrame.Element(frame, "input", 3, 0),
+                frame => AssertFrame.Element(frame, "textarea", 3, 0),
```
rynowak: The diff is wacky. I created another test for textareas.

samples/StandaloneApp/Pages/Counter.cshtml
rynowak: I'm going to go ahead with this PR as is, then I'll follow up on these two items separately. The `Action` one carries some risk of not making it.

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
rynowak: Cool, will fix.

test/testapps/TestContentPackage/ComponentFromPackage.cshtml
```diff
@@ -1,14 +1,14 @@
...
+<div class="special-style">
     This component, including the CSS and image required to produce its
     elegant styling, is in an external NuGet package.
-    <button @onclick(ChangeLabel)>@buttonLabel </button>
+    <button onclick="@ChangeLabel">@buttonLabel </button>
 </div>
 
-@functions
-{
+@functions {
```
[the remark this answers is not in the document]
rynowak: I think the answer is that `@if (foo) {` is different than putting the `{` on a separate line.

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
```diff
@@ -359,42 +331,41 @@ public void SupportsUsingStatements()
         }
 
         [Fact]
-        public void SupportsAttributeFramesEvaluatedInline()
+        public void SupportsTwoWayBindingForTextboxes()
         {
             // Arrange/Act
             var component = CompileToComponent(
-                @"<elem @onclick(MyHandler) />
+                @"<input bind=""MyValue"" />
```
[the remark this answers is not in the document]
rynowak: We can always change it later, the problem will be that folk will not get any sort of warning, it would just become HTML

PR #528, opened 2018-04-10
Remove blazor-component tag + introducing BlazorDOMComponent

uazo:
Removed custom <blazor-component> tag, replacing it with

    
    ...
    

Created BlazorDOMElement and derivates to allow reuse of javascript components from blazor components


SteveSandersonMS:
@uazo Thanks for proposing this. Just so you know, we already have a specific plan for how to remove the `<blazor-component>` wrappers. Don't want you to spend your time on this as we already have it covered :)

uazo:
ok, no problem! so, what is your specific plan?

Also, what do you think about BlazorDOMElement? 
The idea is the separation of concerns between virtual dom and actual rendering.
Encapsulating tags or components would allow you to have a code that is more readable and easier to maintain (see the select bug).
It also allows you to better manage the animations (for now it is impossible to make a fadein of a component) and for example the implementation of Html.Raw would be fast.
Even the reuse of external js components (eg Datatables) would be faster.

PR #531, opened 2018-04-10
Auto rebuilds

SteveSandersonMS:
New approach to replace the didn't-work-properly live reloading system. The new approach is equivalent to how VS auto-rebuilds ASP.NET when you edit a `.cs` file and then begin a new HTTP request to the site.

Currently this only supports VS for Windows, and in other scenarios doesn't auto-rebuild. But the design allows for additional `IRebuildService` implementations such as VS for Mac (and arguably "command line", though I'm not sure that's a good idea because builds are not fast that way).

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AspNetCore.Builder
             }
 
             var currentCompilationTask = Task.CompletedTask;
-            var compilationTaskMustBeRefreshed = false;
+            var compilationTaskMustBeRefreshedIfNotBuiltSince = (DateTime?)null;
 
             WatchFileSystem(config, () =>
             {
@@ -37,17 +37,19 @@ namespace Microsoft.AspNetCore.Builder
                 // HTTP request arrives, because it's annoying if the IDE is constantly rebuilding
                 // when you're making changes to multiple files and aren't ready to reload
                 // in the browser yet.
-                compilationTaskMustBeRefreshed = true;
+                compilationTaskMustBeRefreshedIfNotBuiltSince = DateTime.Now;
             });
 
             appBuilder.Use(async (context, next) =>
             {
                 try
                 {
-                    if (compilationTaskMustBeRefreshed)
+                    if (compilationTaskMustBeRefreshedIfNotBuiltSince.HasValue)
                     {
-                        currentCompilationTask = rebuildService.PerformRebuildAsync(config.SourceMSBuildPath);
-                        compilationTaskMustBeRefreshed = false;
+                        currentCompilationTask = rebuildService.PerformRebuildAsync(
+                            config.SourceMSBuildPath,
+                            compilationTaskMustBeRefreshedIfNotBuiltSince.Value);
+                        compilationTaskMustBeRefreshedIfNotBuiltSince = null;
                     }
 
                     await currentCompilationTask;
--- a/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
@@ -1,24 +1,33 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Blazor.Server;
+using Microsoft.AspNetCore.Blazor.Server.AutoRebuild;
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.AspNetCore.Blazor.Server
+namespace Microsoft.AspNetCore.Builder
 {
-    internal static class AutoRecompilation
+    internal static class AutoRebuildExtensions
     {
-        // TODO: Make this configurable in csproj
-        private static string[] _watchedExtensions = new[] { ".cs", ".cshtml" };
-        private static string[] _excludeSubdirectories = new[] { "obj", "bin" };
+        // Note that we don't need to watch typical static-file extensions (.css, .js, etc.)
+        // because anything in wwwroot is just served directly from disk on each reload. But
+        // as a special case, we do watch index.html because it needs compilation.
+        // TODO: Make the set of extensions and exclusions configurable in csproj
+        private static string[] _includedSuffixes = new[] { ".cs", ".cshtml", "index.html" };
+        private static string[] _excludedDirectories = new[] { "obj", "bin" };
 
-        public static void UseAutoRecompilation(this IApplicationBuilder appBuilder, BlazorConfig config)
+        public static void UseAutoRebuild(this IApplicationBuilder appBuilder, BlazorConfig config)
         {
+            // Currently this only supports VS for Windows. Later on we can add
+            // an IRebuildService implementation for VS for Mac, etc.
+            if (!VSForWindowsRebuildService.TryCreate(out var rebuildService))
+            {
+                return; // You're not on Windows, or you didn't launch this process from VS
+            }
+
             var currentCompilationTask = Task.CompletedTask;
             var compilationTaskMustBeRefreshed = false;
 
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+                    // If there's no listener on the other end of the pipe, or if anything
+                    // else goes wrong, we just let the incoming request continue.
+                    // There's nowhere useful to log this information so if people report
+                    // problems we'll just have to get a repro and debug it.
+                    // If it was an error on the VS side, it logs to the output window.
+                }
+                finally
+                {
+                    compilationTaskMustBeRefreshedIfNotBuiltSince = null;
+                }
```
rynowak: How concerned are you about thread safety of this interchange. In particular if another thread observes a file change during a build, it looks like this would wipe out the `compilationTaskMustBeRefreshedIfNotBuiltSince`, which would have been set to a newer value than you just built with

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/ProcessUtils.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Blazor.Server.AutoRebuild
+{
+    internal static class ProcessUtils
+    {
+        // Based on https://stackoverflow.com/a/3346055
+
+        public static Process GetParent(Process process)
```
rynowak: Interestingly enough in .NET on windows the process APIs don't create a parent-child releationship between processes when you use `System.Process` to launch a new process. Not an issue here since VS does the right thing, but in case we want to use this technique elsewhere, that will come up.

src/Microsoft.AspNetCore.Blazor.Server/BlazorAppBuilderExtensions.cs
```diff
@@ -49,6 +49,11 @@ public static class BlazorAppBuilderExtensions
                 OnPrepareResponse = SetCacheHeaders
             };
 
+            if (env.IsDevelopment() && config.EnableAutoRebuilding)
```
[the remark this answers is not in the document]
SteveSandersonMS: To do that, we'd have to change the `app.UseBlazor` middleware to control this setting explicitly. Definitely open to changing the middleware config style at some point but would rather not introduce that just for this PR.

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/BuildEventsWatcher.cs
```diff
@@ -0,0 +1,174 @@
...
+    /// <summary>
+    /// Watches for Blazor project build events, starts new builds, and tracks builds in progress.
+    /// </summary>
+    internal class BuildEventsWatcher : IVsUpdateSolutionEvents2
+    {
+        private const string BlazorProjectCapability = "Blazor";
+        private readonly IVsSolution _vsSolution;
+        private readonly IVsSolutionBuildManager _vsBuildManager;
+        private readonly object mostRecentBuildInfosLock = new object();
+        private readonly Dictionary<string, BuildInfo> mostRecentBuildInfos = new Dictionary<string, BuildInfo>();
```
rynowak: This shold probably use `OrdinalIgnoreCase`. I wouldn't expect it to happen, but if the build manager used inconsistent casings (legal on windows) that would result in some tasks that hang around forever as 'in progress' builds.

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
SteveSandersonMS: Good point. I'll change the ordering of operations to avoid that issue.

tooling/Microsoft.VisualStudio.BlazorExtension/BlazorPackage.cs
```diff
@@ -4,15 +4,37 @@
...
 using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
 
 namespace Microsoft.VisualStudio.BlazorExtension
 {
     // We mainly have a package so we can have an "About" dialog entry.
     [PackageRegistration(UseManagedResourcesOnly = true)]
     [AboutDialogInfo(PackageGuidString, "ASP.NET Core Blazor Language Services", "#110", "112")]
     [Guid(BlazorPackage.PackageGuidString)]
+    [ProvideAutoLoad(UIContextGuids80.SolutionExists)]
```
rynowak: I think this is good enough for now. In general this is something to avoid, but the solution to avoiding it is fairly complex, and we'd only bother if we're really productionizing this feature.

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/ProcessUtils.cs
SteveSandersonMS: Good to know, thanks.

This is only a way of making sure that if you have multiple VS instances open and showing your project, we only ask the correct one to do the build. We could instead have VS write something like a random GUID into the build output to indicate its identity. But since the PID technique appears to work fine it's probably simpler.

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/AutoRebuildService.cs
```diff
@@ -0,0 +1,103 @@
...
+            AddBuildServiceNamedPipeServer();
+        }
+
+        private void AddBuildServiceNamedPipeServer()
+        {
+            Task.Factory.StartNew(async () =>
+            {
+                try
+                {
+                    using (var serverPipe = new NamedPipeServerStream(_pipeName, PipeDirection.InOut, NamedPipeServerStream.MaxAllowedServerInstances))
```
rynowak: There's a problem here with permissions, and I'm not sure how hardcore the VS folk are about this.

This is an escalation of privilege vector. If VS is running as admin, a non-admin can connect to this an cause VS to rebuild arbitrary projects . Or another user. 

The fix for it is to do this: https://github.com/dotnet/roslyn/blob/14aed138a01c448143b9acf0fe77a662e3dfe2f4/src/Compilers/Server/VBCSCompiler/NamedPipeClientConnection.cs#L93

and this: https://github.com/dotnet/roslyn/blob/14aed138a01c448143b9acf0fe77a662e3dfe2f4/src/Compilers/Server/VBCSCompiler/NamedPipeClientConnection.cs#L160

src/Microsoft.AspNetCore.Blazor.Server/BlazorAppBuilderExtensions.cs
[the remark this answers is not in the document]
rynowak: Sounds good.

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/BuildEventsWatcher.cs
```diff
@@ -0,0 +1,174 @@
...
+            lock (mostRecentBuildInfosLock)
+            {
+                if (mostRecentBuildInfos.TryGetValue(projectPath, out var existingInfo))
+                {
+                    // If there's a build in progress, we'll join that even if it was started
+                    // before allowExistingBuildsSince, because it's too messy to cancel
+                    // in-progress builds. On rare occasions if the user is editing files while
+                    // a build is in progress they *might* see a not-latest build when they
+                    // reload, but then they just have to reload again.
+                    var acceptBuild = !existingInfo.TaskCompletionSource.Task.IsCompleted
```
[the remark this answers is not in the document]
SteveSandersonMS: I was thinking that, in the future, for failed builds we could capture the build error message and that would be stored on the task result. Then we'd use that task and be able to show the build error in the browser. The task would get replaced when the developer edits a file again and reloads, or manually triggers a build. So it was intentional to accept failed builds here if they are the most "up to date" reflection of the state of your app.

Does that seem reasonable, or do you think the dev experience would be better if we changed this? I thought it was pointless to request another build until the source code has been edited again.

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/BuildEventsWatcher.cs
SteveSandersonMS: OK will do

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/AutoRebuildService.cs
SteveSandersonMS: Thanks for pointing this out!

If I'm understanding this correctly, it seems like we only need to do the first one of those things now. The second one was a workaround for .NET Core limitations:

> // The overload of NamedPipeServerStream with the PipeAccessRule
> // parameter was removed in netstandard

... but that limitation seems to have done. The constructor is there, probably since .NET Core 2.0.

Do you think the following is therefore sufficient? https://github.com/aspnet/Blazor/pull/531/files#diff-e51aea7430ab7fad3d28c4afdb59db00R53

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/ProcessUtils.cs
rynowak: Yeah I think the approach here is the right one. I just wanted to point out that this only works because VS uses P/Inovkes to launch the process. Anyone writing tools on .NET is familiar with this old problem 😢

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+                compilationTaskMustBeRefreshedIfNotBuiltSince = DateTime.Now;
+            });
+
+            appBuilder.Use(async (context, next) =>
+            {
+                try
+                {
+                    if (compilationTaskMustBeRefreshedIfNotBuiltSince.HasValue)
+                    {
+                        var rebuildIfNotBuiltSince = compilationTaskMustBeRefreshedIfNotBuiltSince.Value;
```
rynowak: There are still numerous race conditions here, I don't think it's a big problem because the exceptions are going to be handled, but any truly correct solution is going to involve a lock or loop since you have two resources.

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/AutoRebuildService.cs
```diff
@@ -0,0 +1,122 @@
...
+            //   1. Send a "protocol version" number to the client
+            //   2. Receive the project path from the client
+            //      If it is the special string "abort", gracefully disconnect and end
+            //      This is to allow for mismatches between server and client protocol version
+            //   3. Receive the "if not built since" timestamp from the client
+            //   4. Perform the build, then send back the success/failure result flag
+            // Keep in sync with VSForWindowsRebuildService.cs in the Blazor.Server project
+            // In the future we may extend this to getting back build error details
+            await stream.WriteIntAsync(_protocolVersion);
+            var projectPath = await stream.ReadStringAsync();
```
rynowak: This also needs to check if the client is running as admin. That was the second link I provided.

rynowak: or rather, it should check if the client is at the same privilege level

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/BuildEventsWatcher.cs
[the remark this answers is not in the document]
rynowak: So for now, the experience when you have an error will be that we just don't get new content? I guess some things might be updated and some not.

I think this is good enough for preview2.

rynowak:
FYI Steve working on fixes for the two issues I pointed out.

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/AutoRebuildService.cs
rynowak: It would be good to check if the user is at the same permission level. I don't think ACLs check this. Some users run VS as admin, and in theory any process running as a normal user could send data to our pipe.

rynowak: Note that they also compare SIDs which covers the per-user thing. I'm just adding the check for elevation.

rynowak:
Rerunning travis, looks like a transient selenium failure

tooling/Microsoft.VisualStudio.BlazorExtension/AutoRebuild/BuildEventsWatcher.cs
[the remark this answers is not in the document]
SteveSandersonMS: Yes. Typically things won't have updated after a build failure because the compiled assembly represents the last successful build.

Later on I do think we should try to get the build errors to show up in the browser, otherwise this might be confusing.

SteveSandersonMS:
Thanks for the updates @rynowak - that looks great. Now merged.

PR #538, opened 2018-04-11
Update version numbers to 0.2 in preparation for 0.2

rynowak:
/cc @SteveSandersonMS 

There were a whole bunch of references still to 0.1

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,7 +5,6 @@ install:
 branches:
   only:
   - dev
-  - feature/0.1.0
   - /^release\/.*$/
   - /^(.*\/)?ci-.*$/
 environment:
--- a/src/Microsoft.AspNetCore.Blazor.Build/targets/RazorCompilation.targets
+++ b/src/Microsoft.AspNetCore.Blazor.Build/targets/RazorCompilation.targets
@@ -24,7 +24,15 @@
     <!-- Deactivates the Razor SDK's build-time compilation. We do our own -->
     <RazorCompileOnBuild>false</RazorCompileOnBuild>
 
-    <RazorDefaultConfiguration>Blazor-0.1</RazorDefaultConfiguration>
+    <!-- 
+      This version is used to build our RazorConfiguration, and is hardcoded into the tools in VS. If you change
+      This to a value that doesn't match VS, then the Blazor Razor experience won't work for those documents.
+    -->
+    <BlazorLanguageVersion>0.1</BlazorLanguageVersion>
+    <BlazorDefinitionConfiguration>Blazor-$(BlazorLanguageVersion)</BlazorDefinitionConfiguration>
+    <BlazorDeclarationConfiguration>BlazorDeclaration-$(BlazorLanguageVersion)</BlazorDeclarationConfiguration>
+    <RazorDefaultConfiguration>$(BlazorDefinitionConfiguration)</RazorDefaultConfiguration>
+    
   </PropertyGroup>
 
   <!-- Something quick for input/output tracking - the assumptions here should match what the CLI does -->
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNetCore.Blazor.Browser.JS/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blazor.host.client",
-  "version": "0.0.1",
```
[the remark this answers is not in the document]
SteveSandersonMS: We don't ever need to change this, as we're not shipping it as an NPM package. The version number isn't used anywhere, but there does have to be some semver-legal number there otherwise NPM tools complain.

src/Microsoft.AspNetCore.Blazor.Build/targets/RazorCompilation.targets
```diff
@@ -24,7 +24,7 @@
     <!-- Deactivates the Razor SDK's build-time compilation. We do our own -->
     <RazorCompileOnBuild>false</RazorCompileOnBuild>
 
-    <RazorDefaultConfiguration>Blazor-0.1</RazorDefaultConfiguration>
+    <RazorDefaultConfiguration>Blazor-0.2</RazorDefaultConfiguration>
```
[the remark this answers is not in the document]
SteveSandersonMS: From my point of view either option sounds reasonable, as arguably both still give you a "degraded experience" if the tooling and project don't match (i.e., some things won't work or will be inconsistent).

If changing the version number makes it more obvious that people need to upgrade tooling I can see the case for doing that. But either way we're going to have to be clear in all our docs and blog posts that people need to update their tooling and we don't do backwards-compatibility during this experimental phase.

.appveyor.yml
```diff
@@ -5,7 +5,6 @@ install:
 branches:
   only:
   - dev
-  - feature/0.1.0
```
SteveSandersonMS: Thanks for finding all these!

src/Microsoft.AspNetCore.Blazor.Build/targets/RazorCompilation.targets
```diff
@@ -39,18 +39,18 @@
     <ProjectCapability Include="Blazor" />
     <ProjectCapability Include="DotNetCoreRazorConfiguration" />
     
-    <RazorConfiguration Include="Blazor-0.1">
-      <Extensions>Blazor-0.1;Blazor.AngleSharp-0.1;$(CustomRazorExtension)</Extensions>
+    <RazorConfiguration Include="Blazor-0.2">
+      <Extensions>Blazor-0.2;Blazor.AngleSharp-0.2;$(CustomRazorExtension)</Extensions>
```
SteveSandersonMS: To ease future updates, should we be moving all these version numbers out into a property in some shared location?

tooling/Microsoft.VisualStudio.LanguageServices.Blazor/BlazorProjectEngineFactory.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.VisualStudio.LanguageServices.Blazor
 {
-    [ExportCustomProjectEngineFactory("Blazor-0.1", SupportsSerialization = false)]
+    [ExportCustomProjectEngineFactory("Blazor-0.2", SupportsSerialization = false)]
```
SteveSandersonMS: TBH I'm unclear why it's necessary to specify a version in places like this. It looks like an opaque string. If there was no version number to change here, wouldn't people just get whatever version they have installed? How does having a version number here help us?

src/Microsoft.AspNetCore.Blazor.Browser.JS/package.json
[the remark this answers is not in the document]
SteveSandersonMS: To avoid future confusion about the need for updates it would probably be better to revert `package.json` change. I only picked `0.0.1` rather than `1.0.0` to avoid the implication that it was a finished product.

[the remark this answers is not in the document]
rynowak: OK good to know. I will undo this part of the change. If only we could put a comment here.....

tooling/Microsoft.VisualStudio.LanguageServices.Blazor/BlazorProjectEngineFactory.cs
rynowak: You're right, it is an opaque string. This string matches what we're doing in MSBuild.

The value of `DefaultRazorConfiguration` from MSBuild flows through to the razor tools, and that filters `ICustomProjectEngineFactor` implementations based on these attributes.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorMetadata.cs
```diff
@@ -15,7 +15,7 @@ public static class Bind
         {
             public static readonly string RuntimeName = "Blazor.None";
 
-            public readonly static string TagHelperKind = "Blazor.Bind-0.1";
```
rynowak: I don't think there's any value in versioning these things. I copied this design from Razor, but it predates some other (better) decisions we made about versioning there.

tooling/Microsoft.VisualStudio.LanguageServices.Blazor/BlazorProjectEngineFactory.cs
rynowak: I've added comments to these places explaining that it's a wire-up from MSBuild.

SteveSandersonMS:
@rynowak The `release/0.2.0` branch is now made, so when you merge this, can you merge it into `dev` and also cherry-pick to `release/0.2.0`?

rynowak:
@SteveSandersonMS sure

PR #540, opened 2018-04-11
Template/sample update

guardrex:
* Found more single quotes to update.
* Found a couple of spots where there wasn't a blank line between the directive(s) and the first line of markup at the top of Razor files.


rynowak:
Looks fine to me, I'll merge when CI passes

PR #541, opened 2018-04-11
[WIP] Update templates

rynowak:
This is work in progress right now. We'll need to update again once
there's a build with the Action fix in it, and again to point to the
final package versions.

Just getting things staged ahead of time.


SteveSandersonMS:
I know this is not the final version for 0.2.0, but I'm trying to reconcile this PR, #465, and further template updates needed to make it show up for both 2.0 and 2.1 project types. So to get everything consistent I'm going to merge this WIP as-is, and we can then change the package versions further later. Hope that's OK!

SteveSandersonMS:
Merged

rynowak:
Sounds good!

PR #542, opened 2018-04-11
Add support for Action event handlers

rynowak:
This change adds `Action` to the set of types that have an overload on
RenderTreeBuilder. Additionally, we special case `Action` in the runtime
because passing the event args via DynamicInvoke() would throw.

Finally, reverted some of the clutter introduced by the first pass of
the event handler feature.


rynowak:
This ended up being really straightforward. Will merge when the CI passes.

rynowak:
BTW @SteveSandersonMS - I think the support for async event handlers would look just like this, but with some more special casing in the rutime. I'm not too worried about the language part of that feature.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/RuntimeCodeGenerationTest/EventHandler_OnElement_ArbitraryEventName_WithEventArgsMethodGroup/TestComponent.codegen.cs
```diff
@@ -0,0 +1,32 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Blazor;
+    public class TestComponent : Microsoft.AspNetCore.Blazor.Components.BlazorComponent
+    {
+        #pragma warning disable 1998
+        protected override void BuildRenderTree(Microsoft.AspNetCore.Blazor.RenderTree.RenderTreeBuilder builder)
+        {
+            base.BuildRenderTree(builder);
+            builder.OpenElement(0, "input");
+            builder.AddAttribute(1, "onclick", Microsoft.AspNetCore.Blazor.Components.BindMethods.GetEventHandlerValue<Microsoft.AspNetCore.Blazor.UIMouseEventArgs>(OnClick));
```
SteveSandersonMS: Cool. So according to the comment at https://github.com/dotnet/roslyn/pull/6642#issuecomment-297493700, methodgroup-to-delegate conversions will be cached in C# 8. Until then we're allocating an `Action` for every event handler on every render in the best case (noncapturing lambda or static methodgroup), or two for capturing lambdas that need an object to hold hoisted variables.

Previously in one of our meetings I said I saw the C#7-compiled code already caching converted delegates. On more investigation it turns out that happens *only* if the target method is `static` and is in an explicit lambda. For example, `MethodThatTakesAnAction(() => Console.WriteLine("Hello"))` will be cached, but `MethodThatTakesAnAction(() => Console.WriteLine(this.SomeProperty))` won't, and neither will `MethodThatTakesAnAction(SomeStaticMethod)`. Hopefully this just gets better for us in C#8.

So for now, there's probably no reasonable way for us to improve on this. Theoretically we could give more support to the explictly-passed `Action` instance case to be sure we don't allocate anything in that case, so developers could manually cache their own delegates if they wanted. But we really don't want to be telling people to do that, as the experience would suck. *Maybe* component library authors might want to do it in some perf-critical scenarios, but perhaps we can wait and see on that one. Maybe my obsession over not allocating during render might turn out to be pointless, though I am glad we are keeping it very limited.

rynowak: > Maybe my obsession over not allocating during render might turn out to be pointless

Avoiding allocations is always the right decision 👍

PR #545, opened 2018-04-11
Show templates in VS for ASP.NET 2.0 and 2.1

SteveSandersonMS:
Supersedes #465 

Note that whichever of ASP.NET Core 2.0 or 2.1 the developer picks, they still get a `global.json` that locks them to SDK `2.1.300-preview1-008174` (i.e., the released build of preview 1). When I checked, it works fine for both, but @rynowak, do you have any reason to believe this is wrong?

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/.template.config/template.json
+++ b/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/.template.config/template.json
@@ -73,10 +73,14 @@
         {
           "choice": "netcoreapp2.0",
           "description": "Target netcoreapp2.0"
+        },
+        {
+          "choice": "netcoreapp2.1",
+          "description": "Target netcoreapp2.1"
         }
       ],
-      "replaces": "netcoreapp2.0",
-      "defaultValue": "netcoreapp2.0"
+      "replaces": "netcoreapp2.1",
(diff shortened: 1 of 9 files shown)


rynowak:
> Note that whichever of ASP.NET Core 2.0 or 2.1 the developer picks, they still get a global.json that locks them to SDK 2.1.300-preview1-008174 (i.e., the released build of preview 1).

In general SDKs are backwards compatible with all runtimes, so the newest SDK should be able to build 2.0 and 2.1 apps.

However, we expect users of 0.2 to use SDK 2.1.300-preview2-XXXXXX (whatever the preview2 release number is). Do we create this global.json or does the template engine do it?

SteveSandersonMS:
> Do we create this global.json or does the template engine do it?

We create it. It's not based on what the developer actually has installed; it's just hard-coded. It wouldn't really be right to determine it dynamically: just because one developer has a certain SDK installed doesn't mean others on their team do.

So I think we can only put in a preview 2 version number once preview 2 ships with a known final version string.

rynowak:
The version number for preview2 SDK (as it currently stands) is 2.1.300-preview2-008530

I'm making a point of this, because using the preview2 sdk allows us to remove additional reference to the Razor SDK from the templates, which I already did 😆

SteveSandersonMS:
OK, the `global.json` files now specify `2.1.300-preview2-008530`.

PR #551, opened 2018-04-12
Add async events

rynowak:
Implementation of async event handlers

There are lots of changes here due to:
- removal of `UIEventHandler`
- update to C#7.3 and impact on tests of that change


Directory.Build.props
```diff
@@ -15,6 +15,7 @@
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)build\Key.snk</AssemblyOriginatorKeyFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AssemblySigningCertName>Microsoft</AssemblySigningCertName>
+    <LangVersion>7.3</LangVersion>
```
rynowak: Overloading between `Action` and `Func<Task>` **with** method group to delegate conversion will require C# 7.3. In practice that shouldn't be a problem since we're requiring everyone to have a version of the tools that support it. Thoughts?

src/Microsoft.AspNetCore.Blazor/Components/IHandleEvent.cs
```diff
@@ -15,6 +16,6 @@ public interface IHandleEvent
         /// </summary>
         /// <param name="handler">The event handler.</param>
         /// <param name="args">Arguments for the event handler.</param>
-        void HandleEvent(UIEventHandler handler, UIEventArgs args);
+        Task HandleEvent(Func<UIEventArgs, Task> handler, UIEventArgs args);
```
rynowak: I think we should consider dropping the nominal types for delegates and using `Action<T>` and `Func<T, Task>`. I think we'll end up with 6-7 event types minimum which means that we would be defining about 12-14 nominally-typed delegates.

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
```diff
@@ -121,25 +123,53 @@ internal void AssignEventHandlerId(ref RenderTreeFrame frame)
...
+               _eventHandlersById.Add(id, (UIEventArgs e) => 
+               {
+                   var value = @delegate.DynamicInvoke(e);
+                   if (value is Task task)
+                   {
+                       return task;
+                   }
+
+                   return null;
+               });
```
rynowak: One idea that I had about this ugliness, there are basically 4 types of signatures we want to support. We can abstract this behind a struct `EventBinding` so that we have an exchange type we can pass to `DispatchEvent`. Then we don't need to allocate these wrapper delegates to make everything callable via a `Func<UIEventArgs, Task>` without any allocations - since the struct will do that for us.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -169,14 +169,26 @@ private void HandleException(Exception ex)
...
             // This just saves the developer the trouble of putting "StateHasChanged();"
             // at the end of every event callback.
             StateHasChanged();
+
+            if (task.Status == TaskStatus.RanToCompletion)
+            {
+                return task;
+            }
+
+            return task.ContinueWith(OnAsyncTaskCompleted);
```
rynowak: Was this the flow that you had in mind? In the sample I updated things 'just work' without any calls to `StateHasChanged` in user code.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -169,14 +169,26 @@ private void HandleException(Exception ex)
...
+        }
+
+        protected virtual void OnAsyncTaskCompleted(Task task)
```
rynowak: We can also turn this into another extension point `OnAsyncTaskFaulted(Task)` that would be called in the case of failures. I'm not really convinced that this is useful at all really. Because you could just try/catch in your own code.

I also resisted the urge to create and register these tasks in any sort of global queue. This means that you could have a situation where a task is running while a component is removed from the render tree. I'm not sure how sophisticated we want to get with that sort of thing.

SteveSandersonMS: I agree that `OnAsyncTaskFaulted` would not be needed, and would make the same case about `OnAsyncTaskCompleted` (at least, that it should be private).

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
rynowak: I don't think we'll be able to get rid of this cast ladder totally. I'll wait to speculate more until I try it out.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
rynowak: Note that this calls `StateHasChanged` twice if you actually do async work.

SteveSandersonMS: Yes, once synchronously when the task started, and once when the task has finished, right? That's the desired effect.

rynowak: Ok great, my memory was correct then 👍

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
SteveSandersonMS: Ah yes, I agree we most likely shouldn't control that at the framework level. If a component is removed from the render tree (and hence disposed) while an async operation is in progress, we don't know whether or not that means the task should be cancelled. If it's a "Save" operation, for example, it shouldn't be cancelled, whereas if it's a "Fetch some new data to display", it could be cancelled. It's up to the developer to put in their own logic if they want to cancel.

Note that by design it's not harmful to attempt rendering after disposal. The renderer will just ignore the call when it sees the component is already disposed.

rynowak:
@SteveSandersonMS - updated

Let me know if what you think about the changes in 638f52a - I had avoiding allocations in mind, but it kind just moves complexity from one place to another.

src/Microsoft.AspNetCore.Blazor.Browser.JS/package-lock.json
```diff
@@ -1830,15 +1830,6 @@
         "xtend": "4.0.1"
       }
     },
-    "string_decoder": {
```
rynowak: WIll undo

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorApi.cs
```diff
@@ -84,19 +84,12 @@ public static class BindMethods
 
             public static readonly string GetEventHandlerValue = "Microsoft.AspNetCore.Blazor.Components.BindMethods.GetEventHandlerValue";
 
-            public static readonly string SetValue = "Microsoft.AspNetCore.Blazor.Components.BindMethods.SetValue";
```
rynowak: Unused

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Server/BlazorHosted.CSharp.Server.csproj
```diff
@@ -7,6 +7,7 @@
       https://api.nuget.org/v3/index.json;
       https://dotnet.myget.org/f/blazor-dev/api/v3/index.json;
     </RestoreSources>
+    <LangVersion>7.3</LangVersion>
```
rynowak: This isn't strictly needed in this project, but it seemed like a better choice to be consistent.

src/Microsoft.AspNetCore.Blazor/Components/BindMethods.cs
```diff
@@ -108,52 +127,6 @@ public static UIEventHandler SetValueHandler<T>(Action<T> setter, T existingValu
             };
         }
 
-        /// <summary>
-        /// Not intended to be used directly.
-        /// </summary>
-        public static Action<object> SetValue(Action<string> setter, string existingValue)
```
rynowak: Unused

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -175,14 +175,21 @@ private void HandleException(Exception ex)
...
             // This just saves the developer the trouble of putting "StateHasChanged();"
             // at the end of every event callback.
             StateHasChanged();
+            
+            if (task.Status == TaskStatus.RanToCompletion)
+            {
+                return;
+            }
+
+            task.ContinueWith(ContinueAfterLifecycleTask);
```
rynowak: It seemed like a good idea to to do this the same way as the other async things on `BlazorComponent` - which I overlooked before. Let me know if you think is this wrong.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -175,14 +175,21 @@ private void HandleException(Exception ex)
             Console.Error.WriteLine($"[{ex.GetType().FullName}] {ex.Message}\n{ex.StackTrace}");
         }
 
-        void IHandleEvent.HandleEvent(UIEventHandler handler, UIEventArgs args)
+        void IHandleEvent.HandleEvent(EventBinding binding, UIEventArgs args)
```
SteveSandersonMS: Not keen on the name `binding` or `EventBinding`, because the word "binding" is so overloaded and it's hard to tell whose notion of "binding" this refers to.

Based on what the implementation of `EventBinding` does, might `EventHandlerInvoker` or `EventDelegateInvoker` be a clearer name?

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
SteveSandersonMS: Yes, this looks good!

test/testapps/BasicTestApp/HttpClientTest/CookieCounterComponent.cshtml
```diff
@@ -17,22 +17,28 @@
     string testServerBaseUrl;
     string responseText;
 
-    async void DeleteCookie()
+    async Task DeleteCookie()
     {
         await DoRequest("api/cookie/reset");
-        StateHasChanged();
+        await Task.Delay(5000);
```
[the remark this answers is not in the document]
rynowak: Ahhhh oopsies, yeah I meant to revert this. This was a sample I was playing around with.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
rynowak: Good feedback. I'll go with `EventHandlerInvoker`.

Issue #552, opened 2018-04-12
Add compiler error if there's a <script> element inside a component


SteveSandersonMS:
This addresses a usability issue. We often hear of people trying to put `<script>` elements inside components and being surprised about the effects.

It's almost always a mistake to put a `<script>` element inside a component, because fundamentally, components are about adding and removing things dynamically inside the DOM. `<script>` elements don't work that way, in that once they are added, their effects can never be removed. Similarly, trying to use C# expressions inside a `<script>` content will not work as expected (we can't replace the effects of already-registered JS once the expression changes value). The *only* scenario I can think of where a `<script>` element does something vaguely useful inside a component would be for some kind of debug logging, e.g., `<script>console.log('rendering now')</script>`, but that's not what anyone's been doing AFAIK. What developers really should do is put `<script>` elements into their `index.html`, where they will behave as expected.

**Proposed solution**

We'll make it a compiler error to have `<script>` inside a Blazor component. The error message will say something like:

*Script tags should not be placed inside components because they cannot be updated dynamically. To fix this, move the script tag to the 'index.html' file or another static location. For more information see http://some/link*

... and the help docs that we linked to will also say something like:

*If you need to override this, add an attribute named "suppress-error" with value "BL9992" to the <script> tag, but only if you know what you are doing, as in most cases it would be a mistake to do this*.

SteveSandersonMS:
No, definitely an error! You can't have local scripts, that's the whole point 😃

tjbortz1s:
Lets say I want to write a "Canvas" component that has a bunch of Csharp calls to javascript functions.  Would there be a way to go about doing that?

SteveSandersonMS:
@tjbortz1s Yes, use the existing .NET->JS interop APIs. This doesn't involve putting a `<script>` tag inside your component. `<script>` tags go in `index.html`.

aguacongas:
@SteveSandersonMS it could be an issue to integrate [web components](https://www.webcomponents.org/) if we cannot use local script, what you think ?  
Are you going to provide a way to create blazor components library with a mecanism to auto load those library's javascripts ?

SteveSandersonMS:
@aguacongas Yes, we’ve already implemented a mechanism for creating redistributable packages of components that can load custom JS automatically. There’ll be full info about that in the 0.2.0 release blog post and docs, which we’re now targeting for Monday :smile:

aguacongas:
great

dotnetnoobie:
[...]

PR #553, opened 2018-04-12
Usability improvement: Block script tags in components, giving explanation

SteveSandersonMS:
Implementation for #552


SteveSandersonMS:
**STILL TO DO**: Replace the fake link in the error message with a real fwlink.

SteveSandersonMS:
I propose we include this in 0.2.0.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
```diff
@@ -304,6 +306,26 @@ public override void WriteHtmlContent(CodeRenderingContext context, HtmlContentI
...
+                const string suppressErrorAttributeName = "suppress-error";
+                if (string.Equals(tagToken.GetAttribute(suppressErrorAttributeName), "BL9992", StringComparison.Ordinal))
+                {
+                    tagToken.Attributes.RemoveAll(kvp => kvp.Key.Equals(suppressErrorAttributeName, StringComparison.Ordinal));
+                }
+                else
+                {
+                    var adjustedSpan = CalculateSourcePosition(node.Source, tagToken.Position);
+                    var diagnostic = BlazorDiagnosticFactory.Create_DisallowedScriptTag(adjustedSpan);
+                    throw new RazorCompilerException(diagnostic);
```
rynowak: I think this is fine for now, but keep in mind that any time you throw and catch in the compiler, that means that someone hitting the error will lose all intellisense and colorization in the rest of the file after this point. If at all possible, we should just skip the offending tag and its contents.

SteveSandersonMS: Is that the case for code in the runtime node writer as well as the design-time one?

rynowak: The problem with the runtime writer is that you won't run through the rest of the file, so if you throw here it will suppress any other errors in the file.

I don't have a problem with doing this as a temporary fix. I have a todo to remove all of these and the exception.

sjroe:
Note that (as expected) this doesn't prevent a script tag from being when writing a component directly without Razor.

Issue #495, opened 2018-04-05
Add OnRenderComplete lifecycle method


muqeet-khan:
Scenario:

ExampleComponent:

```
<div id="@DivId">
</div>
@functions(){
    public string DivId {get;set;} = "componentDiv";
    protected override void OnParametersSet()
    {
        RegisteredFunction.Invoke<bool>("initializeSomeJSFunction", new[] { DivId });
    }
}
```

**If** the `initializeSomeJSFunction` in JS wants to access the div `componentDiv` at this point, it can't as the component has not yet rendered and is therefore not accessible by document.getElementById('componentDiv'). This is blocking porting existing JS libraries to Blazor, if the library requires that context. 

If there was an override when the render is completed, then, I assume the elements will be visible to the DOM and can be used within the registeredFunction. If this is by design, what is the recommended approach to tackle a scenario like this?

SteveSandersonMS:
I agree. It's been in the back of my mind for a while that we'll need this. I agree it's important for integrating with external JS libraries.

We also need some way of giving the Blazor component a way of referencing its own DOM element in this callback.

SteveSandersonMS:
One important caveat is that this "After render" callback is going to be called *asynchronously*. That is, the browser will have actually painted its DOM elements and *then* we will call your "after render" callback (usually just about instantly). This is inevitable if we want to run the .NET code in a WebWorker, since all communication between the host frame and the web worker is async.

@muqeet-khan Do you think that will be OK for your scenarios?

muqeet-khan:
@SteveSandersonMS  Yes, as long as the elem is visible to the DOM at that moment, that's all I need. 

As a side note, here's another crazy land option I landed in yesterday:

JS:
```
Blazor.RegisterFunction('InitializeJS',function(data){
let elem = document.createElement('p')
elem.setAttribute('id','myNewParagraph');
//also do something useful with data and append it to elem perhaps
return elem;
});
```

MyComponent.cshtml:
```
<div>
@Html.Raw(_modifiedBlob)
</div>
@functions{
 public string SomeBlobOfData {get;set;} = "";
  private string _modifiedBlob {get; private set;} = "";
 protected override OnInit()
 {
   var result = RegisteredFunction.Invoke<RenderFragment\HTMLContent\string>('initializeJS');
   // this is trying to refer to the current component. Or may be the render tree?   
   //this.appendDomElem(result);
   // or store in some string and display the string as Raw in the component somewhere.
[...]

PR #555, opened 2018-04-12
Added OnRenderComplete

sjroe:
Allows a Component to be notified when its content has been added to the DOM.

Addresses #495

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -85,6 +85,19 @@ namespace Microsoft.AspNetCore.Blazor.Components
             => null;
 
         /// <summary>
+        /// Method invoked when the component has rendered into the DOM.
+        /// </summary>
+        protected virtual void OnRenderComplete()
+        {
+        }
+
+        /// <summary>
+        /// Method invoked when the component has rendered into the DOM.
+        /// </summary>
+        protected virtual Task OnRenderCompleteAsync()
+            => null;
+
+        /// <summary>
         /// Notifies the component that its state has changed. When applicable, this will
         /// cause the component to be re-rendered.
         /// </summary>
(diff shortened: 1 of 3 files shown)


muqeet-khan:
@SteveSandersonMS if this PR is ready can it be included in 0.3? Will be very helpful.

AWulkan:
Sorry if this is the wrong plae to ask, but do these lifecycle methods follow any specific logical naming scheme? For example in Vue you have:

> beforeCreated
> created
> 
> beforeMount
> mounted
> 
> beforeUpdate
> updated
> 
> beforeDestroy
> destroyed

Which makes it vey easy to remember the names for them.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -84,6 +84,19 @@ protected virtual void OnParametersSet()
...
+        /// </summary>
+        protected virtual void OnRenderComplete()
+        {
+        }
+
+        /// <summary>
+        /// Method invoked when the component has rendered into the DOM.
+        /// </summary>
+        protected virtual Task OnRenderCompleteAsync()
+            => null;
```
galvesribeiro: Wouldn't be best to return `Task.CompletedTask` instead of `null`? That avoid unnecessary null check(s) and don't allocate an extra `Task` either.

SteveSandersonMS:
Thanks @sjroe for contributing this!

As I looked into it, I found a different approach that avoids the need to track a list of callbacks. Plus there were some other areas where we have patterns for extending the lifecycle methods that needed to be handled a little differently. In the end I implemented the feature in #691.

I hope you don't mind, but we'll merge #691 instead. This does provide exactly the feature you're looking for (I hope!). Is it OK if I close this PR, or do you think there's anything I've missed?

SteveSandersonMS:
The equivalent functionality was just merged to `dev`, so I'll mark this one closed as per above. Hope that's OK!

Issue #258, opened 2018-03-14
Extreme E2E test


SteveSandersonMS:
Now we have project templates and publishing, we should have a maximally end-to-end test that:

 * Installs the project template package
 * Creates a new project via `dotnet new`
 * Restores and builds, launches it in the dev server, verifies it works via browser automation
 * Check that live reloading works when a source file is edited (when live reloading is implemented)
 * Publishes
 * Launches it in the production server, verifies it works via browser automation

We have the majority of the infrastructure we need already (e.g., for starting up test servers and doing browser automation) but this will still be quite involved. There's prior art for installing `dotnet new` packages and using them from E2E tests in the aspnet/templating repo.

PR #559, opened 2018-04-13
Inject MSBuild properties into templates during build (e.g., Blazor dependency version)

### LINKED ISSUE #549: Unable to build source code
Steps to reproduce:-

1. clone the repo.
2. Vs2017 preview 3 is installed.
3. execute the build.cmd.
4. open the solution in Visual studio.
5. try to build.
**tried multiple time to clone the repo and build but no luck. The last thing is pending Reinstall, vs2017 preview again and try.**

github

SteveSandersonMS:
This means that:

 * We don't need to manually update the Blazor package versions in templates to match what we are building in CI - it will now happen on every build.
    * For local development, there's a new MSBuild property `<TemplateBlazorPackageVersion>` that controls your local template dependency version
 * We don't have to remember to remove the `RestoreSources` for final builds - that now happens automatically

More generally it means we can use `dotnet new` template features to update the output code based on the values of MSBuild properties evaluated during our CI builds.

Also this PR adds the Blazor MyGet feed to the main `sources.props` so that VS no longer has trouble restoring the packages for the templates (fixes #549).

### How it works

It's a bit complicated because there are multiple phases of build, packaging, and template execution. This is the technique we *wanted* to use for ASP.NET templates but couldn't at the time due to a `dotnet new` limitation that's since been fixed.

1. **During local development**, when you run the template projects directly from source, the Blazor package versions come from the property `$(TemplateBlazorPackageVersion)` whose value is in the templates project `Directory.Build.props` file. It's still up to us to update this manually occasionally, though we could set the value to `$(VersionPrefix)-$(VersionSuffix)-*` if we wanted.
2. **When `dotnet run` generates a new project either from CLI or in VS**, at that time it replaces the string `$(TemplateBlazorPackageVersion)` with the value of the 'dotnet new' symbol `TemplateBlazorVersionSymbol`, whose value is defined in that template's `.template.config\template.json` file. This value will be a literal string like `0.3.0-preview1-10302`, so that's what would end up in the generated `.csproj` on disk.
3. To get the CI's current package version into the `template.json` files, there's a target in `Microsoft.AspNetCore.Blazor.Templates.csproj` that, during the build, uses the `Sdk_GenerateFileFromTemplate` task from the ASP.NET SDK. It performs replacement on the contents of the `template.json` files. To play well with source control, our original template config files are now in directories called `.template.config.src`, and `.template.config` is generated from these (with substitutions) during the build.

In case you're wondering why we don't just combine steps 2 and 3 into a single step where we directly replace symbols in the `.csproj` files during CI build time (instead of deferring it to template execution time), the reason is that it means we can use all the `dotnet new` template features to perform more sophisticated replacements and conditional logic. For example, in this PR we also dynamically remove the `<RestoreSources>` property from the generated `.csproj` files based on a symbol computed from our `IsFinalBuild` MSBuild property in CI. It's not just about string replacements - it's also about modifying the structure of files in a way that's sensitive to the file type.

Now we're doing not-completely-trivial things to build our template package, it's an even stronger reason to implement #258 as well.


SteveSandersonMS:
I'd be happy to include this in 0.2.0 because, even though its mechanism is nontrivial, it's pretty easy to verify the the final output is correct just by looking at the generated `.nupkg` contents.

SteveSandersonMS:

src/Microsoft.AspNetCore.Blazor.Templates/Directory.Build.props
```diff
@@ -1,8 +1,7 @@
 <Project>
-  <!--
-    Left mostly blank to avoid sharing config with the build for the main repo.
-    For example, we want the templates to declare their own independent package restore sources.
-  -->
+  <PropertyGroup>
+    <TemplateBlazorPackageVersion>0.2.0-preview1-10195</TemplateBlazorPackageVersion>
```
rynowak: Consider defining this at the top level, or where our other versions are defined. Right now we have versions in a few places

src/Microsoft.AspNetCore.Blazor.Templates/Microsoft.AspNetCore.Blazor.Templates.csproj
```diff
@@ -7,4 +7,38 @@
...
   </PropertyGroup>
+
+  <Target Name="TransformTemplateConfigs" BeforeTargets="CoreBuild" DependsOnTargets="SetTemplateJsonSymbolReplacements">
+    <!--
+      For each template, copy its '.template.config.src' directory to '.template.config',
+      removing any earlier output at that location
+    -->
+    <ItemGroup>
+      <_TemplateConfigMainFile Include="content\**\.template.config.src\template.json" />
+      <_TemplateConfigDir Include="@(_TemplateConfigMainFile-&gt;'$([System.IO.Path]::GetDirectoryName('%(_TemplateConfigMainFile.FullPath)'))')" />
```
rynowak: `-&gt;` should this be `->`?

Probably doesn't matter to MSBuild but it surprised me

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted.CSharp/BlazorHosted.CSharp.Client/BlazorHosted.CSharp.Client.csproj
```diff
@@ -1,17 +1,19 @@
...
-    <RestoreSources>$(RestoreSources);https://dotnet.myget.org/f/blazor-dev/api/v3/index.json</RestoreSources>
+    <RestoreSources Condition="'$(IncludeCustomRestoreSourcesSymbol)'=='true'">
+      $(RestoreSources);
+      https://api.nuget.org/v3/index.json;
+      https://dotnet.myget.org/f/blazor-dev/api/v3/index.json;
+    </RestoreSources>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.1.0-preview2-final" PrivateAssets="all" />
```
rynowak: I still plan to remove this from the templates, but I won't get to it for a few days

PR #562, opened 2018-04-13
Enable GenerateDocumentationFile. Add/fix XML docs.

### LINKED ISSUE #558: Add XML doc files to Blazor packages
We need the XML doc files in the Blazor packages for intellisense and for API ref doc generation.

SteveSandersonMS:
Fixes #558


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorCodeDocItems.cs
```diff
@@ -1,11 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace Microsoft.AspNetCore.Blazor.Razor
-{
-    public static class BlazorCodeDocItems
```
SteveSandersonMS: This was not being used (AFAICT).

src/Microsoft.AspNetCore.Blazor/UIEventHandler.cs
```diff
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using Microsoft.AspNetCore.Blazor.RenderTree;
```
SteveSandersonMS: Needed because some of the `<see ...>` tags rely on it.

SteveSandersonMS:

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorCodeDocItems.cs
rynowak: 👍

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -11,32 +11,48 @@
 
 namespace Microsoft.AspNetCore.Blazor.Razor
 {
+    /// <summary>
+    /// A <see cref="DocumentClassifierPassBase"/> that recognizes Blazor components.
+    /// </summary>
     public class ComponentDocumentClassifierPass : DocumentClassifierPassBase, IRazorDocumentClassifierPass
```
rynowak: This should be internal instead, there's not scenario for user code to interact with this

src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
```diff
@@ -73,6 +73,7 @@ public void SetParameters(ParameterCollection parameters)
             _renderHandle.Render(Render);
         }
 
+        /// <inheritdoc />
```
rynowak: Note that this doesn't actually work (I don't think). Our docs stuff doesn't actually honour this

PR #564, opened 2018-04-13
Route only to component types in app assembly to avoid antipatterns now and to avoid a breaking change later

SteveSandersonMS:
This would be fixed when the CR feedback on #266 was finished anyway, but I want to be sure we include this in 0.2.0 regardless of that.

This is because when we ship the component libraries feature, people might get the wrong idea about how it's meant to interact with routing and start relying on logic that we plan to change (i.e., improve, but without this PR, it would be a breaking improvement).


SteveSandersonMS:
Changed my mind. We can still update this logic in the future without having to do this right now. If anyone has concerns about this, let me know.

Issue #472, opened 2018-04-03
Blazor client class Program not public


fileman:
the "Program" class created by new Blazor project (not hosted) not set as public on VS 15.7preview2 - Blazor 0.1.0 ... I can't find other issue for this, I hope is not a duplicate.

fileman:
even with myget 15.7.10160 vsix

danroth27:
Looks like this is only the case for the standalone Blazor template. 

@SteveSandersonMS I assume this isn't intentional?

fileman:
@danroth27 @SteveSandersonMS Solved with PR #574 if and when is merged

SteveSandersonMS:
@danroth27 I just checked, and the default new console app also *doesn't* use the `public` modifier on either the `Program` class or its `static Main` method. So to be consistent with that, don't you think we should instead *remove* the public modifier from the "Hosted" template?

SteveSandersonMS:
Oh wait, I remember why it’s there. In the hosted template, it has to be public so the server middleware can reference it. Yes let’s update the standalone in case people try to add it as a reference from a server project.

fileman:
Yes, this is my case, I've used standalone project to an existing webapi project

PR #574, opened 2018-04-16
Fix protection level in standalone template

fileman:
Set Program class and his Main method as public in standalone project template #472

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorStandalone.CSharp/Program.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorStandalone.CSharp/Program.cs
@@ -5,9 +5,9 @@ using System;
 
 namespace BlazorStandalone.CSharp
 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorStandalone.CSharp/Program.cs
```diff
@@ -5,9 +5,9 @@
 
 namespace BlazorStandalone.CSharp
 {
-    class Program
+    public class Program
     {
-        static void Main(string[] args)
+        public static void Main(string[] args)
```
SteveSandersonMS: Although the `Program` class should be `public` for consistency with `BlazorHosted`, the `Main` method should not (there's no need). If you could revert the change on line 10, we can merge the change on the other line :)

SteveSandersonMS:
Thanks @fileman! This is now merged into `dev`.

PR #575, opened 2018-04-16
AutoRebuild: System.ComponentModel.Win32Exception : 'Access denied'

Daddoon:
Current AutoRebuild implementation crash if trying to launch the app hosted in IIS.

Current found hierarchy:

> - svchost
> |
>    -- w3wp
>    |
>      -- VSIISExeLauncher
>      |
>        -- dotnet

devenv is not found of course in this scenario.

Even if the try catch seem to be good, when accessing the root process svchost, this exception is thrown:

> System.ComponentModel.Win32Exception : 'Access denied'

in the **'while'** statement

Due to this behavior, the try/catch need to be bubbled on the top of the while loop in order to have the correct behavior in case of fatal failure.


SteveSandersonMS:
Thanks @Daddoon - this looks great!

Daddoon:
Actually, i may commit an update in some hour, that allow to find the right devenv process from a debug in an hosted in IIS project, by reading the **IISExeLauncherArgs.txt** file, argument **-owningPid** .

Not finished yet.
I will open another pulling request if it's ok.

PR #579, opened 2018-04-16
Added VSIISExeLauncher autorebuild support

Daddoon:
Added **VSIISExeLauncher** autorebuild support.

When searching for devenv process in FindAncestorVSProcess method, if "VSIISExeLauncher" is found in the process tree, then we are likely in a debugable IIS hosted app.

Retrieving the initial launcher to find the owningPid number, actually the current devenv instance.

Returning devenv process if found and valid.

This behavior will fail of course if the current executing process lack of permissions, like not being in the administrator group.

In my case, added my current ApplicationPool in the Administrators group, then restarted IIS, was sufficient to allow to read devenv process.

Correct me if i'm wrong, but as the VSIISExeLauncher will be called only if launched from Visual Studio, this behavior should never happen in production.

**NOTE**: FindIISHostedVSProcess is called within FindAncestorVSProcess **while** loop if its relevant, however if you prefer by design that this method is called after FindAncestorVSProcess in TryCreate (if relevant), i will change this.


Daddoon:
Damn, actually commited something not related to this but to an other pull request. I will revert back shortly.

PR #583, opened 2018-04-17
SimpleJson: more informative exception when attempting to deserialize object with non-empty constructor

zbecknell:
This is in response to #578.  Although Simple JSON will possibly be replaced in the future, I figured for now this more informative exception message may prevent future issues from being created.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
@@ -1430,7 +1430,13 @@ namespace SimpleJson
                             obj = value;
                         else
                         {
-                            obj = ConstructorCache[type]();
+                            try{
+                                obj = ConstructorCache[type](); 
+                            }
+                            catch(NullReferenceException)
+                            {
+                                throw new InvalidOperationException($"Objects must have a constructorless parameter to deserialize. Offending type '{type}'");
+                            }
                             foreach (KeyValuePair<string, KeyValuePair<Type, ReflectionUtils.SetDelegate>> setter in SetCache[type])
                             {
                                 object jsonValue;
--- a/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
@@ -1430,12 +1430,13 @@ namespace SimpleJson
                             obj = value;
                         else
                         {
-                            try{
+                            try
+                            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -1430,7 +1430,13 @@ public virtual object DeserializeObject(object value, Type type)
                             obj = value;
                         else
                         {
-                            obj = ConstructorCache[type]();
+                            try{
```
SteveSandersonMS: The brace style convention for ASP.NET repo projects project is:

```
try
{
    ... some code ...
}
catch (...) // Note the space after 'catch'
{
    ... some code ...
}
```

Could you update the PR to match?

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -1430,7 +1430,13 @@ public virtual object DeserializeObject(object value, Type type)
...
+                                obj = ConstructorCache[type](); 
+                            }
+                            catch(NullReferenceException)
+                            {
+                                throw new InvalidOperationException($"Objects must have a constructorless parameter to deserialize. Offending type '{type}'");
```
SteveSandersonMS: Could we change the message to the following format?

    Cannot deserialize JSON into type '{type.FullName}' because it does not have a public parameterless constructor.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -1430,7 +1430,13 @@ public virtual object DeserializeObject(object value, Type type)
...
+                            }
```
SteveSandersonMS: Minor point, but please add a blank line after the closing try/catch brace.

SteveSandersonMS:
Oh one other thing, could you please add a unit test covering this scenario? There's already a class of unit tests for JSON handling at https://github.com/aspnet/Blazor/blob/dev/test/Microsoft.AspNetCore.Blazor.Test/JsonUtilTest.cs so it should be straightforward to add one further test there.

SteveSandersonMS:
Thanks @zbecknell! This is now merged.

I tweaked the logic a little further to deal correctly with the case where there is a parameterless constructor but it throws a `NullReferenceException` when executing. Hope that's OK.

zbecknell:
@SteveSandersonMS Cleaner and covers that scenario, very nice!

Issue #384, opened 2018-03-26
Avoid blazor-component wrappers for components that supply exactly one top-level node


chassq:
Looking at the Elements output in chrome I see several blazor-component nested tags and then I also see blazor-component tags surrounding some of the NavLinks. I think in some cases this will mess with css cascade for some existing bootstrap templates, for example, that have css like **li > a** vs **li > blazor-component > a**. Is there some reason for this blazor-componsnt tag in the dom when something like a data attribute could be set on the a tag instead maybe?

Suchiman:
You can find the explanation here https://github.com/aspnet/Blazor/blob/b64218fa31cc2fce3825875e5dd4eb8995b00c3e/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts#L137-L151

bdparrish:
Could the DOM element be added with the component id attribute?

`<span class="btn btn-success" data-blazor-id="b-abc123" ... other attributes ... >Blazor!</span>`

SteveSandersonMS:
The current short-term plan is to put in an optimisation where:

* If your component renders **exactly one** top-level element, then we'll not insert the `<blazor-component>` wrapper
* If your component renders 0 or 2+ top-level elements, then we will insert the `<blazor-component>` wrapper

This preserves the O(1) lookup during diff applications while giving developers an easy way to avoid the weird wrapper elements in most realistic cases. Normally it does make sense for a component to supply exactly one top-level element.

I know there are still some edge cases where you really want to return multiple top-level elements and really don't want a wrapper, but that's something we'd probably leave until further in the future to implement. It's certainly possible; we just need to do more complex things to keep the optimizations working.

chassq:
But is the wrapper around <a> tags a bug?

SteveSandersonMS:
No, it's currently by design, with the intention of removing it soon as described above.

chassq:
ok thanks!

bdparrish:
@SteveSandersonMS 

Trying to track down the reference for all this in the code base, but I am coming up empty.  Where is the complexity pieces for this?  Blazor.Browser.JS or Blazor project?

SteveSandersonMS:
@bdparrish I'm not totally certain what you're asking, but if you mean "where would this be implemented", it would be in `BrowserRenderer.ts`. It's probably best if I implement this myself as I already have 90% of it done in a private local branch (including tests) and am pretty clear about what's needed for the last 10%.

PR #586, opened 2018-04-17
In Razor compilation, trim leading and trailing whitespace nodes

SteveSandersonMS:
This is going to be needed for my approach to fixing #384, plus it's good in general not to be doing work for irrelevant whitespace.

Theoretically a person might actually want leading or trailing whitespace, but I think it's rare enough that we can just say you have to use `&nbsp;` or `<span> </span>` or something similar in that case. I can't think of any scenario where it's essential for components to emit leading/trailing text nodes that literally just contain spaces or linebreaks.


SteveSandersonMS:
In the end I found a way of fixing #384 without specifically needing this change, but I still think we should do this change because it's good to reduce the number of runtime operations.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TrimWhitespacePass.cs
```diff
@@ -0,0 +1,94 @@
...
+                    case HtmlContentIntermediateNode htmlContentIntermediateNode:
+                        RemoveContiguousWhitespace(htmlContentIntermediateNode.Children, direction);
+                        shouldRemoveNode = htmlContentIntermediateNode.Children.Count == 0;
+                        shouldContinueIteration = shouldRemoveNode;
+                        break;
+
+                    case CSharpExpressionIntermediateNode _:
+                        shouldRemoveNode = false;
+                        shouldContinueIteration = false; // Because it may produce nonwhitespace output
+                        break;
```
rynowak: Can you try something in VS with this change?

Write something like:
```
<div class=@( )>

</div>
```

Then add:
```
<div class=@( )>

</div>
@functions {
  string myclass;
}
```

The put your cursor here:
```
<div class=@(| )>

</div>
@functions {
  string myclass;
}
```

Do you get completion for `myclass` at that location?

I'm concerned that this is a case where design time will have to function differently

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TrimWhitespacePass.cs
```diff
@@ -0,0 +1,94 @@
...
+
+                    case CSharpCodeIntermediateNode codeIntermediateNode:
```
rynowak: What's the need to remove whitespace inside non-HTML? I'm not sure I understand why that part is important

SteveSandersonMS: We're not removing whitespace inside non-HTML. This `case` on line 59 is checking whether we hit a `CSharpCodeIntermediateNode` (other than the call to `base.BuildRenderTree`, which we skip). If we do, then we stop the trimming process because we've found something that at runtime might represent non-whitespace.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TrimWhitespacePass.cs
SteveSandersonMS: Yes, I do still get completion in that case.

I'm interested to know what you're suspicious might be broken though in case I'm not understanding the rules. How would removing some whitespace-only `IntermediateNode` instances from the beginning and end of the method change completion?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TrimWhitespacePass.cs
rynowak: Ok I see that now, I think I misread the code. I was looking at this pretty early in the morning, and I think I got about 4 hours of sleep 😴 

Are you only concerned about *HtmlContent* - or do you care about inside of attributes as well?


You could also consider using `RazorIntermediateNodeWalker` as well if it helps. I think what's here's is OK as it is, but you may need to handle some more cases. Have you looked at what happens when the code is inside a tag helper - inside the body of a component?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TrimWhitespacePass.cs
rynowak: OK great.

This concern was based on my earlier belief that you wanted to remove whitespace inside C#. The problem with that at design time is that we need a span in the generated code that maps to where your cursor is. If your cursor is inside C# whitespace then we need to generate that whitespace and make sure it matches so that the right completion will trigger.

If you are observing `documentNode.Options.DesignTime` from inside VS, that will **always** be the case. The code generation of the files we use in the build is driven by msbuild, which is a separate process.

You should be able to see `DesignTime == false` in any of the tests for runtime code generation.

https://github.com/aspnet/Blazor/blob/dev/test/Microsoft.AspNetCore.Blazor.Build.Test/RazorIntegrationTestBase.cs#L201

rynowak: I think it would be fine to skip over this detail, or to make this pass no-op for design time if you think that makes sense

rynowak:
Sorry, I've been commenting on this throughout the day while I've been doing other things. I'm also currently working on a change that will move the HTML parsing to the beginning of the IR phase, which won't conflict with this change.



I still have a concern related to this comment:
>You could also consider using RazorIntermediateNodeWalker as well if it helps. I think what's here's is OK as it is, but you may need to handle some more cases. Have you looked at what happens when the code is inside a tag helper - inside the body of a component?

SteveSandersonMS:
> Have you looked at what happens when the code is inside a tag helper - inside the body of a component

@rynowak Great point. Thanks. I've updated the code to account for this, plus test cases. I think that now covers all the scenarios, but please let me know if you suspect otherwise.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TrimWhitespacePass.cs
SteveSandersonMS: OK, I've now updated it to skip `TrimWhitespacePass` both during design-time builds *and* during the first phase of the two-phase compile, since it wasn't achieving anything useful during those compilations.

SteveSandersonMS:
BTW I'm not convinced the new line-endings treatment for baseline files is working as expected. Since I regenerated the baselines on Windows, Git is showing a lot of changes that appear to be line-ending-only.

rynowak:
> BTW I'm not convinced the new line-endings treatment for baseline files is working as expected. Since I regenerated the baselines on Windows, Git is showing a lot of changes that appear to be line-ending-only.

OK go ahead and write over them with the windows endings for now, and I'll take a look when I get a chance.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TrimWhitespacePass.cs
```diff
@@ -0,0 +1,100 @@
...
+                        break;
+
+                    case HtmlContentIntermediateNode htmlContentIntermediateNode:
+                        RemoveContiguousWhitespace(htmlContentIntermediateNode.Children, direction);
+                        shouldRemoveNode = htmlContentIntermediateNode.Children.Count == 0;
+                        shouldContinueIteration = shouldRemoveNode;
+                        break;
+
+                    case CSharpExpressionIntermediateNode _:
+                    case TagHelperIntermediateNode _:
```
rynowak: What about a case with child content? Is that covered?

```
<MyComponent>


  <button onclick="@foo">

      @bar

  </button>


</MyComponent>
```

In each of these cases, this will be lowered to a form like:
```
Method
  TagHelper
    ...
    TagHelperBody
      ...
      HtmlContent
      ...
```

SteveSandersonMS: In that scenario the algorithm terminates as soon as it sees the `TagHelperIntermediateNode` - the contents of that node don't matter. We're only trying to trim the top-level leading and trailing whitespace in the entire document.

rynowak: OK cool 👍

PR #602, opened 2018-04-18
Eliminate wrapper elements

### LINKED ISSUE #384: Avoid blazor-component wrappers for components that supply exactly one top-level node

SteveSandersonMS:
Fixes #384. In fact it goes further and eliminates the wrapper elements in *all* scenarios, not just when components produce exactly one top-level node.

This PR changes how `BrowserRenderer` works so that it no longer inserts the extra `blazor-component` elements around child components. This simplifies the DOM structure, but more importantly, simplifies working with CSS because you no longer have to account for the wrappers in your selectors. The DOM structure is now completely under the control of the developer. For details on how this works (and how it preserves O(1) traversal), see the comments inside `LogicalElement.ts`.

All future DOM mutation logic in `BrowserRenderer` must use the `LogicalElement` APIs (unstead of directly calling native DOM insertion/removal APIs) to retain compatibility.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
@@ -4,13 +4,14 @@ import { getTreeFramePtr, renderTreeFrame, FrameType, RenderTreeFramePointer } f
 import { platform } from '../Environment';
 import { EventDelegator } from './EventDelegator';
 import { EventForDotNet, UIEventArgs } from './EventForDotNet';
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
Merged because it's unlikely anyone would be requesting review changes, but @javiercn @rynowak please feel free to review if you wish.

masonwheeler:
Yay!  Keep up the good work!

PR #619, opened 2018-04-19
Bootstrap 4 and new styling

SteveSandersonMS:
This builds on @fabridamazio's work in #363

I've reimplemented the layout and styling rules to make it look shinier and adapt to different size screens better. This PR also implements a working "hamburger" button for toggling the nav UI on mobile screens.


SteveSandersonMS:
@danroth27 Are you happy with these changes to the third-party notices?

SteveSandersonMS:
@danroth27 Actually it's possible that the correct license for the Iconic icons is their "font license" (also included in this PR). We are redistributing their font files here. Do you think this is OK?

danroth27:
> Actually it's possible that the correct license for the Iconic icons is their "font license" (also included in this PR). We are redistributing their font files here.

@SteveSandersonMS I think you're right. We're not distributing the icons at all, correct?

SteveSandersonMS:
This PR only uses the font files and theu

masonwheeler:
> This PR also implements a working "hamburger" button for toggling the nav UI on mobile screens.

Can you do this as an externally-visible Property?  I've got some screens where having it around would be preferable, and others where I'd love to be able to hide the nav with a hamburger menu, even on a desktop.

SteveSandersonMS:
@danroth27 I'm not sure it's *only* the font license, since it sounds like that applies only to the font files, and we are also using one of their CSS files. I've included both licenses in the notices now. If you still think this is wrong please let me know and we can change it.

@masonwheeler The default template doesn't need any feature like that, and it's important to keep it streamlined. Of course, it's only a starting point, and developers are expected to modify it in whatever way is applicable to their projects.

danroth27:
@SteveSandersonMS 👍

PR #629, opened 2018-04-20
Added VSIISExeLauncher autorebuild support.

Daddoon:
(Re-opened pull, as i branched correctly my code)

Added **VSIISExeLauncher** autorebuild support.

When searching for devenv process in FindAncestorVSProcess method, if "VSIISExeLauncher" is found in the process tree, then we are likely in a debugable IIS hosted app.

Retrieving the initial launcher to find the owningPid number, actually the current devenv instance.

Returning devenv process if found and valid.

This behavior will fail of course if the current executing process lack of permissions, like not being in the administrator group.

In my case, added my current ApplicationPool in the Administrators group, then restarted IIS, was sufficient to allow to read devenv process.

Correct me if i'm wrong, but as the VSIISExeLauncher will be called only if launched from Visual Studio, this behavior should never happen in production.

**NOTE**: FindIISHostedVSProcess is called within FindAncestorVSProcess **while** loop if its relevant, however if you prefer by design that this method is called after FindAncestorVSProcess in TryCreate (if relevant), i will change this.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/VSForWindowsRebuildService.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/VSForWindowsRebuildService.cs
@@ -97,7 +97,7 @@ namespace Microsoft.AspNetCore.Blazor.Server.AutoRebuild
             try
             {
                 var vsProcess = Process.GetProcessById(owningPid);
-                if (vsProcess != null && !vsProcess.HasExited)
+                if (vsProcess != null && vsProcess.ProcessName.Equals("devenv", StringComparison.OrdinalIgnoreCase) && !vsProcess.HasExited)
                     return vsProcess;
             }
             catch (Exception)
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks for contributing this, @Daddoon! I appreciate the elegance and simplicity of what you've done here.

I've been trying it out, and it seems to me that the only scenario where this can work is if the IIS worker process is running as an administrator. That's because (1) VS can only communicate with IIS when VS is running elevated, and (2) this means IIS can only observe the elevated VS process if IIS itself is running the app as an administrator.

If that is the case, I'm not certain how to make use of this, assuming we don't recommend IIS worker processes run as administrator. @danroth27 @javiercn @rynowak - what do you think? Can we tell people they should run their dev-mode IIS worker process as administrator, or is it better to say that auto-rebuild isn't supported under IIS?

danroth27:
I don't think we should introduce a requirement to run as admin.

SteveSandersonMS:
OK, I thought so.

In that case, I don't think we will be able to support IIS for autorebuild, at least not without bigger changes that we might do in the future. @Daddoon, are you OK if we close this? Sorry it's not something we can do at this stage!

Daddoon:
No problems ! It's easily understandable. A next time maybe ;)

PR #630, opened 2018-04-20
Added binding support on double/float/decimal

### LINKED ISSUE #623: Elegant way to wrap binding for unsupported type
I get it, this is work in progress. So this will be solved in future I guess:

```
<input bind="@Amount" /><br/>

@functions
{
    public decimal Amount { get; set; }
}
``` 

Now throws:

`Error: System.ArgumentException: 'bind' does not accept values of type System.Decimal. To read and write this value type, wrap it in a property of type string with suitable getters and setters`

So I can make this work by adding this property (needs more failure checks, I know);

```
    public decimal Amount { get; set; }

    [NotMapped]
    public string AmountStr
    {
        get
        {
            return this.dec.ToString();
        }
        set
        {
            this.dec = Convert.ToDecimal(value);
        }
    }
```

And binding to AmountStr.

But if you are using (EF poco) objects this really pollutes my model classes.
Does anybody can suggest a elegant workaround for this?

Daddoon:
Added binding support on double/float/decimal, by adding corresponding delegates for theses use case.

**NOTE:** Binding parsed and showed value for decimal is correspond to current thread culture format (System.Threading.Thread.CurrentThread.CurrentCulture or System.Threading.Thread.CurrentThread.CurrentUICulture , doesn't have made deep test), so the ouput may be influenced and corresponding to culture requirement.

**Suggestion:** What about  System.OverflowException ? Actually mocking the previous implementation (btw doing nothing specially).

Fix #623

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/BindMethods.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BindMethods.cs
@@ -77,6 +77,30 @@ namespace Microsoft.AspNetCore.Blazor.Components
         /// <summary>
         /// Not intended to be used directly.
         /// </summary>
+        public static UIEventHandler SetValueHandler(Action<double> setter, double existingValue)
+        {
+            return _ => setter(double.Parse((string)((UIChangeEventArgs)_).Value));
+        }
+
+        /// <summary>
+        /// Not intended to be used directly.
+        /// </summary>
+        public static UIEventHandler SetValueHandler(Action<float> setter, float existingValue)
+        {
+            return _ => setter(float.Parse((string)((UIChangeEventArgs)_).Value));
+        }
+
+        /// <summary>
+        /// Not intended to be used directly.
+        /// </summary>
+        public static UIEventHandler SetValueHandler(Action<decimal> setter, decimal existingValue)
+        {
+            return _ => setter(decimal.Parse((string)((UIChangeEventArgs)_).Value));
+        }
+
+        /// <summary>
+        /// Not intended to be used directly.
+        /// </summary>
         public static UIEventHandler SetValueHandler(Action<DateTime> setter, DateTime existingValue)
         {
             return _ => SetDateTimeValue(setter, (object)((UIChangeEventArgs)_).Value, null);
@@ -125,6 +149,24 @@ namespace Microsoft.AspNetCore.Blazor.Components
         /// </summary>
         public static Action<object> SetValue(Action<int> setter, int existingValue)
             => objValue => setter(int.Parse((string)objValue));
+
+        /// <summary>
+        /// Not intended to be used directly.
+        /// </summary>
+        public static Action<object> SetValue(Action<double> setter, double existingValue)
+            => objValue => setter(double.Parse((string)objValue));
+
+        /// <summary>
+        /// Not intended to be used directly.
+        /// </summary>
+        public static Action<object> SetValue(Action<float> setter, float existingValue)
+            => objValue => setter(float.Parse((string)objValue));
+
+        /// <summary>
+        /// Not intended to be used directly.
+        /// </summary>
+        public static Action<object> SetValue(Action<decimal> setter, decimal existingValue)
+            => objValue => setter(decimal.Parse((string)objValue));
 
         /// <summary>
         /// Not intended to be used directly.
--- a/test/testapps/BasicTestApp/BindCasesComponent.cshtml
+++ b/test/testapps/BasicTestApp/BindCasesComponent.cshtml
@@ -11,6 +11,18 @@
     <span id="textbox-initially-populated-value">@textboxInitiallyPopulatedValue</span>
 </p>
 
+<h2>Numeric Stepper</h2>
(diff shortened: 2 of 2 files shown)


rynowak:
I think this is a good unblocker for now. We might want to consider a pattern that scales in the future.

SteveSandersonMS:
Thanks Ryan!

@Daddoon OK, this looks like the right approach to us. Would you be able to add test cases (e.g., in `BindCasesComponent.cshtml`)? After that we'd be very pleased to merge this.

Daddoon:
Done !

SteveSandersonMS:
Sorry for not being clear - we'd be looking for test cases for all three types you've added support for (`float`, `double`, `decimal`), not just `decimal`. As well as adding the additional UI to `BindCasesComponent.cshtml`, we also need actual test cases added in `BindTests.cs` in the E2E test project.

Daddoon:
Haha ok sorry Steve, pretty new to the test projects !
I will add the test case later, or tomorrow in the morning.

Thanks for your guideline.

Daddoon:
I was on other things today, and i was now looking for adding the test case.

I already know that it will fail if take the same route as the string Textbox. Also, i see that the **int** test is not even here, and it was already available as a binding.

Why ? The user can freely empty the text / numeric input and lost focus. It will throw an exception in the Browser as int.Parse / decimal.Parse (whatever) because of the null value.

As an evolution and suggestion, we should change the Binding in order to fallback to the default value if the string is empty. We can also imagine that we can set a "default" value if we are unable to parse a valid value, but this is out of context.

So can we assume this in the code ? I will speak for **int** but it will be the same for the other numeric primitive types:

`int Amount`

If Input is **Null** or **Empty** or **Invalid** should return:

`Amount = default;`

Also:

`int? Amount`

If Input is **Null** or **Empty** or **Invalid** should return:

`Amount = null;`

With this implementation, we should also change the int.Parse by int.TryParse (same for other numeric types).

Also, considering Min or Max values for numeric, should we assume that an overflow error should reset the value to default, Max (or Min) value, or to the nearest precision (for decimal numbers). Or do you have another idea about that ?

I think that managing theses scenario on the Binded inputs should allow a pretty good and safe using.

Daddoon:
Actually there is some edge case where i need some advice:

I'm actually trying to constraint the data to fallback to a default or previous value if it fail. Or actually even if i wouldn't want to do any change, the problem that if we change the received value in the handler or even just give the old existing value, or if it fail (throwing), the data binded value seem correct on other component, but the original input component does not update his value, so it may be incorrect.

Imagine that my first value is **500**, then i write **504fabc**, i would try to fallback to the previous value or even trim numerical caracters (assuming there is no custom input/output formatter available at the moment). I can notify the setter that i will set instead 504 or another valid value for this primitive type. Everything work on the binding side, except the input text / input numeric field, that keep the user interaction value.

That mean that if we don't have any other source of data binded, the end user don't know that the seen value in the input is actually not the same in memory.

- Do you have any methods or helper from BindMethod to notify the input from the internal change ? Or is it a missing check ?
- Do you plan to have some kind of formatter bindable on the input, from **string to type** (input) and **type to string** (output) ?

Thanks

SteveSandersonMS:
Thanks @Daddoon for considering designs for edge cases like unparseable values and overflows.

However for now I think it would be fine just to let the framework throw in these cases. That's because we don't really know what the developer wants to do to handle it. Depending on how their UI is presented, it might be very misleading to reinterpret some large input as zero, or to convert a nonnumeric value back to a previous value without explicitly notifying the user.

Developers will need to use suitable input restrictions to avoid errors. For example, they can use `input type=numeric`. Alternatively they can continue binding to a `string` and put their own error handling logic inside the setter of whatever they are binding to. The framework can't be sure not to mislead users otherwise.

So hopefully your existing implementation is fine, and just needs more tests.

Daddoon:
I'm actually doing some cool stuff with Blazor hosted in Xamarin and strongly typed interop support between Blazor and Xamarin.

I will come back on this pull request in few days.

SteveSandersonMS:
@Daddoon I don't mean to hassle you on this, but just FYI, if you wanted this to be included in the 0.4.0 release then we'd need the completed PR some time in the next 24 hours (roughly), as we'll be making the final builds soon. I'm happy to review and merge the PR if we do have it in time.

No problem if you prefer to leave it for 0.5.0 instead.

Daddoon:
Sorry @SteveSandersonMS  , but i was pretty busy theses last days, as i was finishing my Xamarin (with Interop) support for Blazor.

I'm at the end, but i will not be available this week exceptionaly, so i think we should report this for the 0.5.0 (or if someone else finish it before me haha).

If everything go right on my side, i will take a look at it to finish it next week.

Sorry for the delay.

SteveSandersonMS:
Completed in #926. Hope that's OK, @Daddoon!

Daddoon:
Nice ! Better that someone taken my original PR, as i was pretty busy. Thanks @Suchiman @SteveSandersonMS !

PR #633, opened 2018-04-20
Fix diffing when "bind" is combined with conditional attribute. Fixes #624

### LINKED ISSUE #624: Computed disabled attribute does not rerender.
I have tested a simple form concept.  There is a checkbox called "Active", which when you check the following fields become enabled.

```
    <div class="form-group row">
        <label class="form-check-label col-sm-4" for="Active">Is Active</label>
        <div class="col-sm-8">
            <input class="form-check-input" type="checkbox" name="Active" id="Active" bind="Active" />
        </div>
    </div>

    <div class="form-group row">
        <label class="col-form-label col-sm-4" for="Minimum">% Minimum</label>
        <div class="col-sm-8">
            <input class="form-control" style="width:160px;" type="text" name="Minimum" id="Minimum" bind="MinimumStr" disabled="@(!Active)" />
        </div>
    </div>
```

However, the rendering of being disabled only works on initial load, based on the default checked status of Active checkbox.  When you check or uncheck the Active box, it doesn't update the disabled status of the Minimum textbox, even if you hook into the Active setter and manually call StateHasChanged().

SteveSandersonMS:
The issue was that we were using incorrect logic to update the indices during the attribute diffing. The problem only manifests given a very specific diffing scenario because of a fluke in how the memory for a `RenderTreeFrame` is laid out.

Fortunately the fixed code is much simpler. And we get a minor perf boost for free.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
+++ b/src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
@@ -179,8 +179,8 @@ namespace Microsoft.AspNetCore.Blazor.RenderTree
                     // These two attributes have the same sequence and name. Keep merging.
                     AppendDiffEntriesForAttributeFrame(ref diffContext, oldStartIndex, newStartIndex);
 
-                    oldStartIndex = NextSiblingIndex(oldTree[oldStartIndex], oldStartIndex);
-                    newStartIndex = NextSiblingIndex(newTree[newStartIndex], newStartIndex);
+                    oldStartIndex++;
+                    newStartIndex++;
                     hasMoreOld = oldEndIndexExcl > oldStartIndex;
                     hasMoreNew = newEndIndexExcl > newStartIndex;
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -179,8 +179,8 @@ public static void DisposeFrames(RenderBatchBuilder batchBuilder, ArrayRange<Ren
                     // These two attributes have the same sequence and name. Keep merging.
                     AppendDiffEntriesForAttributeFrame(ref diffContext, oldStartIndex, newStartIndex);
 
-                    oldStartIndex = NextSiblingIndex(oldTree[oldStartIndex], oldStartIndex);
-                    newStartIndex = NextSiblingIndex(newTree[newStartIndex], newStartIndex);
+                    oldStartIndex++;
+                    newStartIndex++;
```
SteveSandersonMS: Should always have just been "++" here, since attribute frames are always a continuous flat list. Attributes don't have descendants to skip over.

WojcikMM:
Still Not Working.

Issue #293, opened 2018-03-19
Routing enhancements


SteveSandersonMS:
This issue is to track routing features we're likely to want to add in the future:

* Parameter constraints (e.g, `/products/{id:int}`)
* Catch-all segments (e.g., `/products/{*categoryAndName}`)
* 404 handler (it's probably not enough just to have a component that matches `{*anythingElse}`, because for SSR you'd want to know it's really not a match so you could actually return a 404)

struks93:
@SteveSandersonMS Thanks for raising this!

grahamehorner:
is a routine regex constraint on the road map?

SteveSandersonMS:
Regex constraints: definitely maybe :) We're not in a rush to put in all possible constraints because in practice it's rare that you really need them. It's often only server-side API endpoints that need to be strict about that. It's a matter of seeing what scenarios evolve and what kinds of feature requests we're getting from a wide range of people.

dlr1:
Would it be possible to define all the routes and destinations in a single place like in other JS frameworks?

Nathan-Ryan:
@dlr1 with the ability to nest routes too, perhaps similar to the way angular works with the router-outlet component.

TheAifam5:
Also saving states for every route would be nice feature to have ;)

bastienJS:
Please replace 'Multiple nested routers' with 'Unlimited nested routers' else we get the same dilemma they found out later in angular 2 that the router`s nesting skills were limited to sth. like 3 child routers :P

dlr1:
It would be nice to have some virtual methods that are called at appropriate time to either allow navigation to a page or to dis-allow navigation from a page

DNF-SaS:
Is it possible to restrict the routing from a certain slug onwards?
(e.g. everything below http://myserver/blazor will be processed, but e.g. http://myserver/otherdata won't?)

danroth27:
@DNF-SaS I believe your looking for support for catch-all segments, which hasn't been implemented yet.

conficient:
Is there anything tracking how routes behave if a user is unauthenticated vs authenticated? In MVC we can use filters on controllers to create secure routes and insecure routes.

rpedretti:
I'm building a Bing Maps component and it generates lots of anchor tags with href="#". As of version 0.5.1 whenever I click any of those links i get redirected to /. Is there a way to proper handle this?

RyoukoKonpaku:
[...]

PR #636, opened 2018-04-20
[WIP] Routing cleanup and feedback

javiercn:
Addresses feedback from https://github.com/aspnet/Blazor/pull/266:
* Consider assemblies as candidates when any of their dependencies references the blazor assembly.
* Provides an extensibility endpoint for implementing your own custom logic when subclassing the router component.
* Makes RouteTable.Routes get only.
* Removes the unnecessary page directive from the links component in the basic test app.
* Caches route tables per entry assembly
* Discovers routes with [Route] applied to a base class

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
@@ -22,37 +22,49 @@ namespace Microsoft.AspNetCore.Blazor.Components
         {
             var blazorAssembly = typeof(IComponent).Assembly;
 
-            return EnumerateAssemblies(appAssembly.GetName(), blazorAssembly, new HashSet<Assembly>(new AssemblyComparer()))
+            return EnumerateAssemblies(appAssembly.GetName(), blazorAssembly)
                 .SelectMany(a => a.ExportedTypes)
                 .Where(t => typeof(IComponent).IsAssignableFrom(t));
         }
 
         private static IEnumerable<Assembly> EnumerateAssemblies(
-            AssemblyName assemblyName,
-            Assembly blazorAssembly,
-            HashSet<Assembly> visited)
+            AssemblyName mainAssemblyName,
+            Assembly blazorAssembly)
         {
-            var assembly = Assembly.Load(assemblyName);
-            if (visited.Contains(assembly))
-            {
-                // Avoid traversing visited assemblies.
-                yield break;
-            }
-            visited.Add(assembly);
-            var references = assembly.GetReferencedAssemblies();
-            if (!references.Any(r => string.Equals(r.FullName, blazorAssembly.FullName, StringComparison.Ordinal)))
-            {
-                // Avoid traversing references that don't point to blazor (like netstandard2.0)
-                yield break;
-            }
-            else
-            {
-                yield return assembly;
+            var visited = new HashSet<Assembly>(new AssemblyComparer());
+            var candidates = new HashSet<Assembly>(new AssemblyComparer());
 
-                // Look at the list of transitive dependencies for more components.
-                foreach (var reference in references.SelectMany(r => EnumerateAssemblies(r, blazorAssembly, visited)))
+            EnumerateAssembliesCore(mainAssemblyName);
+
+            return candidates;
+
+            void EnumerateAssembliesCore(AssemblyName assemblyName)
+            {
+                var assembly = Assembly.Load(assemblyName);
+                if (visited.Contains(assembly))
                 {
-                    yield return reference;
+                    return;
+                }
+
+                visited.Add(assembly);
+                var references = assembly.GetReferencedAssemblies();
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
```diff
@@ -22,37 +22,49 @@ public static IEnumerable<Type> ResolveComponents(Assembly appAssembly)
         {
             var blazorAssembly = typeof(IComponent).Assembly;
 
-            return EnumerateAssemblies(appAssembly.GetName(), blazorAssembly, new HashSet<Assembly>(new AssemblyComparer()))
+            return EnumerateAssemblies(appAssembly.GetName(), blazorAssembly)
                 .SelectMany(a => a.ExportedTypes)
                 .Where(t => typeof(IComponent).IsAssignableFrom(t));
         }
 
         private static IEnumerable<Assembly> EnumerateAssemblies(
```
SteveSandersonMS: I'm unsure about this updated logic. Does it correctly handle a dependency graph like the following?

```
AppAssembly
 - Direct1
   - Indirect
     - Blazor
 - Direct2
   - Indirect
     - Blazor
```

If I'm following the algorithm correctly, it looks to me like the second time it sees `Indirect`, it will skip it because it's already in `visited`. Then the candidate count will not be increased, which means that it will not recognize that `Direct2` transitively references `Blazor`, so `Direct2` will not be in the output list even though it should be.

If I'm wrong about this, could you let me know what I'm misunderstanding?

If I'm right about this, maybe this approach is overcomplicated, unless we are going to add a set of test cases for it as well.

TBH I'd be satisfied with the simpler rule of "consider all transitively-referenced assemblies to be candidates, but don't recurse into any assemblies whose name starts `System.*` or is mscorlib".

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
```diff
@@ -78,11 +88,24 @@ private void Refresh()
...
-            var context = new RouteContext(locationPath);
+            var resolvedComponent = GetComponentForPath(locationPath);
+
+            _renderHandle.Render(builder => Render(builder, resolvedComponent.Handler, resolvedComponent.ComponentParameters));
+        }
+
+        /// <summary>
+        /// Decides what <see cref="IComponent"/> should handle a given <paramref name="path"/>
+        /// an the parameters associated with it.
+        /// </summary>
```
SteveSandersonMS: Typos: Decides **which** `<see cref="IComponent"/>` should handle a given <paramref name="path"/> **and** the parameters associated with it.

src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
davidfowl: Assembly scanning destroys startup time. Can we consider an approach that does most of the filtering at build time? Is this what is happening here?

SteveSandersonMS: As long as the scanning skips `mscorlib` plus any `System.*` assemblies and their dependencies (which it doesn't yet, but I'm proposing it should), then the total amount of scanning shouldn't be too bad.

I agree it *would* be better to try to resolve as much as we can at build time, but that's quite a bit of extra machinery. It's likely that we'll change how this stuff works again when we implement routing enhancements (to mount external assembly components at a certain path in the URL space) and lazy-loading. So I'd be fine with a limited amount of runtime scanning on startup until we get closer to a complete design with those other features.

javiercn: I will add some tests here just to make sure. If we want to prevent assembly scanning we can just do the same thing Razor does and put an assembly level attribute on the assembly that indicates what the components in the assembly are, but as @SteveSandersonMS said. That can wait until later.

galvesribeiro: Just to drop my 2 cents. As @davidfowl suggested, in Orleans we dropped assembly scanning and instead we adopted the concept of _Application Parts_ from ASP.Net. At startup time, the user register with an `IAppPartsManager` all the assemblies that we will need to look for a given type (in this case the router types). 

Something like this:

```
...
.ConfigureApplicationParts(parts => parts.AddApplicationPart(typeof(ValueGrain).Assembly).WithReferences())
...
```

An assembly attribute is not good as it hides from the user the real purpose of that. Being explicit is always the best option to avoid confusion IMHO.

galvesribeiro:
How this ended up? Will not continue the work on it or something else?

SteveSandersonMS:
@galvesribeiro The work here was pending while we were focused on some other areas and is no longer expected to reach the top of the queue soon. It's now so old that merging would be a challenge, and other design questions have emerged. So for clarity this PR is now closed and the items it represents are now tracked in #293. When implementing parts of #293, we can cherry-pick bits of the implementation from this PR if we want.

Issue #627, opened 2018-04-20
Packing javascript with a component


MihaMarkic:
Is scenario where I'd create a component including its javascript counterpart and pack it into NuGet library supported?

SteveSandersonMS:
Yes, please see *Build reusable component libraries* in this post: https://blogs.msdn.microsoft.com/webdev/2018/04/17/blazor-0-2-0-release-now-available/

MihaMarkic:
Hi @SteveSandersonMS,

Read that but it doesn't say if it is possible and how to include a javascript. A scenario I imagine is to create a reusable javascript-c# interop component.

SteveSandersonMS:
Sorry if it was unclear. In that blog post, the following section describes JS interop and how the component library bundles JS for this purpose. So yes, it is supported :)

floreseken:
Also, if you do a:

`dotnet new -i Microsoft.AspNetCore.Blazor.Templates`

and then:

`dotnet new blazorlib`

You will get a perfect example.

MihaMarkic:
Indeed, a good example. I'm curious though, who and when exactly invokes exampleJsInterop.js to register the methods. Do I have to place the component somewhere? Or the scripts in /content directory are automatically run?

SteveSandersonMS:
@MihaMarkic I think you should try it out and see how it works. The blog post provides steps for how to try out a library project.

dotnetnoobie:
@MihaMarkic you may also like to look at this example

MihaMarkic:
@SteveSandersonMS From what I see I think you run every .js in the component library when app starts, am I correct?

MihaMarkic:
Perhaps one more question. How does one control casing of json when calling a registered function (RegisteredFunction.Invoke<T>)?
Let's say I have a class 

```
class Tubo
{
  public string Value { get; }
}
```
I'd like to get camelCased json, like:

> { value: "xxx" }

MihaMarkic:
FYI [here](https://github.com/MihaMarkic/BlazorReduxDevToolsExtension) it is what I'm working on.

danroth27:
@MihaMarkic It's an unfortunate limitation of our JSON serialization support that the casing currently has to match. We should just change the default behavior.

danroth27:
@MihaMarkic Also, have you seen https://github.com/torhovland/blazor-redux? It looks similar to what you are attempting.

RyoukoKonpaku:
[...]

PR #644, opened 2018-04-21
SimpleJson Add support for camelCase json to .NET and vice versa

RyoukoKonpaku:
## What was added/changed
Added support for Deserializing JSON which has camel case properties which is common on many apis. This also adds support for Serializing .NET Objects to JSON that has camel case properties.

Also changed some Tests on JsonUtilTest to adhere to the changes and added some which covers the changes. Although I could not find a way to add a test that confirms that the default is ```CamelCase``` since the tests run isn't ordered.
## Note
The setting can be set to either ```None``` or ```CamelCase``` via ```JsonUtil.SetParsingStrategy``` and the default value is ```CamelCase``` as discussed #627 at the end. Though it can be changed to ```None``` if not wanted.


RyoukoKonpaku:
Thanks for the informative feedback @SteveSandersonMS. Will gladly implement those feedback once I get home.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -1434,10 +1465,22 @@ public virtual object DeserializeObject(object value, Type type)
...
+                            IDictionary<string, string> pascalCaseKeys = null;
+                            if (PropertyNaming == PropertyNamingSceme.CAMEL_CASE_PROPERTY_NAMING)
+                            {
+                                pascalCaseKeys = GetPascalCasePropertyNames(jsonObject.Keys);
+                            }
+
                             foreach (KeyValuePair<string, KeyValuePair<Type, ReflectionUtils.SetDelegate>> setter in SetCache[type])
                             {
+                                string setterKey = null;
+                                if (PropertyNaming == PropertyNamingSceme.CAMEL_CASE_PROPERTY_NAMING)
```
RyoukoKonpaku: @SteveSandersonMS I'm abit fuzzy on how the deserialize part should be done to be case sensitive again. I need abit of clarification on that part.

So far what I did is transferred the computation to the last part where the value class properties is set to the deserialized json value is done instead of blindly converting the property names to pascal case like last time. This one makes a mapping to the ```setter.key => JSON.key``` instead and gets the keys there only when the property naming is ```CamelCase```.

The mapping could have been done with ```Linq.Select``` but I thought otherwise as to follow how SimpleJson's structure was done.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -592,9 +593,16 @@ public static T DeserializeObject<T>(string json, IJsonSerializerStrategy jsonSe
 
         public static T DeserializeObject<T>(string json)
         {
+            CurrentJsonSerializerStrategy.SetPropertyNamingScheme(DefaultPropertyNamingScheme);
```
RyoukoKonpaku: Did this instead of recreating a new instance of ```PocoJsonSerializerStrategy``` class as I realized the class has a cache for future usage on the same type. So I did this to make sure that the CurrentJsonSerializationStrategy persists.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -610,6 +618,13 @@ public static string SerializeObject(object json, IJsonSerializerStrategy jsonSe
 
         public static string SerializeObject(object json)
         {
+            CurrentJsonSerializerStrategy.SetPropertyNamingScheme(DefaultPropertyNamingScheme);
+            return SerializeObject(json, CurrentJsonSerializerStrategy);
+        }
+
+        public static string SerializeObject(object json, int propertyNaming)
```
SteveSandersonMS: Can we make the `propertyNaming` type here `PropertyNaming` instead of `int`? Otherwise it's hard to make sense of what's going on.

You could move the `PropertyNaming` enum into the `SimpleJson.cs` file so it can be in the same namespace.

src/Microsoft.AspNetCore.Blazor/Json/JsonUtil.cs
```diff
@@ -13,13 +13,22 @@ namespace Microsoft.AspNetCore.Blazor
...
             => SimpleJson.SimpleJson.SerializeObject(value);
 
+        /// <summary>
+        /// Serializes the value as a JSON string with the specified property naming scheme.
+        /// </summary>
+        /// <param name="value">The value to serialize.</param>
+        /// <param name="naming">The property naming scheme for serializing.</param>
+        /// <returns>The JSON string.</returns>
+        public static string Serialize(object value, PropertyNaming naming)
+            => SimpleJson.SimpleJson.SerializeObject(value, (int)naming);
```
SteveSandersonMS: As per other comment, hopefully we can change the param type to `PropertyNaming` so there's no `(int)` cast here.

src/Microsoft.AspNetCore.Blazor/Json/JsonUtil.cs
```diff
@@ -28,5 +37,30 @@ public static string Serialize(object value)
...
+        /// <param name="json">The JSON string.</param>
+        /// <param name="propertyNaming">The property naming of the JSON string.</param>
+        /// <returns>An object of the specified type.</returns>
+        public static T Deserialize<T>(string json, PropertyNaming propertyNaming)
+            => SimpleJson.SimpleJson.DeserializeObject<T>(json, (int)propertyNaming);
+
+        /// <summary>
+        /// Property Naming scheme to be used by simple json when deserializing or serializing JSON.
+        /// </summary>
+        public enum PropertyNaming
```
SteveSandersonMS: Actually could we rename this to `JsonPropertyNaming`? Otherwise it's ambiguous whether we're referring to the JS property names or the C# ones.

SteveSandersonMS: Let's try to make it clearer still:

```csharp
public enum PropertyNameConversion
{
    /// <summary>
    /// Indicates that the property names should be left unchanged.
    /// </summary>
    None,

    /// <summary>
    /// Indicates that the JSON representation should use camel case.
    /// </summary>
    JsonCamelCase
}
```

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
SteveSandersonMS: This isn't thread safe. If two different threads are trying to use different serialization strategies, one of them will get the wrong result. I know this issue doesn't occur in browser-side code today, but it would be severe once we enable server-side rendering.

Does `SimpleJson` fundamentally rely on `CurrentJsonSerializerStrategy` being static? If so we might not be able to use this approach.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -1234,6 +1250,13 @@ interface IJsonSerializerStrategy
         [SuppressMessage("Microsoft.Design", "CA1007:UseGenericsWhereAppropriate", Justification="Need to support .NET 2")]
         bool TrySerializeNonPrimitiveObject(object input, out object output);
         object DeserializeObject(object value, Type type);
+        void SetPropertyNamingScheme(int propertyNaming);
+    }
+
+    static class PropertyNamingSceme
```
SteveSandersonMS: Once the enum is moved into this file, it should be possible to remove this class and its consts.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -1478,6 +1521,18 @@ public virtual object DeserializeObject(object value, Type type)
             return obj;
         }
 
+        internal virtual IDictionary<string, string> GetPascalCasePropertyNames(ICollection<string> keys)
+        {
+            var pascalCaseJsonPropertyKeyPair = new Dictionary<string, string>();
```
SteveSandersonMS: Doing extra allocations for this isn't ideal.

On further investigation, it turns out that Json.NET does case-insensitive matches during deserialization. If it's good enough for Json.NET, then it's probably good enough for SimpleJson. I think we can accept the limitation that deserialization picks property names case-insensitively. Will this allow you to remove these changes that allocate a dictionary?

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
RyoukoKonpaku: Not a problem. I guess I was being too stubborn on trying to follow how SimpleJson was coded xD. Will move the enum here so that we can avoid unnecessary duplication of code as per feedback.

Terricide:
I wrote some code on my version of simplejson to ignore/exclude properties that equal default(T) to shrink down the json string size on serialization. Is that something you would find useful?

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
RyoukoKonpaku: Indeed you're right this will be a problem as it can lead to undesired results on both serializing and deserializing. 

As per investigation last time I reverted instantiating a new ```PocoJsonSerializerStrategy``` class as it used it for setting a Setter/Getter cache per type for future serialization/deserialization. So far that's the main reason it was done as a static property as per what I could see. 

I could disregard that cache altogether for every call and create a new instance, but I feel it might affect performance. I have some ideas to try to preserve the cache somehow per new instance. Will investigate a bit more on this.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
SteveSandersonMS: Great. If possible please try to keep the overall code diff minimal, as we are still trying to manage this as a “fork” of SimpleJson, not an independent implementation. If the diff is too big and hard to reason about we might not be able to accept the change. Hope that’s ok.

SteveSandersonMS:
@Terricide Thanks but no, developers who want features like that should use a different JSON library.

RyoukoKonpaku:
@SteveSandersonMS Thanks for the feedback as well. Haha no worries I love feedback as I learn alot from them 😄. I'll try to do these tomorrow after work, might also add a test for that multi thread scenario once I'd figure out on how to preserve the caches.

@Terricide That sounds a nice option. As for that I think @SteveSandersonMS 's feedback on that is more appropriate as it might have some side effects that I am unaware of.

src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
RyoukoKonpaku: 👍  Understood will keep that in mind.

RyoukoKonpaku:
@SteveSandersonMS it seems CI E2E tests are failing, is this related to my changes I made? E2E + Unit Tests passed on my local PC. So I couldn't really pinpoint the problem causing this.

Overall I managed to do as little diff as possible and made it thread safe while also managed to preserve that cache implementation that SimpleJson had. Added test for thread-safety as well. 

I think this satisfies the feedbacks overall, or are there areas that needs tweaking?
I do think the E2E problem on CI needs to be resolved. I'll be checking if it's really caused by my changes later.

RyoukoKonpaku:
After some checks seems this is actually something I messed up here. Will investigate the root cause of this.

SteveSandersonMS:
@RyoukoKonpaku Thanks for your continuing work on this! I hope it's OK with you, but I'm also going to attempt to see if I can find a much shorter way of achieving the desired result, in case it's possible to avoid forking SimpleJson so significantly. My idea might turn out not to work but I want to give it a go at least 😄

RyoukoKonpaku:
@SteveSandersonMS not a problem 😄 The goal here in the end is to make it possible to have casing flexibility on deserialization/serialization. I'd love to see your solution once it's done, as it'd be a good learning exercise on my part on software/solution design overall as well xD I guess i'll leave this PR open in the meantime?

SteveSandersonMS:
@RyoukoKonpaku As discussed, I've done a PR with an alternate implementation in #746. It's much simpler because it doesn't attempt to make the casing strategy configurable or overridable.

Sorry I asked you to solve some complex problems about thread safety when it ultimately turns out we're taking a different strategy! We still appreciate your contribution. You figured out most of the challenges with this and set us on the path to constraining the scope of the problem better.

RyoukoKonpaku:
@SteveSandersonMS That's great to hear! No worries though. I learned a lot on trying to solve those problems, it was a good learning experience 😄. I guess I'll can now close this PR then? 

So by default all serialize/deserialize is camel casing as per the PR details. I'm glad to help.

SteveSandersonMS:
Thanks again. Yes this can be closed now, so I'll do that.

Issue #662, opened 2018-04-24
" rzc generate exited with code 1" when updtating packages on VS2017 Blazor template


RemiBou:
### Title
When we try to update all the packages on the VS2017 preview template for Blazor we can't build our project.

### Functional impact
Unable to create a new project with Blazor latest features as the template targets "Microsoft.AspNetCore.Blazor.Build" version 0.10

### Minimal repro steps

1. Open VS201715.7.0 preview 4*
2. New project Asp.Net core => Blazor (asp.net core hosted)
3. Right click on solution => Manage package
4. Install all the updates
5. Build solution

### Expected result
Build succeeded
### Actual result
Build failed

Output shows 
```

2>   at System.RuntimeTypeHandle.CreateInstance(RuntimeType type, Boolean publicOnly, Boolean& canBeCached, RuntimeMethodHandleInternal& ctor)
2>   at System.RuntimeType.CreateInstanceSlow(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, StackCrawlMark& stackMark)
2>   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
2>   at System.Activator.CreateInstance(Type type)
2>   at Microsoft.AspNetCore.Razor.Language.AssemblyExtension.CreateInitializer()
2>   at Microsoft.AspNetCore.Razor.Language.RazorProjectEngine.LoadExtensions(RazorProjectEngineBuilder builder, IReadOnlyList`1 extensions)
2>   at Microsoft.AspNetCore.Razor.Language.RazorProjectEngine.Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action`1 configure)
2>   at Microsoft.AspNetCore.Razor.Tools.GenerateCommand.ExecuteCore(RazorConfiguration configuration, String projectDirectory, String tagHelperManifest, List`1 sources, List`1 outputs, List`1 relativePaths)
2>   at Microsoft.AspNetCore.Razor.Tools.GenerateCommand.ExecuteCoreAsync()
2>   at Microsoft.AspNetCore.Razor.Tools.CommandBase.<ExecuteAsync>d__10.MoveNext()
2>C:\Users\remi\.nuget\packages\microsoft.aspnetcore.blazor.build\0.2.1\targets\RazorCompilation.targets(187,5): error : rzc generate exited with code 1.
2>Génération du projet "BlazorEmpty.Client.csproj" terminée -- ÉCHEC.
[...]

PR #672, opened 2018-04-25
Make Razor.Design a transitive dependency

rynowak:
By making Razor.Design a transitive dependency of Blazor.Build we can
avoid the need for users to care which version of Razor we depend upon,
and take it out of the templates.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.csproj
+++ b/src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.csproj
@@ -19,7 +19,10 @@
   <Target Name="PublishAll" BeforeTargets="_IntermediatePack">
     <PropertyGroup>
       <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
-      <NuspecProperties>version=$(PackageVersion);publishDir=$([MSBuild]::NormalizeDirectory($(IntermediatePackDir)))</NuspecProperties>
+      <NuspecProperties>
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.csproj
```diff
@@ -64,6 +67,10 @@
     <ProjectReference Include="..\Microsoft.AspNetCore.Blazor.Browser.JS\Microsoft.AspNetCore.Blazor.Browser.JS.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Blazor.Razor.Extensions\Microsoft.AspNetCore.Blazor.Razor.Extensions.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Blazor\Microsoft.AspNetCore.Blazor.csproj" />
+
+    <!-- Intentionally include Razor.Design's MSBuild assets for everyone who reference Blazor.Build -->
+    <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="$(RazorPackageVersion)" PrivateAssets="None" />
```
rynowak: This is mostly here as a safeguard to make sure we don't build with a mismatched version of Razor.

RemiBou:
This linked to Issue #662 isn't it ?

rynowak:
Yes that's right.

rynowak:
Note for #662 - this will still be a problem until we pick up 2.1.0-rc of Razor. This change https://github.com/aspnet/Razor/pull/2277 is needed.

PR #684, opened 2018-04-26
Added OnRemove and OnRemoveAsync method on BlazorComponent

Daddoon:
Added **OnRemove** and **OnRemoveAsync** method on **BlazorComponent**.

Like the opposite of OnInit and OnInitAsync, theses methods are overridable, and will be called once when disposing in ComponentState.DisposeInBatch method.

This way we can leverage some specific event based on type or data or whatever.

My actual needs were more for caching some component somewhere, like some "singleton" component, but being able to track them if any dispose occur, in order to free their reference/existence somewhere in my app for consistency and security.

I searched for an existing unit test for this case, something like OnInit, but i didn't find it.

Tested on the template project, with a nested component in a foreach loop, then leaving the page, checking that OnRemove was called.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Blazor.Rendering;
 using Microsoft.AspNetCore.Blazor.RenderTree;
 using System;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
This is meant to be covered already by `IDisposable`.

Is there some aspect of the existing `Dispose` support that's not sufficient for your requirements? If so, could you clarify what?

Daddoon:
Hello Steve,

Actually, by reading the source code, i thought that your `IDisposable `implementation was more intended for dispose some unmanaged content in the future.

Additionnaly, what about disposing Component manually ? Initially, i was going on the Disposable route,  but there were some ambiguity to me:

- Should the Dispose method being a no-op in the rendering pipeline, and actually juste firing the method and/or the overload ?
- Should the Dispose method add the current component in ComponentDisposableQueue ? Like having a public base implementation on Dispose to notify the disposing, then only allowing a protected virtual method to be overidded ?

As i don't know what is the Blazor route about component disposing per Design. Should a component being Disposable (in the DOM too) at any moment by user interaction if needed ? Or is it forbidden and we should never try to directly dispose a component in the DOM and therefore using some kind of data/rendering logic only for appending/removing component ?

If there is no conflict for the first or the second scenario, i will update the code to reflect the expected behavior.

Daddoon:
My assume about Dispose were probably overthinked so. I will go to the IDisposable route.
Thanks Steve.

Closing.

SteveSandersonMS:
Thanks @Daddoon!

PR #685, opened 2018-04-26
Support 'ref' syntax for capturing references to elements and components

SteveSandersonMS:
First, apologies that this is such a massive PR. It involves every level of the system, including tooling, compilation, component APIs, rendering/diffing, and JS interop. I'm providing the following information in the hope that it will be useful when writing docs for these features.

This PR provides two related features:

### Capturing references to elements

Example:

```
    <input ref="username" ... />

    @functions {
        ElementRef username;
    }
```

As far as .NET code is concerned, an `ElementRef` is an opaque handle. The *only* thing you can do with it is pass it through to JS code via JSON-marshalled interop. When you do so, the JS-side code receives an `HTMLElement` instance which it can use in normal DOM APIs.

For example, if you registered the following:

```js
Blazor.registerFunction('MyLib.FocusElement', function (element) {
    element.focus();
});
```

... then in .NET code you could also define an extension method:

```csharp
namespace MyLib
{
    public static class MyLibElementRefExtensions
    {
        public static void Focus(this ElementRef elementRef)
        {
            RegisteredFunction.Invoke<object>("MyLib.FocusElement", elementRef);
        }
    }
}
```

Now you can focus inputs in any of your components:

```
    @using MyLib

    <input ref="username" />
    <button onclick="@SetFocus">Set focus</button>

    @functions {
        ElementRef username;

        void SetFocus()
        {
            username.Focus();
        }
    }
```

*Important*: The `username` variable will only be populated after the component has rendered and its output includes the `<input>` element, because until then there is nothing to reference. If you try to pass an unpopulated `ElementRef` to JS code, then the JS code will receive `null`. Once we implement support for an "after render" lifecycle method, you'll be able to put code in it that lifecycle method that, for example, sets the initial focus on an element as soon as the component is first rendered.

**Use case**

The intended use cases for this feature are:

1. Integration with third-party UI libraries (such as Bootstrap) whose APIs sometimes require you to pass them references to elements for initialization
2. Calling DOM APIs that represent "commands" such as `focus` or `play` (but *not* APIs that are for constructing content, such as setting text).

It is **strongly not recommended** to use this as a way of populating the DOM routinely in cases where you could instead use Blazor's normal declarative rendering. Declarative rendering is strongly preferred because it knows how to update automatically at the correct time with minimal diffs.

### Capturing references to components

Similarly, you can use `ref` on child components:

```
    <MyLoginDialog ref="loginDialog" ... />

    @functions {
        MyLoginDialog loginDialog;

        void OnSomething()
        {
            loginDialog.Show();
        }
    }
```

As you can see, in this case the value type is not `ElementRef`, but instead is the type of the component being referenced. Again, the `loginDialog` field will only be non-`null` once the component is rendered, because until then there is no child component to reference.

*Note:* This is not a JS interop feature. It's not expected that you will pass instances of Blazor components to JS code, because the JS code would not be able to do anything meaningful with it. It's only expected that you'll use the references in .NET code.

**Use case**

The intended use case is to allow parent components to issue commands to child components such as "show" or "reset".

Even then, architecturally it's a compromise because it would be cleaner still for your child components to be stateless (that is, not acting on any state other than their parameters) and in that case it's not even theoretically possible for it to make sense to issue an "action" other than by changing their child's parameters, in which case you don't need `ref` at all.

It is **strongly not recommended** that you use `ref` as a way of mutating the state of child components. Instead, always use normal declarative parameters to pass data to child components. This will cause child components to re-render at the correct times automatically. We are moving towards changing how parameters on components are represented so that by default they are encapsulated and not possible to read/write from outside.

### Notes for PR reviewers

**Perf**

I had to introduce a new `RenderTreeFrame` type to hold a delegate that assigns the reference. This is because we can't obtain the reference during the render phase; we can only obtain it during diffing because that's when children are instantiated.

With current versions of C#, this means it will cause one allocation per `ref` per render. Later versions of C# that cache noncapturing lambda delegate instances will only require one allocation per `ref` per component (not per render).

That is, except for when it's a capturing lambda (e.g., `<div ref="someLoopVariable.SomeReference" />`. In this case there will always be more allocations, but I consider that enough of an edge case not to mind. Plus it's not like I have any reasonable alternative.

Other than that, it should be pretty neutral on perf.

**Issues**

* The `ref` tag helper doesn't show any intellisense if you prefix its value with `@` (e.g., `<div ref="@myThing" />`), even though it's fine without (e.g., `<div ref="myThing" />`). I have not been able to track down why this difference exists. The source map data *looks* correct. Would appreciate any insight from @rynowak on this! It does work correctly in runtime builds with or without `@`.
* The `ref` tag helper does show correct warning squigglies if you reference a nonexistent variable (e.g., `<div ref="@somethingThatIsNotDefined" />`). But it does *not* show squigglies if you reference something of the wrong type (e.g., `<div ref="@someIntField" />`). I know why, which is because in the generated code we can only associate the source token with either the assigned variable **or** the value we're assigning to it. In the case of a type mismatch, the error is associated with the value being assigned. I consider it more important to catch the "nonexistent variable" case. In both cases the runtime compiler output is correct and makes it easy to diagnose the error. I didn't find any trick to produce alternate design-time code that catches both error types using a single token.


SteveSandersonMS:
Yes (except the syntax is `<AppCountdownTimer ref="timer"...`).

Kaffeegangster:
This is an feature i have been waiting for :-) Does it make sense to use an HTMLElement from https://github.com/kjpou1/wasm-dom/blob/master/Mono.WebAssembly.Browser/DOM/HTMLElement.cs instead of ElementRef? I know the goal is to stay in declarative world but HTMLElement would make interop or translating legacy javascript code even nicer. I would alose love to see https://github.com/mono/mono/pull/8442 merged into Mono/Blazor. Very cool work to improve inteop.

SteveSandersonMS:
We may add an easy way to convert an ElementRef to a HTMLElement but it’s not part of the immediate scope.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RefLoweringPass.cs
```diff
@@ -0,0 +1,83 @@
...
+                else if (attributeNode.Children[0] is CSharpExpressionIntermediateNode csharpNode)
+                {
+                    if (csharpNode.Children.Count > 0)
+                    {
+                        foundToken = csharpNode.Children[0] as IntermediateToken;
+                    }
+                }
+            }
+            
+            return !string.IsNullOrWhiteSpace(foundToken?.Content) ? foundToken : null;
```
rynowak: Not sure to what degree we care, but an identifier *could* appear as multiple tokens if it's separated by a comment.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RefLoweringPass.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (identifierToken != null)
+            {
+                // Determine whether this is an element capture or a component capture, and
+                // if applicable the type name that will appear in the resulting capture code
+                var componentTagHelper = node.TagHelpers.FirstOrDefault(x => x.IsComponentTagHelper());
+                var refExtensionNode = componentTagHelper == null
+                    ? new RefExtensionNode(identifierToken)
+                    : new RefExtensionNode(identifierToken, componentTagHelper.GetTypeName());
+
+                node.Children.Add(refExtensionNode);
```
rynowak: I'm glad that this worked about to be relatively straightforward. This looks good 😀

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RefLoweringPass.cs
```diff
@@ -0,0 +1,83 @@
...
+                        attributeNode.TagHelper != null &&
+                        attributeNode.TagHelper.IsRefTagHelper())
+                    {
+                        RewriteUsage(@class, node, j, attributeNode, isDesignTime);
+                    }
+                }
+            }
+        }
+
+        private void RewriteUsage(ClassDeclarationIntermediateNode classNode, TagHelperIntermediateNode node, int index, ComponentAttributeExtensionNode attributeNode, bool isDesignTime)
```
rynowak: Looks like `isDesignTime` is unused

SteveSandersonMS: Ah, well spotted! Missed that during some updates.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/DesignTimeCodeGenerationTest/Component_WithRef_CSharpReservedWord/TestComponent.codegen.cs
```diff
@@ -0,0 +1,47 @@
...
+        #pragma warning restore 219
+        #pragma warning disable 0414
+        private static System.Object __o = null;
+        #pragma warning restore 0414
+        #pragma warning disable 1998
+        protected override void BuildRenderTree(Microsoft.AspNetCore.Blazor.RenderTree.RenderTreeBuilder builder)
+        {
+            base.BuildRenderTree(builder);
+#line 2 "x:\dir\subdir\Test\TestComponent.cshtml"
+                                         @class = default(Test.MyComponent);
```
rynowak: > The ref tag helper does show correct warning squigglies if you reference a nonexistent variable (e.g., <div ref="somethingThatIsNotDefined" />). But it does not show squigglies if you reference something of the wrong type (e.g., <div ref="someIntField" />). 

I see now what you mean about the errors doing the wrong thing. If you want to address that problem, I'd suggest generating something like.

```
BindMethods.AssignComponent<Test.MyComponent>(
#line .....
              @class
#line default
#line hidden
);
@class = default;
```

Where you've defined:
```
public void BindMethods.AssignComponent<T>(T value)
```

This would help in a few ways:
1. It provides a context where the type of the user-provided expression has to be known
2. It makes sure the value gets assigned at some point, preventing a warning.

How this could break down is if the expression provided is not an LValue. We already have this problem with bind, and I'm not sure there's a 100% perfect solution without more love from C#.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RefLoweringPass.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            node.Children.Remove(attributeNode);
+
+            // If we can't get a nonempty attribute name, do nothing because there will
+            // already be a diagnostic for empty values
+            var identifierToken = DetermineIdentifierToken(attributeNode);
+            if (identifierToken != null)
```
rynowak: This is likely why you're not getting intellisense in some cases. What will happen is that the region where you're typing has **disappeared** from the generated C# until you type something non-empty. My guess is that partial parsing is behaving differently between a tag helper attribute and an implicit expression inside a tag helper attribute (that's with `@`).

Basically, you need to provide a 'target' for the generated code even if what's there now isn't valid because it's empty. An easy way to do this would be move the 'remove' into this if.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RefLoweringPass.cs
SteveSandersonMS: Good point. For simplicity, I'll amend the code so that it only accepts the scenario where there is exactly one child, so that developers will find it just doesn't let them do that. Otherwise we'll have to get into the complexity we have around "mixed content" elsewhere.

test/Microsoft.AspNetCore.Blazor.Build.Test/DesignTimeCodeGenerationTest.cs
```diff
@@ -607,5 +607,164 @@ public void BindToElementFallback_WithFormat_WritesAttributes()
...
+");
+
+            // Assert
+            AssertDocumentNodeMatchesBaseline(generated.CodeDocument);
+            AssertCSharpDocumentMatchesBaseline(generated.CodeDocument);
+            CompileToAssembly(generated);
+        }
+
+        [Fact]
+        public void Component_WithRef_CSharpReservedWord()
```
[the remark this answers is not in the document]
SteveSandersonMS: OK, I'll remove the redundant tests to avoid spamming ourselves.

test/Microsoft.AspNetCore.Blazor.Build.Test/DesignTimeCodeGenerationTest.cs
```diff
@@ -607,5 +607,164 @@ public void BindToElementFallback_WithFormat_WritesAttributes()
...
+            {
+            // Arrange
+            AdditionalSyntaxTrees.Add(CSharpSyntaxTree.ParseText(@"
```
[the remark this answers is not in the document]
SteveSandersonMS: Thanks - updated.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/RefLoweringPass.cs
SteveSandersonMS: You're right! That makes sense and does fix it :)

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/DesignTimeCodeGenerationTest/Component_WithRef_CSharpReservedWord/TestComponent.codegen.cs
SteveSandersonMS: I did try something along these lines yesterday but it led to the allowed conversions being the wrong way round:

    SomeMethod(MyComponent instance)

... allows you to pass anything that implicitly converts *to* a `MyComponent`, but I need an expression that allows you to pass anything that implicitly *from* `MyComponent`.

I also experimented with `SomeMethod(out MyComponent instance)`, but that doesn't work because `out` (like `ref`) only matches exact types, not things that are implicitly convertible.

AFAICT the only construction that has the correct behavior is an assignment (`yourIdentifier = (MyComponent)anything;`), which is what comes with the limitation that it can only catch one of the two types of errors.

If I'm missing something obvious please let me know!

rynowak: That makes sense. There are limitations here that we can't really avoid.

BTW WPF solves this problem by not letting you write the field declaration - which means that it's always a field and always private.

dlr1:
Multiple ref's with same name and type should throw an error,

DNF-SaS:
What's the proposed way for a ref'ed component to tell it's parent component that some state-property has changed? StateHasChanged() in the child seems not to bubble up.

SteveSandersonMS:
It can expose a public event or can take a callback as a parameter.

sven5:
In the docs you say:

> The intended use case is to allow parent components to issue commands to child components such as "show" or "reset".

Is this functionality already implemented in 0.5.1 ? I cannot find the mentioned Show() method.

Thanks
Sven

sven5:
Ok thank you. The wording appeared to me as if it were built into the Framework.

PR #691, opened 2018-04-26
OnAfterRender / OnAfterRenderAsync

SteveSandersonMS:
Gives components notification when rendering has occurred. This makes it possible to perform additional initialization, such as activating third-party JS libraries that want to operate on the rendered DOM elements.

I've chained this PR onto #685 because I wanted the E2E test to cover the combination of these two features, as it's the most interesting and important use case.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -117,14 +117,14 @@ namespace Microsoft.AspNetCore.Blazor.Components
             => true;
 
         /// <summary>
-        /// Method invoked each time after the component has been rendered.
+        /// Method invoked after each time the component has been rendered.
         /// </summary>
         protected virtual void OnAfterRender()
         {
         }
 
         /// <summary>
-        /// Method invoked each time after the component has been rendered. Note that the component does
+        /// Method invoked after each time the component has been rendered. Note that the component does
         /// not automatically re-render after the completion of any returned <see cref="Task"/>, because
         /// that would cause an infinite render loop.
         /// </summary>
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNetCore.Blazor.Components
     /// Optional base class for Blazor components. Alternatively, Blazor components may
     /// implement <see cref="IComponent"/> directly.
     /// </summary>
-    public abstract class BlazorComponent : IComponent, IHandleEvent
+    public abstract class BlazorComponent : IComponent, IHandleEvent, IHandleAfterRender
     {
         /// <summary>
         /// Specifies the name of the <see cref="RenderTree"/>-building method.
@@ -87,6 +87,7 @@ namespace Microsoft.AspNetCore.Blazor.Components
         /// Method invoked when the component has received parameters from its parent in
         /// the render tree, and the incoming values have been assigned to properties.
         /// </summary>
+        /// <returns>A <see cref="Task"/> representing any asynchronous operation, or <see langword="null"/>.</returns>
         protected virtual Task OnParametersSetAsync()
             => null;
 
@@ -114,6 +115,22 @@ namespace Microsoft.AspNetCore.Blazor.Components
         /// <returns></returns>
         protected virtual bool ShouldRender()
             => true;
+
+        /// <summary>
+        /// Method invoked each time after the component has been rendered.
+        /// </summary>
+        protected virtual void OnAfterRender()
+        {
+        }
+
+        /// <summary>
+        /// Method invoked each time after the component has been rendered. Note that the component does
+        /// not automatically re-render after the completion of any returned <see cref="Task"/>, because
+        /// that would cause an infinite render loop.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing any asynchronous operation, or <see langword="null"/>.</returns>
+        protected virtual Task OnAfterRenderAsync()
+            => null;
 
         void IComponent.Init(RenderHandle renderHandle)
         {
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -184,5 +201,24 @@ void IHandleEvent.HandleEvent(UIEventHandler handler, UIEventArgs args)
             // at the end of every event callback.
             StateHasChanged();
         }
+
+        void IHandleAfterRender.OnAfterRender()
```
SteveSandersonMS: We're lacking unit test coverage for the `BlazorComponent` base class presently. Initially that was deliberate because the set of lifecycle methods was constantly in flux, and my goal was to keep the `BlazorComponent` small and simple enough to reason about trivially.

But we're now reaching the end of that phase I think. We should do some kind of design review of the lifecycle methods that now exist (for naming, exact semantics, etc.) and then nail it all down with tests.

The E2E coverage of it is reasonable already though.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -115,6 +116,22 @@ protected void StateHasChanged()
...
+        }
+
+        /// <summary>
+        /// Method invoked each time after the component has been rendered. Note that the component does
+        /// not automatically re-render after the completion of any returned <see cref="Task"/>, because
+        /// that would cause an infinite render loop.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing any asynchronous operation, or <see langword="null"/>.</returns>
+        protected virtual Task OnAfterRenderAsync()
+            => null;
```
javiercn: Shouldn't this be Task.CompletedTask?

src/Microsoft.AspNetCore.Blazor/Components/IHandleAfterRender.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Interface implemented by components that receive notification that they have been rendered.
+    /// </summary>
+    public interface IHandleAfterRender
+    {
+        /// <summary>
+        /// Notifies the component that it has been rendered.
+        /// </summary>
+        void OnAfterRender();
```
javiercn: Why did you choose to make the interface synchronous and then have BlazorComponent have an asynchronous overload?

Feels to me more natural to have this interface be asynchronous and just reverse what you do in the implementation.
```csharp
public virtual Task OnAfterRenderAsync()
{
    OnAfterRender();

    return Task.CompletedTask;
}

protected virtual void OnAfterRender() 
{
}
```
We do a similar pattern in filters in MVC. See
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Filters/ActionFilterAttribute.cs

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
```diff
@@ -115,6 +116,22 @@ protected void StateHasChanged()
...
+        /// Method invoked each time after the component has been rendered.
+        /// </summary>
+        protected virtual void OnAfterRender()
+        {
+        }
+
+        /// <summary>
+        /// Method invoked each time after the component has been rendered. Note that the component does
+        /// not automatically re-render after the completion of any returned <see cref="Task"/>, because
+        /// that would cause an infinite render loop.
```
rynowak: If this is the case, is there a good reason to provide an async version of this? error handling?

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
rynowak: I'm on board with doing a conceptual review once the features are in. I think what's here follows the same pattern as the other stuff we have and that's good enough for right now.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
rynowak: Sounds like you answered that already further down :)

test/testapps/BasicTestApp/AfterRenderInteropComponent.cshtml
```diff
@@ -0,0 +1,13 @@
...
+
+<input ref="myInput" value="Value set during render" />
+
+@functions {
+    ElementRef myInput;
+
+    protected override void OnAfterRender()
+    {
+        RegisteredFunction.Invoke<object>("setElementValue", myInput, "Value set after render");
+    }
```
rynowak: It's really nice to see this featuring reaching completeness.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
SteveSandersonMS: Great. One other thing to include in a design review about this (and other lifecycle methods) is failure handling. Currently if you have an async exception we deal with it, but synchronous ones kill the entire render batch. That’s probably not what we want!

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
SteveSandersonMS: I agree that would be better. There are other equivalent lifecycle methods whose base implementations also return `null`, so in this PR I'll keep this as-is for consistency, and we can review all of them together.

src/Microsoft.AspNetCore.Blazor/Components/IHandleAfterRender.cs
SteveSandersonMS: That's a good point, and again I agree that would be better. Also again I'll keep it consistent with the rest of the lifecycle methods for this PR and we can do a broader review of them all later.

src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
DNF-SaS: @javiercn Thanks, I just wanted to suggest exactly that. :-)

PR #692, opened 2018-04-27
Fixup project templates to handle project names with dashes

danroth27:
This template update is a partial fix for when the project name has a dash in it. 

It's  only a partial fix because it doesn't address sanitizing the namespace for the generated component classes. @rynowak Is this something you could help me out with?


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentClassifierPass.cs
```diff
@@ -141,9 +143,20 @@ private bool TryComputeNamespaceAndClass(string filePath, string relativePath, o
...
 
             var builder = new StringBuilder();
-            builder.Append(baseNamespace); // Don't sanitize, we expect it to contain dots.
 
-            var segments = relativePath.Split(PathSeparators, StringSplitOptions.RemoveEmptyEntries);
+            // Sanitize the base namespace, but leave the dots.
+            var segments = baseNamespace.Split(NamespaceSeparators, StringSplitOptions.RemoveEmptyEntries);
+            builder.Append(CSharpIdentifier.SanitizeClassName(segments[0]));
+            if (segments.Length > 1)
+            {
```
rynowak: If is not necessary. The loop will noop if the length == 0.

danroth27:
:up::date:

test/Microsoft.AspNetCore.Blazor.Build.Test/WorkingDirectoryRazorIntegrationTest.cs
```diff
@@ -0,0 +1,41 @@
...
+            // Act
+            var result = CompileToAssembly(relativePath, "");
+
+            // Assert
+            Assert.Empty(result.Diagnostics);
+
+            var type = Assert.Single(result.Assembly.GetTypes());
+            Assert.Equal(expectedNamespace, type.Namespace);
+            Assert.Equal(expectedClassName, type.Name);
+        }
```
rynowak: Nice!

PR #707, opened 2018-04-29
Rewrite of HTML handling for Blazor

rynowak:
This change replaces the parsing of HTML that we perform during the code
generation phase, which parsing of HTML during the IR lowering phase.
The main benefit of this change is that the structure of the HTML is
reflected in the IR tree, allowing us to do more more advance
transformations.

As an example, see how the the handling of `<script>` tags is now a
separate pass.

As an aside from this I also redesigned the structure of component IR
nodes to match the new HTML element nodes. Passes are now more easily
aware of the nodes they are expected to handle and are more easily aware
of the difference between a component and element. This still isn't as
clean as I would like, but I think it's a reasonable improvement.

Another benefit of this is that the code generation is much simpler and
requires much less bookkeeping and statefulness.


rynowak:
@SteveSandersonMS - realize this is a huge PR and I can understand if you don't want to look through every detail 👍 It would be good to get your eyeballs on the new component and html element nodes, as well as the changes I made to the 'ref' pass.

Note that with one small exception there should be no **codgen** changes in this PR. That gives me a good deal of confidence. Most of the IR changes will be no surprise.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
```diff
@@ -30,23 +30,14 @@ public static RazorDiagnostic Create_CodeBlockInAttribute(SourceSpan? source, st
...
 
-        public static readonly RazorDiagnosticDescriptor ExpressionInAttributeList =
-            new RazorDiagnosticDescriptor(
+        public static readonly RazorDiagnosticDescriptor UnclosedTag = new RazorDiagnosticDescriptor(
             "BL9980",
-            () =>
-                "Expressions like '{0}' inside of a tag must be part of an attribute. " +
-                "Previous releases of Blazor supported constructs like '@onclick(...)' or '@bind(...)'." +
-                "These features have been changed to use attribute syntax. " +
-                "Use 'onclick=\"@...\"' or 'bind=\"...\" respectively.",
```
rynowak: I removed this error. We had discussed removing it in 0.3 previously, and it would have required some work to retain.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
```diff
@@ -72,12 +72,13 @@ public static void Register(RazorProjectEngineBuilder builder)
...
+            builder.Features.Add(new ComponentDocumentRewritePass());
+            builder.Features.Add(new ScriptTagPass());
             builder.Features.Add(new ComplexAttributeContentPass());
-            builder.Features.Add(new BindLoweringPass());
-            builder.Features.Add(new EventHandlerLoweringPass());
             builder.Features.Add(new ComponentLoweringPass());
+            builder.Features.Add(new EventHandlerLoweringPass());
             builder.Features.Add(new RefLoweringPass());
-            builder.Features.Add(new OrphanTagHelperLoweringPass());
+            builder.Features.Add(new BindLoweringPass());
```
rynowak: These passes did not match the stated ordering. This has been cleaned up. 

Now the component pass runs pretty early, and makes the decision of whether the node is an element or component. It leaves behind 'tag helper property' nodes for the other 3 passes to handle. The expectation is that each 'tag helper property' must belong to either of these 4 later passes, and should never make it to the code generation phase without being replaced.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
```diff
@@ -522,57 +400,42 @@ public override void WriteComponentAttribute(CodeRenderingContext context, Compo
...
-            else
+            var methodName = node.IsComponentCapture
+                ? nameof(BlazorApi.RenderTreeBuilder.AddComponentReferenceCapture)
+                : nameof(BlazorApi.RenderTreeBuilder.AddElementReferenceCapture);
+            codeWriter
+                .WriteStartMethodInvocation($"{_scopeStack.BuilderVarName}.{methodName}")
+                .Write((_sourceSequence++).ToString())
+                .WriteParameterSeparator();
+
+            const string refCaptureParamName = "__value";
```
rynowak: The new approach that I took for the HTML element and component nodes just builds the correct ordering into the design. This way features like attributes and ref can't be broken by putting the nodes in the wrong order.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
```diff
@@ -24,112 +24,265 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
...
+
+            public override void VisitTagHelperBody(TagHelperBodyIntermediateNode node)
+            {
+                for (var i = 0; i < node.Children.Count; i++)
+                {
+                    _children.Add(node.Children[i]);
+                }
+            }
+
+            public override void VisitTagHelperHtmlAttribute(TagHelperHtmlAttributeIntermediateNode node)
```
rynowak: I'm still not happy with how complicated this is, but I don't really want to spend more time on it unless there's something to be gained.

rynowak:
I'm also OK with deferring this until after 0.3 ships since we don't need it for any experiences we are trying to build for the current release.

SteveSandersonMS:
Wow, yes, this is huge but it’s fantastic that the resulting codegen output is identical. 

I will review soon. As per your comment above, I think it would be good if we merged this after branching for 0.3.0 because of it not changing any behaviour and it being advantageous to have max time to discover any possible unintended changes. That ok?

rynowak:
Merging after an in person discussion 👍

PR #708, opened 2018-04-30
Resolves #297

### LINKED ISSUE #297: Support data-*=@expression properly and remove the data_* workaround
This requires changes in Razor.

Assigning to @rynowak, but it's not urgent and I know it won't be done for 0.1.0.

rynowak:
Removes the workaround for #297 and sets the langauge version to
experimental.

This will require 2.1.0-rc of Razor.


rynowak:
I expect this to fail for now. I just wanted to post the results somewhere so we're ready to merge when we update to RC.

SteveSandersonMS:
@rynowak Do you anticipate we can update `<RazorPackageVersion>` for our 0.4.0 release?

rynowak:
yes, we can at least do rc. Depending on timing we might be able to do RTM as well.

danroth27:
We should plan on targeting RTM.

PR #709, opened 2018-04-30
Add [Parameter] for component parameters

rynowak:
This change introduces ParameterAttribute to specify a bindable
component parameter. As of the 0.3 release of Blazor we plan to make
[Parameter] required to make a property bindable by callers.

This also applies to parameters when their value is set by the
infrastructure, such as `Body` for layouts, and route paramters.

The rationale behind this change is that we think there is a need to
separate the definition of properties from their suitability for a
caller to set them through markup. We plan to introduce more features in
this area in the future such as marking parameters as required. This is
first step, and we think that this approach will scale nicely as we add
more functionaly.

The 0.3 release seems like the right time to change this behavior since
we're also introducing `ref` for captures in this release.

### DIFF
--- a/samples/StandaloneApp/Shared/MainLayout.cshtml
+++ b/samples/StandaloneApp/Shared/MainLayout.cshtml
@@ -15,5 +15,6 @@
 </div>
 
 @functions {
+    [Parameter]
     public RenderFragment Body { get; set; }
 }
--- a/test/Microsoft.AspNetCore.Blazor.Build.Test/BindRazorIntegrationTest.cs
+++ b/test/Microsoft.AspNetCore.Blazor.Build.Test/BindRazorIntegrationTest.cs
@@ -24,8 +24,10 @@ namespace Test
 {
     public class MyComponent : BlazorComponent
     {
+        [Parameter]
         public int Value { get; set; }
 
+        [Parameter]
         public Action<int> ValueChanged { get; set; }
     }
 }"));
@@ -96,8 +98,10 @@ namespace Test
 {
     public class MyComponent : BlazorComponent
     {
+        [Parameter]
         public int Value { get; set; }
 
+        [Parameter]
         public Action<int> OnChanged { get; set; }
     }
 }"));
--- a/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorApi.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorApi.cs
@@ -16,6 +16,11 @@ namespace Microsoft.AspNetCore.Blazor.Razor
             public static readonly string BuildRenderTree = nameof(BuildRenderTree);
         }
 
+        public static class ParameterAttribute
+        {
+            public static readonly string FullTypeName = "Microsoft.AspNetCore.Blazor.Components.ParameterAttribute";
+        }
+
         public static class LayoutAttribute
         {
             public static readonly string FullTypeName = "Microsoft.AspNetCore.Blazor.Layouts.LayoutAttribute";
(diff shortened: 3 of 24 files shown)


samples/StandaloneApp/Shared/MainLayout.cshtml
```diff
@@ -15,5 +15,6 @@
 </div>
 
 @functions {
+    [Parameter]
     public RenderFragment Body { get; set; }
```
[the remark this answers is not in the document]
SteveSandersonMS: I agree a base class would be the easy solution here. 

But also I recognise the new design prevents us from usefully using interfaces to indicate component contracts more generally, which is not something I’d really considered before. Still I think the advantages outweigh the disadvantages.

test/Microsoft.AspNetCore.Blazor.Build.Test/BindRazorIntegrationTest.cs
```diff
@@ -18,14 +18,17 @@ public void Render_BindToComponent_SpecifiesValue_WithMatchingProperties()
...
             AdditionalSyntaxTrees.Add(Parse(@"
 using System;
+using Microsoft.AspNetCore.Blazor;
 using Microsoft.AspNetCore.Blazor.Components;
 
 namespace Test
 {
     public class MyComponent : BlazorComponent
     {
+        [Parameter]
```
SteveSandersonMS: What do you think about maybe putting the new attribute into the .Components namespace? It’s only usable in the context where you are also using IComponent or BlazorComponent. Also would avoid all these extra usings.

SteveSandersonMS: I don’t mind hugely BTW.

rynowak: Yeah I think this is fine. We might want to consider consolidating all the things we expect 99% of users to use into a single namespace but your suggesting is in line with the current design so let's do that.

schotime:
Would be great to have some more rationale behind this. Why was having public properties for those being called an issue? Could've then used the required attribute for required ones by the calling component. Thanks @rynowak

SteveSandersonMS:
Sure, happy to explain. 

The reason is for encapsulation. You wouldn’t want C# to force all your properties to be public read/write, but that’s equivalent to what the previous Blazor parameters design required. It’s bad because external code might mutate your state in unexpected ways at any time, and you can’t reason about or control it. Encapsulation is a basic OO principle. 

So people should default to making parameter-backing properties private. But then which of the properties should be treated as parameters? That’s why there’s an attribute to signal it. The attribute also gives a natural place to put other options such as ‘required’.

MisinformedDNA:
I get that you want to protect against bad state, but I disagree that this is the approach to take. 

First off, `private` is being used incorrectly and contrary to its [definition](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/private): 

> Private access is the least permissive access level. Private members are accessible only within the body of the class or the struct in which they are declared

It's also counter-intuitive. Especially given that we've added custom attributes in XAML and TagHelper and we've always used public there as well.

Finally, I don't see how you are reaching your desired goal anyway. Parameters obviously accept variables, variables that can change, so the component still has no control over what is sent to it.

Components that need to account for state, should do so with logic in the property setter and the use of a private backing field.

SteveSandersonMS:
@MisinformedDNA Thanks for the additional perspectives. If C# had a notion of a `framework` access modifier (values supplied by some kind of "framework" and not user-settable) then we'd use that, but it doesn't, so we're having to best model the desired constraints within the available model.

I'm not going to try to counter-argue each of your points, but rather I'll accept your perspective and we'll bear that in mind as we continue reviewing this design.

vertonghenb:
@MisinformedDNA I do agree, it's a property which can be set by a parent component, state logic should be handled in the setter of the Property. Apart from breaking the idea of a private member (which be definition it isn't, it's a parameter), it felt more intuitive to use a property.

Although I agree with @SteveSandersonMS and I think this is the main reason why, not sure though.
> which of the properties should be treated as parameters?

But what if we have Properties as always being declared as Parameters and Privates members used as for what they really are, you can't interact with them outside the class => State?

When working on the Required or Optional we can re-use the same idea of the [Required] data attribute. Which would really feel like making a ViewModel thus being very intuitive.

PR #713, opened 2018-04-30
Encourage encapsulation of component parameter properties

SteveSandersonMS:
* Implemented the caching-for-perf inside `ParameterCollection` that I've been meaning to get to for a while (i.e., cache fully reified delegates, not `PropertyInfo` or anything like that)
* More test coverage for `ParameterCollection` now its behavior is a bit more sophisticated
* Enable binding to nonpublic properties
* Roslyn analyzer to warn about `[Parameter] public SomeProperty ...`, including codefix that removes the `public` (not that this currently works in `.cshtml` files, but hopefully will in the future - for now VS only shows the codefixes in `.cs` files)
* Change the tag helper descriptor so it also accepts nonpublic `[Parameter]`

Still to do:

* Change templates/samples to encapsulate the parameters

Asking for review now so hopefully I can merge tomorrow morning once the final bit is done.


src/Microsoft.AspNetCore.Blazor.Build.Analyzers/ComponentParametersShouldNotBePublicAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        private void AnalyzeSyntax(SyntaxNodeAnalysisContext context)
+        {
+            var semanticModel = context.SemanticModel;
+            var declaration = (PropertyDeclarationSyntax)context.Node;
+
+            var parameterAttribute = declaration.AttributeLists
+                .SelectMany(list => list.Attributes)
+                .Where(attr => semanticModel.GetTypeInfo(attr).Type?.ToDisplayString() == "Microsoft.AspNetCore.Blazor.ParameterAttribute")
```
rynowak: Consider including `BlazorApi` as a shared source file and using the named constant.

src/Microsoft.AspNetCore.Blazor.Build.Analyzers/ComponentParametersShouldNotBePublicCodeFixProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Blazor.Build.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(ComponentParametersShouldNotBePublicCodeFixProvider)), Shared]
+    public class ComponentParametersShouldNotBePublicCodeFixProvider : CodeFixProvider
```
rynowak: As you noticed, I'm sure, this won't actually work in .cshtml. It won't be fixed any time soon but we're making progress.

src/Microsoft.AspNetCore.Blazor.Build.Analyzers/Microsoft.AspNetCore.Blazor.Build.Analyzers.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
rynowak: Do we want to call this `Microsoft.AspNetCore.Blazor.Analyzers` instead? Do you think we'd ever have more than one package of analyzers?

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
```diff
@@ -26,47 +38,89 @@ public static class ParameterCollectionExtensions
...
-                    $"Unable to set property '{parameter.Name}' on object of " +
-                    $"type '{target.GetType().FullName}'. The error was: {ex.Message}", ex);
+                var propertySetter = MemberAssignment.CreatePropertySetter(targetType, propertyInfo);
+
+                var propertyName = propertyInfo.Name;
+                if (result.ContainsKey(propertyName))
+                {
+                    throw new InvalidOperationException(
+                        $"The type '{targetType.FullName}' declares more than one parameter matching the " +
+                        $"name '{propertyName.ToLowerInvariant()}'. Parameter names are case-insensitive and must be unique.");
```
rynowak: I think the other case you could hit here is property visibility. Using `new` to hide a property will trigger this error I think because `MemberAssignment.GetBindableProperties` doesn't account for that.

test/Microsoft.AspNetCore.Blazor.Build.Analyzers.Test/Microsoft.AspNetCore.Blazor.Build.Analyzers.Test.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Remove="iomjthbk.12p~" />
```
rynowak: Any idea what this is?

SteveSandersonMS: Ugh, sorry, will remove. VS adds temporary files with names like this during some operations like renames. I guess it's meant to clean up after itself but didn't do so correctly this time.

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
SteveSandersonMS: Yes, that's deliberately the case. We treat it as an error to shadow a property that has `[Parameter]`, because if you do that there's no way for anyone to set the shadowed parameter, and yet the inherited class can legitimately require it to be set. There's a unit test for that scenario too.

src/Microsoft.AspNetCore.Blazor.Build.Analyzers/Microsoft.AspNetCore.Blazor.Build.Analyzers.csproj
SteveSandersonMS: Good point. I'll rename.

src/Microsoft.AspNetCore.Blazor.Build.Analyzers/ComponentParametersShouldNotBePublicCodeFixProvider.cs
SteveSandersonMS: Yes. I implemented it anyway because it's cool and quite easy and hopefully one day will work.

SteveSandersonMS:
@rynowak This is now merged including the `WithMetadataImportOptions` technique for enabling visibility of private `[Parameter]` members in external assemblies. I did check this works both during builds and visibly in the VS editor. Let me know if you can anticipate anything problematic about that.

While stepping through it in the debugger, I noticed that now we're searching private members, the considerable majority of the scanning operations take place on the `BlazorComponent` base class because it has a lot more members than a typical derived component class. Since we don't currently have any `[Parameter]` members on the base class (and of course not on the `System.Object` ancestor class), I've tweaked the logic so it bails out as soon as it reaches `BlazorComponent`: https://github.com/aspnet/Blazor/pull/713/commits/2009e434c886532e3308a55a9a0e29d0d2415bbc#diff-f981900570ad8f305eb7d490851799ecR178. We'd have to change this if one day we do choose to put `[Parameter]` on a `BlazorComponent` property, but there's a good chance we never will. Again, please shout at me if you think this is undesirable for any reason. Hopefully this change partly makes up for all the extra members we have to step through now we include private/internal ones.

rynowak:
I think that sounds fine

PR #718, opened 2018-05-01
WIP Partial for fix #546 - Add more events

rynowak:
Work in progress adding more built-in event types


src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRendererEventDispatcher.cs
```diff
@@ -38,8 +38,10 @@ private static UIEventArgs ParseEventArgsJson(string eventArgsType, string event
                     return JsonUtil.Deserialize<UIKeyboardEventArgs>(eventArgsJson);
                 case "change":
                     return JsonUtil.Deserialize<UIChangeEventArgs>(eventArgsJson);
+                case "focus":
+                    return JsonUtil.Deserialize<UIFocusEventArgs>(eventArgsJson);
                 default:
-                    throw new ArgumentException($"Unsupported value '{eventArgsType}'.", nameof(eventArgsType));
+                    return JsonUtil.Deserialize<UIEventArgs>(eventArgsJson);
```
rynowak: Changing this to allow 'unknown' to flow through makes it possible to at least pipe through any event without data.

We might want to consider something in the future that makes this abitrarily extensible by mapping 'unknown' to something with a dictionary or JSON blob. You can imagine that components might use their own bubbling events for notifications - this pattern was used in WPF as well

rynowak: OR alternatively, we could allow you to do strongly typed things by subclassing UIEventArgs

test/Microsoft.AspNetCore.Blazor.E2ETest/Infrastructure/BrowserFixture.cs
```diff
@@ -13,6 +13,10 @@ public class BrowserFixture : IDisposable
     {
         public IWebDriver Browser { get; }
 
+        public IKeyboard Keyboard => (Browser as IHasInputDevices)?.Keyboard;
+
+        public IMouse Mouse => (Browser as IHasInputDevices)?.Mouse;
+
```
rynowak: Will revert this, I ended up not needing it.

test/Microsoft.AspNetCore.Blazor.E2ETest/Infrastructure/BrowserTestBase.cs
```diff
@@ -11,20 +11,22 @@ namespace Microsoft.AspNetCore.Blazor.E2ETest.Infrastructure
...
+
+        public static ILogs Logs => _fixture.Value.Logs;
 
         public static ITestOutputHelper Output => _output.Value;
 
         public BrowserTestBase(BrowserFixture browserFixture, ITestOutputHelper output)
         {
-            _browser.Value = browserFixture.Browser;
-            _logs.Value = browserFixture.Logs;
+            _fixture.Value = browserFixture;
```
rynowak: Will revert this, I ended up not needing it.

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/EventTest.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            var output = Browser.FindElement(By.Id("output"));
+            Assert.Equal(string.Empty, output.Text);
+
+            var other = Browser.FindElement(By.Id("other"));
+
+            // Mouse over the button and then back off
+            var actions = new Actions(Browser)
+                .MoveToElement(input)
+                .MoveToElement(other);
```
rynowak: I'm far from a webdriver expert but this seemed to be what they recommend for general interactivity scripting. 

We also have the ability to run javascript to trigger events, but I wanted to avoid that until its absolutely necessary because that seems like a 'less good' way to test.

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRendererEventDispatcher.cs
SteveSandersonMS: Yes, that would be an improvement. As part of fixing the non-bubbling events, I'll enable passthrough of `unknown` event types, mapping them simply to `UIEventHandler`. We can consider ways of including custom data as a future enhancement.

Just the notifications alone cover a large proportion of scenarios.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -28,12 +34,15 @@ function isCheckbox(element: Element | null) {
...
 // The following interfaces must be kept in sync with the UIEventArgs C# classes
 
-type EventArgsType = 'mouse' | 'keyboard' | 'change' | 'unknown';
+type EventArgsType = 'mouse' | 'keyboard' | 'change' | 'focus' | 'unknown';
 
 export interface UIEventArgs {
   Type: string;
 }
 
+export interface UIFocusEventArgs extends UIEventArgs {
```
SteveSandersonMS: I don't think this interface needs to be exported, as we're only using the type parameters within this file as a way of checking that the constructor args are of a recognized format.

It wouldn't be *harmful* to export it either, but if we are doing so, it would be good to export all the others too for consistency.

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRendererEventDispatcher.cs
SteveSandersonMS: BTW I've handled this in #722 so if that PR goes in first, you could revert this change to `default`.

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/EventTest.cs
SteveSandersonMS: Oh, well, I should have read this earlier. Just realised I've also implemented similar E2E events coverage in #722. Sorry for duplicating your work! I even used the same filename 😞 

I guess it would be good to merge what we've done together. Your test has more coverage of the mouse events, whereas mine has more coverage of the bubbling behaviors.

test/testapps/BasicTestApp/wwwroot/index.html
```diff
@@ -8,32 +8,34 @@
...
+          <option value="BasicTestApp.BindCasesComponent">@bind cases</option>
+          <option value="BasicTestApp.ExternalContentPackage">External content package</option>
+          <option value="BasicTestApp.SvgComponent">SVG</option>
+          <option value="BasicTestApp.SvgWithChildComponent">SVG with child component</option>
+          <option value="BasicTestApp.LogicalElementInsertionCases">Logical element insertion cases</option>
+          <option value="BasicTestApp.ElementRefComponent">Element ref component</option>
+          <option value="BasicTestApp.ComponentRefComponent">Component ref component</option>
+          <option value="BasicTestApp.AfterRenderInteropComponent">After-render interop component</option>
+          <!--<option value="BasicTestApp.RouterTest.Default">Router</option> Excluded because it requires additional setup to work correctly when loaded manually -->
+      </select>
```
SteveSandersonMS: Yes, I guess it's time to stop fighting VS over the indentation here!

rynowak: The alternative would be to use an `.editorsettings` - VS should honour that.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
rynowak: Ah right, I didn't notice that. Will undo the export until we actually need these types exported

rynowak:
@SteveSandersonMS - I added everything from my list.

I'm planning to merge when the CI is green unless you have major feedback. I felt like adding the types was important because you have 4-5 things that are required just to make the type work end to end. I didn't add the data because I don't feel good doing that without tests. This is makes it much easier for someone to do a contribution if they care about a specific event or piece of data.

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/EventTest.cs
```diff
@@ -0,0 +1,130 @@
...
+            actions.Perform();
+            Assert.Equal("onmousedown,", output.Text);
+
+            actions = new Actions(Browser).Release(input);
+
+            actions.Perform();
+            Assert.Equal("onmousedown,onmouseup,", output.Text);
+        }
+
+        private string[] GetLogLines()
```
SteveSandersonMS: Looks like `GetLogLines` and `TriggerCustomBubblingEvent` aren't used, if I'm reading this correctly.

SteveSandersonMS:
@rynowak Excellent - thanks for being so quick with all this! Minor point about possibly unused test code, but don't let it stop you from cherry-picking to `release/0.3.0`.

rynowak:
6903327e404916731d723ffb3c2e27c6af4bbcc3

PR #720, opened 2018-05-01
For checkboxes, bind to 'checked'. Fix special property handling in BrowserRenderer.ts. Fixes #659 and #703

### LINKED ISSUE #659: Support removing 'value' attribute
**Runtime Environment:**
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.300-preview2-008533\

**Repro**

```
@page "/"

<input value="@Text" onchange="@onTextChange" />
<div>Currently text: @Text</div>
<div onclick="@onButtonClick">SET TEXT TO NULL</div>

@{ 
	Console.WriteLine($"Render with: '{Text ?? "NULL"}'");
}

@functions
{
	public string Text { get; set; }

	protected void onTextChange(UIChangeEventArgs e)
	{
		Text = (string)e.Value;
	}

	protected void onButtonClick()
	{
		Text = null;
	}
}
```

When i type some text to input (click outside for call "onchange"). And then i click to "SET TEXT TO NULL". In console i see `Render with: 'NULL'` but input keep last value.

It fail same if i use `<input bind="@Text" />` (and remove onTextChange).

**Excepted behavior**

Reset input text. (same like for `<div>@Text</div>`).

**Workaround**

`<input value="@(Text ?? string.Empty)" onchange="@onTextChange" />`

### LINKED ISSUE #703: value handling for checkbox initial state is dependent on the attribute ordering
It looks like this logic: https://github.com/aspnet/Blazor/blob/dev/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts#L227

Depends on the order in which attributes are applied to an element.

On the initial render pass if you end up with C# code like:
```
            builder.AddAttribute(60, "id", "checkbox-initially-unchecked");
            builder.AddAttribute(61, "value", Microsoft.AspNetCore.Blazor.Components.BindMethods.GetValue(checkboxInitiallyUncheckedValue));
            builder.AddAttribute(62, "onchange", Microsoft.AspNetCore.Blazor.Components.BindMethods.SetValueHandler(__value => checkboxInitiallyUncheckedValue = __value, checkboxInitiallyUncheckedValue));
            builder.AddAttribute(63, "type", "checkbox");
```

The `type` attribute will be set after the value attribute, which will prevent it from being remapped to 'checked'.

Some changes I'm making in the HTML handling exposed this issue by making the actual attribute order in the generated code reflect what the user typed in more deterministically, which means that this code from our E2E tests breaks.
```
<p>
    Initially checked:
    <input id="checkbox-initially-checked" bind="checkboxInitiallyCheckedValue" type="checkbox" />
    <span id="checkbox-initially-checked-value">@checkboxInitiallyCheckedValue</span>
</p>
```

Reordering the `bind` and `type` attributes in the test code also works around the issue.

I'm planning to work around this issue for `bind` for now by forcibly adding the value attribute at the end.

SteveSandersonMS:
At first I was going to fix #703 by changing how attributes are applied so we capture all their values up front and can make decisions in an order-independent way. But then I realised that if we treated checkboxes properly then the ordering issue wouldn't exist anyway.

This PR changes `bind` for checkboxes to use the `checked` property instead of `value`. It should always have done that really, but didn't as a historical consequence of the old `@bind(...)` syntax. With the new syntax we can just treat checkboxes properly and the issue goes away.

This PR also fixes #659 since now we have more consistent (and simpler) handling of "special" attributes.

### Breaking change

People who previously used `<input type='checkbox' value=@someBoolValue />` should now instead write `<input type='checkbox' checked=@someBoolValue />`. This makes a lot more sense anyway.

It doesn't affect any code that uses `bind=@someBoolValue`, because we now map that to the new correct property.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
@@ -224,18 +224,14 @@ export class BrowserRenderer {
       case 'INPUT':
       case 'SELECT':
       case 'TEXTAREA':
-        if (isCheckbox(element)) {
-          (element as HTMLInputElement).checked = value === '';
-        } else {
-          (element as any).value = value;
+        (element as any).value = value;
 
-          if (element.tagName === 'SELECT') {
-            // <select> is special, in that anything we write to .value will be lost if there
-            // isn't yet a matching <option>. To maintain the expected behavior no matter the
-            // element insertion/update order, preserve the desired value separately so
-            // we can recover it when inserting any matching <option>.
-            element[selectValuePropname] = value;
-          }
+        if (element.tagName === 'SELECT') {
+          // <select> is special, in that anything we write to .value will be lost if there
+          // isn't yet a matching <option>. To maintain the expected behavior no matter the
+          // element insertion/update order, preserve the desired value separately so
+          // we can recover it when inserting any matching <option>.
+          element[selectValuePropname] = value;
         }
         return true;
       case 'OPTION':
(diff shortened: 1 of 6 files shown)


SteveSandersonMS:
@dlr1 Well, the good news is that issue is fixed by this exact PR 😺

dlr1:
Ok!, will wait for the merge, Thanks

PR #751, opened 2018-05-04
In Blazor cshtml, auto-reference Microsoft.AspNetCore.Blazor and .Components

### LINKED ISSUE #749: ParameterAttribute could not be found in a new library
`dotnet new blazorlib`

Add sample property

```
@functions {
int currentCount = 0;

[Parameter]
private int IncrementAmount { get; set; } = 1;

void IncrementCount()
{
    currentCount += IncrementAmount;
}
}
```

error CS0246: The type or namespace name 'ParameterAttribute' could not be found (are you missing a using directive or an assembly reference?)

I have to add
`@using Microsoft.AspNetCore.Blazor.Components`
in order to make it work.

SteveSandersonMS:
Fixes #749.

I know that arguably we don't *have* to do this (C# developers already add `using` for things they need), but for a Blazor cshtml file there's basically no scenario where you'd ever not want to import `Microsoft.AspNetCore.Blazor` and `Microsoft.AspNetCore.Blazor.Components`. So this PR adds those namespaces to the ones implicitly referenced both at design time and build time. Can't think of any reason we shouldn't do this.

It looks like a big PR because so many baseline files are affected, but really it's quite a small change. For easier review, you might want to consider each of the commits separately.

I'm not proposing to auto-add other namespaces like `.Routing`, `.Layouts`, `.Browser.Interop`, etc., because there are plenty of cases where you might not want them, and their type names might clash with some other library you're using.

### DIFF
--- a/samples/StandaloneApp/_ViewImports.cshtml
+++ b/samples/StandaloneApp/_ViewImports.cshtml
@@ -1,6 +1,4 @@
 ﻿@using System.Net.Http
-@using Microsoft.AspNetCore.Blazor
-@using Microsoft.AspNetCore.Blazor.Components
 @using Microsoft.AspNetCore.Blazor.Layouts
 @using Microsoft.AspNetCore.Blazor.Routing
 @using StandaloneApp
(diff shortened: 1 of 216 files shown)


rynowak:
This seems like a good idea to me.

PR #771, opened 2018-05-05
Event payload

galvesribeiro:
Enhanced DOM events with proper payload.

Every commit comes with a particular (or set of related) event(s). 

I'll rebase/reset once the review is over.

~~Blockers:~~

~~- Touch events are blocked by #776~~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+# All Files
+[*]
+charset = utf-8
+end_of_line = crlf
+indent_style = space
+indent_size = 4
+insert_final_newline = false
+trim_trailing_whitespace = true
+
+# Solution Files
+[*.sln]
+indent_style = tab
+
+# Markdown Files
+[*.md]
+trim_trailing_whitespace = false
+
+# Web Files
+[*.{htm,html,js,ts,css,scss,less}]
+insert_final_newline = true
+indent_size = 2
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@ artifacts/
 msbuild.binlog
 .vscode/
 BenchmarkDotNet.Artifacts/
-node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ artifacts/
 msbuild.binlog
 .vscode/
 BenchmarkDotNet.Artifacts/
+node_modules/
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ msbuild.binlog
 .vscode/
 BenchmarkDotNet.Artifacts/
 node_modules/
-package-lock.json
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
@@ -1,4 +1,4 @@
-﻿export class EventForDotNet<TData extends UIEventArgs> {
+export class EventForDotNet<TData extends UIEventArgs> {
   constructor(public readonly type: EventArgsType, public readonly data: TData) {
   }
 
@@ -38,7 +38,8 @@
       case 'keydown':
       case 'keyup':
       case 'keypress':
-        return new EventForDotNet<UIKeyboardEventArgs>('keyboard', { Type: event.type, Key: (event as any).key });
+        const keyEvent = event as any;
+        return new EventForDotNet<UIKeyboardEventArgs>('keyboard', { Type: keyEvent.type, Char: keyEvent.char, Key: keyEvent.key, Code: keyEvent.code, Location: keyEvent.location, Repeat: keyEvent.repeat, Locale: keyEvent.locale, CtrlKey: keyEvent.ctrlKey, ShiftKey: keyEvent.shiftKey, AltKey: keyEvent.altKey, MetaKey: keyEvent.metaKey });
 
       case 'contextmenu':
       case 'click':
@@ -110,7 +111,16 @@ interface UIFocusEventArgs extends UIEventArgs {
 }
 
 interface UIKeyboardEventArgs extends UIEventArgs {
+  Char: string;
   Key: string;
+  Code: string;
+  Location: number;
+  Repeat: boolean;
+  Locale: string;
+  CtrlKey: boolean;
+  ShiftKey: boolean;
+  AltKey: boolean;
+  MetaKey: boolean;
 }
 
 interface UIMouseEventArgs extends UIEventArgs {
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
@@ -76,14 +76,24 @@ export class EventForDotNet<TData extends UIEventArgs> {
       case 'pointerup':
         return new EventForDotNet<UIPointerEventArgs>('pointer', parsePointerEvent(event));
 
+      case 'wheel':
       case 'mousewheel':
-        return new EventForDotNet<UIWheelEventArgs>('wheel', { Type: event.type });
-
+        return new EventForDotNet<UIWheelEventArgs>('wheel', parseWheelEvent(event));
 
       default:
         return new EventForDotNet<UIEventArgs>('unknown', { Type: event.type });
     }
   }
+}
+
+function parseWheelEvent(event: any) {
+  return Object.assign(parseMouseEvent(event),
+    {
+      DeltaX: event.deltaX,
+      DeltaY: event.deltaY,
+      DeltaZ: event.deltaZ,
+      DeltaMode: event.deltaMode
+    });
 }
 
 function parseProgressEvent(event: any) {
@@ -251,5 +261,9 @@ interface UITouchPoint {
   PageY: number;
 }
 
-interface UIWheelEventArgs extends UIEventArgs {
+interface UIWheelEventArgs extends UIMouseEventArgs {
+  DeltaX: number;
+  DeltaY: number;
+  DeltaZ: number;
+  DeltaMode: number;
 }
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
@@ -24,7 +24,7 @@ export class EventForDotNet<TData extends UIEventArgs> {
       case 'dragover':
       case 'dragstart':
       case 'drop':
-        return new EventForDotNet<UIDragEventArgs>('drag', { Type: event.type });
+        return new EventForDotNet<UIDragEventArgs>('drag', parseDragEvent(event));
 
       case 'focus':
       case 'blur':
@@ -83,6 +83,25 @@ export class EventForDotNet<TData extends UIEventArgs> {
       default:
         return new EventForDotNet<UIEventArgs>('unknown', { Type: event.type });
     }
+  }
+}
+
+function parseDragEvent(event: any) {
+  return {
+    Type: event.type,
+    Detail: event.detail,
+    DataTransfer: event.dataTransfer,
+    ScreenX: event.screenX,
+    ScreenY: event.screenY,
+    ClientX: event.clientX,
+    ClientY: event.clientY,
+    Button: event.button,
+    Buttons: event.buttons,
+    MozPressure: event.mozPressure,
+    CtrlKey: event.ctrlKey,
+    ShiftKey: event.shiftKey,
+    AltKey: event.altKey,
+    MetaKey: event.metaKey
   }
 }
 
@@ -187,6 +206,32 @@ interface UIClipboardEventArgs extends UIEventArgs {
 }
 
 interface UIDragEventArgs extends UIEventArgs {
+  Detail: number;
+  DataTransfer: UIDataTransfer;
+  ScreenX: number;
+  ScreenY: number;
+  ClientX: number;
+  ClientY: number;
+  Button: number;
+  Buttons: number;
+  MozPressure: number;
+  CtrlKey: boolean;
+  ShiftKey: boolean;
+  AltKey: boolean;
+  MetaKey: boolean;
+}
+
+interface UIDataTransfer {
+  DropEffect: string;
+  EffectAllowed: string;
+  Files: string[];
+  Items: UIDataTransferItem[];
+  Types: string[];
+}
+
+interface UIDataTransferItem {
+  Kind: string;
+  Type: string;
 }
 
 interface UIErrorEventArgs extends UIEventArgs {
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
@@ -26,9 +26,6 @@ export class EventForDotNet<TData extends UIEventArgs> {
       case 'drop':
         return new EventForDotNet<UIDragEventArgs>('drag', { Type: event.type });
 
-      case 'error':
-        return new EventForDotNet<UIProgressEventArgs>('error', { Type: event.type });
-
       case 'focus':
       case 'blur':
       case 'focusin':
@@ -50,8 +47,14 @@ export class EventForDotNet<TData extends UIEventArgs> {
       case 'dblclick':
         return new EventForDotNet<UIMouseEventArgs>('mouse', parseMouseEvent(event));
 
+      case 'loadstart':
+      case 'timeout':
+      case 'abort':
+      case 'load':
+      case 'loadend':
+      case 'error':
       case 'progress':
-        return new EventForDotNet<UIProgressEventArgs>('progress', { Type: event.type });
+        return new EventForDotNet<UIProgressEventArgs>('progress', parseProgressEvent(event));
 
       case 'touchcancel':
       case 'touchend':
@@ -83,8 +86,16 @@ export class EventForDotNet<TData extends UIEventArgs> {
   }
 }
 
+function parseProgressEvent(event: any) {
+  return {
+    Type: event.type,
+    LengthComputable: event.legthComputable,
+    Loaded: event.loaded,
+    Total: event.total
+  };
+}
+
 function parseTouchEvent(event: any) {
-  console.log(event);
   return {
     Type: event.type,
     Detail: event.detail,
@@ -95,7 +106,7 @@ function parseTouchEvent(event: any) {
     ShiftKey: event.shiftKey,
     AltKey: event.altKey,
     MetaKey: event.metaKey
-  }
+  };
 }
 
 function parseKeyboardEvent(event: any) {
@@ -111,7 +122,7 @@ function parseKeyboardEvent(event: any) {
     ShiftKey: event.shiftKey,
     AltKey: event.altKey,
     MetaKey: event.metaKey
-  }
+  };
 }
 
 function parsePointerEvent(event: any) {
@@ -143,7 +154,7 @@ function parseMouseEvent(event: any) {
     ShiftKey: event.shiftKey,
     AltKey: event.altKey,
     MetaKey: event.metaKey
-  }
+  };
 }
 
 function isCheckbox(element: Element | null) {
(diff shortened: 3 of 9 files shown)


galvesribeiro:
Looks like this is the cause for the warning:

https://github.com/aspnet/Blazor/blob/dev/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts#L312

Not sure yet if related to the error...

galvesribeiro:
Ok, redy for review.

All existent events were filled with payload properties of types that aren't references.

@rynowak here is what we discussed in the other issue.

Please let me know what you think.

senthilnathansk:
Hi. "oninput" event of text field's payload (value) covered? that would be helpful. thanks.

galvesribeiro:
@senthilnathansk 

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent), `oninput` is an experimental API. The event which will carry the `value` of the change is `onchange` and it is already implemented even before this PR.

I've added payload to all events that was implemented already (added a few missing) and were stable on MDN/W3C.

Lets see what @rynowak and other folks from the team think about experimental APIs and if that is the case, it can be easily added later on.

senthilnathansk:
@galvesribeiro thanks for the reply. 

yes. I guess the difference in "onchange" and "oninput" is that "onchange" fired when the field is blured. in React "onchange" on text input behaves like "oninput" event. This event is helpful to implement validation as soon as the user enter values. 

And also I could not see "oninput" event is experiment here.

 https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oninput 

https://developer.mozilla.org/en-US/docs/Web/Events/input

And may be pull request is not the right place for this discussion? I am very new to open source project interaction so apologies in advance.

galvesribeiro:
image

On the link you posted, click on `InputEvent` interface. There is where you will find the _experimental_ note. I believe React by the concept of _syntactic_ events wrap the DOM events and do what you are asking for.

> And may be pull request is not the right place for this discussion? I am very new to open source project interaction so apologies in advance.

Nah 😄 If the subject of the discussion _is_ the PR, feel free. Perhaps if you want to ask for experimental APIs, it would be best to open a new issue so we can discuss it there.

galvesribeiro:
@rynowak I think the Jenkins error is unrelated to the PR. Can you fire the build again just to be sure?

galvesribeiro:
As @SteveSandersonMS asked on #774 I've updated the casing on all properties of the events.

rynowak:
Sorry for the slow responses. I'll take a look at this in more detail soon.

.gitignore
```diff
@@ -6,4 +6,6 @@ launchSettings.json
 artifacts/
 msbuild.binlog
 .vscode/
-BenchmarkDotNet.Artifacts/
\ No newline at end of file
+BenchmarkDotNet.Artifacts/
+node_modules/
+package-lock.json
```
rynowak: The lock file is not intended to be ignored, we've seen problems with this file wanting to change between different versions of node and windows/non-windows - but this isn't a solution

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -90,40 +214,120 @@ function isCheckbox(element: Element | null) {
...
+  pageX: number;
+  pageY: number;
 }
 
-interface UIWheelEventArgs extends UIEventArgs {
+interface UIWheelEventArgs extends UIMouseEventArgs {
+  deltaX: number;
+  deltaY: number;
+  deltaZ: number;
+  deltaMode: number;
```
rynowak: This all seems like it makes sense to me. /cc @SteveSandersonMS in case there's anything he wants to comment on

src/Microsoft.AspNetCore.Blazor/Components/EventHandlers.cs
```diff
@@ -96,11 +99,14 @@ namespace Microsoft.AspNetCore.Blazor.Components
...
-    [EventHandler("onerror", typeof(UIErrorEventArgs))]
-
     // Progress events
+    [EventHandler("onloadstart", typeof(UIProgressEventArgs))]
+    [EventHandler("ontimeout", typeof(UIProgressEventArgs))]
+    [EventHandler("onabort", typeof(UIProgressEventArgs))]
+    [EventHandler("onload", typeof(UIProgressEventArgs))]
+    [EventHandler("onloadend", typeof(UIProgressEventArgs))]
     [EventHandler("onprogress", typeof(UIProgressEventArgs))]
+    [EventHandler("onerror", typeof(UIErrorEventArgs))]
```
rynowak: These additions are also defined below in the next section.. what am I missing here?

src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs
```diff
@@ -62,43 +193,330 @@ public class UIFocusEventArgs : UIEventArgs
...
+
+        /// <summary>
+        /// The scroll amount for the z-axis.
+        /// </summary>
+        public double DeltaZ { get; set; }
+
+        /// <summary>
+        /// The unit of the delta values scroll amount. 
+        /// </summary>
+        public long DeltaMode { get; set; }
```
rynowak: This is really nice 👍

test/testapps/BasicTestApp/KeyPressEventComponent.cshtml
```diff
@@ -13,6 +13,7 @@ Type here: <input onkeypress=@OnKeyPressed />
 
     void OnKeyPressed(UIKeyboardEventArgs eventArgs)
     {
+        Console.WriteLine(JsonUtil.Serialize(eventArgs));
```
rynowak: Assuming this just here for crude debugging so ignoring for now.

test/testapps/BasicTestApp/TouchEventComponent.cshtml
```diff
@@ -0,0 +1,38 @@
...
+    {
+        message += e.Type;
+        Console.WriteLine(JsonUtil.Serialize(e));
+        StateHasChanged();
+    }
+
+    void Clear()
+    {
+        message = string.Empty;
+    }
```
rynowak: Would it be possible to write an E2E test for this? Similar to https://github.com/aspnet/Blazor/blob/dev/test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/EventTest.cs#L91

You should be able to use https://seleniumhq.github.io/selenium/docs/api/dotnet/html/T_OpenQA_Selenium_Interactions_TouchActions.htm

.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+indent_style = tab
+
+# Markdown Files
+[*.md]
+trim_trailing_whitespace = false
+
+# Web Files
+[*.{htm,html,js,ts,css,scss,less}]
+insert_final_newline = true
+indent_size = 2
```
rynowak: Did you add this to make it easier to work with the ts formatting settings? If so then I approve. These settings largely match what we do always so that's fine.

galvesribeiro: Yeah, I added that after other community members that made previous contributions say what was the standard. The default on VS is 4 spaces so that would mess all the files. Having that file makes future contributions easier as VS2017 now respect those files by default.

test/testapps/BasicTestApp/KeyPressEventComponent.cshtml
galvesribeiro: Yeah, I left it there and on touch test so we can watch the full events while testing without we break the current E2E.

src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs
galvesribeiro: Yeah... That data allow some _infinity scroll_ behavior to be implemented.

.gitignore
galvesribeiro: Yeah, I agree. In all other node-based projects, we have the lock file checked in. As it wasn't on Blazor repo, I just made it ignored. When I have the approval for the PR, I'll reset/rebase and push again and will remove the lock file from there and add to the repo.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
galvesribeiro: I made them all camel case as per @SteveSandersonMS request on another issue. They should match the same structure on the c# side.

src/Microsoft.AspNetCore.Blazor/Components/EventHandlers.cs
galvesribeiro: Yes, that  next section is wrong. I think I mistakenly ctr+z on the delete line of the next section. Fixing.

.gitignore
```diff
@@ -6,4 +6,5 @@ launchSettings.json
 artifacts/
 msbuild.binlog
 .vscode/
-BenchmarkDotNet.Artifacts/
\ No newline at end of file
+BenchmarkDotNet.Artifacts/
+node_modules/
```
SteveSandersonMS: If we're adding `node_modules` here, then can we also *remove* it from being repeated in any other `.gitignore` files in subdirectories? Otherwise we can continue gitignoring it only in the subdirectories.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -69,18 +74,137 @@
...
+    type: event.type,
+    detail: event.detail,
+    dataTransfer: event.dataTransfer,
+    screenX: event.screenX,
+    screenY: event.screenY,
+    clientX: event.clientX,
+    clientY: event.clientY,
+    button: event.button,
+    buttons: event.buttons,
+    mozPressure: event.mozPressure,
```
SteveSandersonMS: Since in general we've managed to avoid vendor-specific things so far, would it be OK not to have `mozPressure`? Once it becomes a nonprefixed standard we could re-add it.

.gitignore
galvesribeiro: Hey @SteveSandersonMS 

Yeah, I can remove from the subdirectories. In fact, there is no need for having multiple .gitignore on subdirectories. The code base is relatively small, so having one on the root should be enough.

The way it is now, in the subdirectory, for whatever reason was still making git to try add _node_modules_ to my changelist.

Do you want me to remove the ones from the subdirectories or just remove the _node_modules_ entry on those?

Thanks

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
galvesribeiro: Fixing.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -69,18 +74,137 @@
...
 
       default:
-        return new EventForDotNet<UIEventArgs>('unknown', { Type: event.type });
+        return new EventForDotNet<UIEventArgs>('unknown', { type: event.type });
     }
   }
 }
 
+function parseDragEvent(event: any) {
+  return {
```
SteveSandersonMS: Unimportant stylistic point: with TypeScript this can be expressed without duplicating the key names, e.g.:

    return { event.type, event.detail, event.dataTransfer, ...}

It compiles to the same result.

galvesribeiro: ahhh ok ;) live and learn 👍 

Fixing.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -69,18 +74,137 @@
...
+    mozPressure: event.mozPressure,
+    ctrlKey: event.ctrlKey,
+    shiftKey: event.shiftKey,
+    altKey: event.altKey,
+    metaKey: event.metaKey
+  }
+}
+
+function parseWheelEvent(event: any) {
+  return Object.assign(parseMouseEvent(event),
```
SteveSandersonMS: Another possible TypeScript syntax simplification:

```
return { ...parseMouseEvent(event), event.deltaX, event.deltaY, event.deltaZ, event.deltaMode };
```

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
galvesribeiro: Looks like it didn't like that:
image

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -69,18 +74,137 @@
...
+      deltaY: event.deltaY,
+      deltaZ: event.deltaZ,
+      deltaMode: event.deltaMode
+    });
+}
+
+function parseProgressEvent(event: any) {
+  return {
+    type: event.type,
+    lengthComputable: event.legthComputable,
```
SteveSandersonMS: Typo in `legthComputable`. 

This indicates that we should be using TypeScript types to avoid the risk of errors like that. In this example, `event` could be typed as `ProgressEvent`. Would it be possible to strongly-type all the `event` objects we're using in this file?

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -69,18 +74,137 @@
...
+    loaded: event.loaded,
+    total: event.total
+  };
+}
+
+function parseTouchEvent(event: any) {
+
+  function parseTouch(touchList: TouchList) {
+    const touches: UITouchPoint[] = [];
```
SteveSandersonMS: Stylistic point: when mapping arrays in TypeScript, can we use `.map` instead of a `for` loop? This function can be reduced to pretty much a one-liner.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -90,40 +214,120 @@ function isCheckbox(element: Element | null) {
...
 
 interface UIMouseEventArgs extends UIEventArgs {
+  detail: number;
+  screenX: number;
+  screenY: number;
+  clientX: number;
+  clientY: number;
+  button: number;
+  buttons: number;
+  mozPressure: number;
```
SteveSandersonMS: As above, would prefer not to have vendor-prefixed items.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
galvesribeiro: Agreed. Will do that.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
SteveSandersonMS: OK then so based on the key names for `deltaX`, `deltaY`, etc would have to be duplicated. But the `...parseMouseEvent(event)` should still work. It compiles to an "assign" call anyway.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
SteveSandersonMS: Just to make sure we're talking about the same thing, TypeScript has *built-in* types for all these DOM events (e.g., `ProgressEvent`, `PointerEvent`, etc. are built into the standard library). You don't need to write those type definitions yourself.

SteveSandersonMS: (However I do agree with your decision to map a specific subset of properties out of them into the `UiXyzEventArgs` types)

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
galvesribeiro: Yeah I know... But unfortunately, the `TouchList` type can't be iterated :(

We have to call `tl.item(idx)`... Weird, but yeah, that is the way...

.gitignore
galvesribeiro: Ok. So for now I'll remove it from the .gitignore.

galvesribeiro: Just for the reference:

image

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
galvesribeiro: Yeah, I know. That is the reason I made the types myself. Some of them will not be useful in Blazor _today_ like the ones which are references to other objects. 

We can ofc in the future drop those as they are there just to guide ourselves and no public API would be affected.

galvesribeiro: Btw, I only left the `parseDragEvent(event: any)` as `any` because the `datatransfer` object is really tricky so to make it easier I left it `any`. However, I did the test before changing to `DragEvent` and no types were found :)

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
SteveSandersonMS: Ah OK, thanks for clarifying!

SteveSandersonMS:
Thanks for the updates, @galvesribeiro! This looks great. There does seem to be a build error now though: https://ci.appveyor.com/project/aspnetci/blazor/build/1.0.861#L629

Once that's fixed I think we'd be good to merge this.

galvesribeiro:
Sorry... For whatever reason sometimes VS don't invoke the webpack tasks... I had the same problems with Blazor.Extensions projects and had to manually delete the dist/ to force webpack to run again. After a clean build I found the issue.

galvesribeiro:
Hello @SteveSandersonMS 

Is there anything else you want to me to change?

Thanks

galvesribeiro:
Ping

SteveSandersonMS:
This looks great, @galvesribeiro! No further changes needed. We'll merge this in time for the 0.4.0 milestone.

Thanks so much for contributing this. It will be extremely useful.

SteveSandersonMS:
Thanks again. This is now merged into `dev`.

galvesribeiro:
Thanks @stevesandersonms

Just curious... any particular reason why _this_ PR wasn’t merged from here?

SteveSandersonMS:
It had to be rebased and squashed first.

galvesribeiro:
Ahh ok. I could do that if you want. I even mentioned first :) 

Next time I’ll do that before the merge. 

Thanks

rpedretti:
I Have found a problem when bindig to ondragstart and ondragend to a C# function where the ondragstart whould fire but the ondragend would not. But changing to a javscript function it would work perfectly. Does this change fix this?

galvesribeiro:
I think so @rpedretti 

I tested the drag events on both chrome and Edge here and they worked properly.

rpedretti:
So I got myself some time to fork the repo and tested it. Unfrtunately the bug is there.
```
@page "/"
@using Microsoft.AspNetCore.Blazor

<h1 draggable="true" ondragstart="@Dragstart" ondragend="@Dragend">Hello, world!</h1>

Welcome to your new app.

@functions {
    private void Dragstart(UIDragEventArgs a)
    {
        Console.WriteLine("Start");
    }


    private void Dragend(UIDragEventArgs a)
    {
        Console.WriteLine("end");
    }
}
```

Having this at homepage show the bug. I'll file a new issue

galvesribeiro:
Weird... Look at this comment https://github.com/aspnet/Blazor/pull/771#issuecomment-387100232 The picture show the end event just fine...

Are you sure you are not restoring anything from the wrong place? I mean, the sample page to me works perfectly on both drag and touch events...

rpedretti:
Well, i cloned, built and started on a different port then my project. I tested all combination of _ondrag*_ events and only the _ondragstart_ break things.

PR #787, opened 2018-05-07
Simple autorebuild mechanism for hosted apps

SteveSandersonMS:
In the 0.3.0 release we had to disable autorebuild for hosted apps due to a bug. This PR re-enables it, replacing the implementation with a much simpler alternative.

This implementation should be understood as a stopgap measure until some future version of ASP.NET Core and VS can hopefully extend the native autorebuild support to rebuild when there are changes in referenced projects. Once (and if) that is implemented, both the standalone and hosted variants of Blazor's autorebuild feature could be removed because both would inherit it from ASP.NET Core.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
@@ -3,7 +3,9 @@
 
 using Microsoft.AspNetCore.Blazor.Server;
 using Microsoft.AspNetCore.Blazor.Server.AutoRebuild;
+using Microsoft.Extensions.Logging;
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
@@ -19,7 +21,37 @@ namespace Microsoft.AspNetCore.Builder
         private static string[] _includedSuffixes = new[] { ".cs", ".cshtml", "index.html" };
         private static string[] _excludedDirectories = new[] { "obj", "bin" };
 
-        public static void UseAutoRebuild(this IApplicationBuilder appBuilder, BlazorConfig config)
+        public static void UseHostedAutoRebuild(this IApplicationBuilder appBuilder, BlazorConfig config, string hostAppContentRootPath)
+        {
+            var isFirstFileWrite = true;
+            WatchFileSystem(config, () =>
+            {
+                if (isFirstFileWrite)
+                {
+                    try
+                    {
+                        var fileToTouch = FindFirstFileInDirectoryRecursive(hostAppContentRootPath, "*.cs");
+                        if (!string.IsNullOrEmpty(fileToTouch))
+                        {
+                            File.SetLastWriteTime(fileToTouch, DateTime.Now);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        // If we don't have permission to write these files, autorebuild will not be enabled
+                        var loggerFactory = (ILoggerFactory)appBuilder.ApplicationServices.GetService(typeof(ILoggerFactory));
+                        var logger = loggerFactory?.CreateLogger(typeof (AutoRebuildExtensions));
+                        logger?.LogInformation(ex,
+                            "Cannot autorebuild because there was an error when writing to a file in '{0}'.",
+                            hostAppContentRootPath);
+                    }
+
+                    isFirstFileWrite = false;
+                }
+            });
+        }
+
+        public static void UseDevServerAutoRebuild(this IApplicationBuilder appBuilder, BlazorConfig config)
         {
             // Currently this only supports VS for Windows. Later on we can add
             // an IRebuildService implementation for VS for Mac, etc.
@@ -76,6 +108,12 @@ namespace Microsoft.AspNetCore.Builder
                 await next();
             });
         }
+
+        private static string FindFirstFileInDirectoryRecursive(string rootDirectory, string pattern)
+            => Directory.GetFiles(rootDirectory, pattern).FirstOrDefault()
+            ?? Directory.GetDirectories(rootDirectory)
+                .Select(dir => FindFirstFileInDirectoryRecursive(dir, pattern))
+                .FirstOrDefault();
 
         private static void WatchFileSystem(BlazorConfig config, Action onWrite)
         {
--- a/samples/HostedInAspNet.Client/HostedInAspNet.Client.csproj
+++ b/samples/HostedInAspNet.Client/HostedInAspNet.Client.csproj
@@ -3,9 +3,6 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <OutputType>Exe</OutputType>
-
-    <!-- Temporary workaround for a VS build issue -->
-    <BlazorRebuildOnFileChange>false</BlazorRebuildOnFileChange>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Microsoft.AspNetCore.Blazor.Server/BlazorAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/BlazorAppBuilderExtensions.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.FileProviders;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Net.Http.Headers;
 using System.Net.Mime;
+using System;
 
 namespace Microsoft.AspNetCore.Builder
 {
@@ -16,6 +17,8 @@ namespace Microsoft.AspNetCore.Builder
     /// </summary>
     public static class BlazorAppBuilderExtensions
     {
+        const string DevServerApplicationName = "dotnet-blazor";
+
         /// <summary>
         /// Configures the middleware pipeline to work with Blazor.
         /// </summary>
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
```diff
@@ -19,7 +21,37 @@ internal static class AutoRebuildExtensions
...
+                        var fileToTouch = FindFirstFileInDirectoryRecursive(hostAppContentRootPath, "*.cs");
+                        if (!string.IsNullOrEmpty(fileToTouch))
+                        {
+                            File.SetLastWriteTime(fileToTouch, DateTime.Now);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        // If we don't have permission to write these files, autorebuild will not be enabled
+                        var loggerFactory = (ILoggerFactory)appBuilder.ApplicationServices.GetService(typeof(ILoggerFactory));
```
attilah: Why not using the generic versions of the DI methods?

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
```diff
@@ -77,6 +109,12 @@ public static void UseAutoRebuild(this IApplicationBuilder appBuilder, BlazorCon
             });
         }
 
+        private static string FindFirstFileInDirectoryRecursive(string rootDirectory, string pattern)
```
attilah: What do you think of leveraging the new .Net Core 2.1 FileSystemEnumerable, that's more lightweight and offers plenty of options.

Something similar:

```
            var enumerationOptions = new EnumerationOptions
            {
                IgnoreInaccessible = true,
                RecurseSubdirectories = true
            };

            var fileToTouch = new FileSystemEnumerable<string>(hostAppContentRootPath, (ref FileSystemEntry entry) => entry.ToFullPath(), enumerationOptions)
            {
                ShouldIncludePredicate = (ref FileSystemEntry entry) => !entry.IsDirectory && entry.FileName.EndsWith(".cs", StringComparison.OrdinalIgnoreCase)
            };
 ```

Comment (rynowak) [inline src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs]: Is https://msdn.microsoft.com/en-us/library/dd383571(v=vs.110).aspx a suitable replacement for this or is something non-obvious happening here?

File: src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs [hunk 3]
```diff
@@ -19,7 +21,37 @@ internal static class AutoRebuildExtensions
...
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        // If we don't have permission to write these files, autorebuild will not be enabled
+                        var loggerFactory = (ILoggerFactory)appBuilder.ApplicationServices.GetService(typeof(ILoggerFactory));
+                        var logger = loggerFactory?.CreateLogger(typeof (AutoRebuildExtensions));
+                        logger?.LogInformation(ex,
+                            "Cannot autorebuild because there was an error when writing to a file in '{0}'.",
+                            hostAppContentRootPath);
```
rynowak: Should this be a warning?

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
```diff
@@ -19,7 +21,37 @@ internal static class AutoRebuildExtensions
...
+                        if (!string.IsNullOrEmpty(fileToTouch))
+                        {
+                            File.SetLastWriteTime(fileToTouch, DateTime.Now);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        // If we don't have permission to write these files, autorebuild will not be enabled
+                        var loggerFactory = (ILoggerFactory)appBuilder.ApplicationServices.GetService(typeof(ILoggerFactory));
+                        var logger = loggerFactory?.CreateLogger(typeof (AutoRebuildExtensions));
```
rynowak: Is it actually possible for DI to not be registered? I think it's baked in at the lowest possible level in hosted ASP.NET Core.

You could also shorten this to `appBuilder.ApplicationServices.GetRequiredService<ILogger<AutoRebuildExtensions>>()` if you so desire

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
attilah: I've checked the implementation and now ```Directory``` is backed by FileSystemEntry* stuff, which is really cool, the whole recursive file search could be a single liner with (static) options:

```
var enumerationOptions = new EnumerationOptions
{
    IgnoreInaccessible = true,
    RecurseSubdirectories = true
};

var fileToTouch = Directory.GetFiles(hostAppContentRootPath, "*.cs", enumerationOptions).FirstOrDefault();
```

SteveSandersonMS: I would have thought that `RecurseSubdirectories = true` would force it to build a list containing every single `.cs` file in the entire directory hierarchy, of which there might be many. I was trying to let to stop as soon as it finds any directory containing any `.cs` file.

attilah: Fortunately not, if you look at the implementation the same API is used with a different predicate:

https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/Directory.cs#L160

rynowak: From EnumerateFiles remarks: 

> The EnumerateFiles and GetFiles methods differ as follows: When you use EnumerateFiles, you can start enumerating the collection of names before the whole collection is returned; when you use GetFiles, you must wait for the whole array of names to be returned before you can access the array. Therefore, when you are working with many files and directories, EnumerateFiles can be more efficient.

attilah: Yup, that's true, but the single liner above by changing ```GetFiles``` to ```EnumerateFiles``` would still remove the need for Linq and others.

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
```diff
@@ -19,7 +21,37 @@ internal static class AutoRebuildExtensions
         private static string[] _includedSuffixes = new[] { ".cs", ".cshtml", "index.html" };
         private static string[] _excludedDirectories = new[] { "obj", "bin" };
 
-        public static void UseAutoRebuild(this IApplicationBuilder appBuilder, BlazorConfig config)
+        public static void UseHostedAutoRebuild(this IApplicationBuilder appBuilder, BlazorConfig config, string hostAppContentRootPath)
+        {
+            var isFirstFileWrite = true;
+            WatchFileSystem(config, () =>
```
MisinformedDNA: You'll have to excuse my ignorance, but won't `FileSystemWatcher` be GCed?

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
SteveSandersonMS: Done

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
SteveSandersonMS: Good point - done.

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
SteveSandersonMS: `Directory.EnumerateFiles` is a good fit here, so have switched to that. Thanks for the tip.

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
SteveSandersonMS: Yes, that would be reasonable. Done.

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
SteveSandersonMS: I'm sure that would work fine, but `Directory.EnumerateFiles` is even simpler and given the use case (invoked at most once per app build) there is no chance that any perf difference would be observable.

src/Microsoft.AspNetCore.Blazor.Server/AutoRebuild/AutoRebuildExtensions.cs
SteveSandersonMS: Good point! Now accounted for.

PR #795, opened 2018-05-08
Fix 773

rynowak:
We weren't handling a few cases that can occur during typing correctly.
Our passes that look at the content of attributes need to be prepared
for it to be empty in cases where the attribute has been partially
typed in the editor.

I added a smoke test for this that attempts to simulate typing and found
another issue to fix.

The end result of this is that the design for this kind of code is
simpler and takes a more 'brute-force' approach to understanding
attributes. I think this is a good change based on the problems with how
this code has been written today, there are too many possible cases to
try and have the code express and document them all.


rynowak:
@SteveSandersonMS sending this as an FYI - not sure there's too much you'd be interested in

PR #813, opened 2018-05-10
ability to deserialize arrays through IDictionary<string, object>

### LINKED ISSUE #812: Unable to deserialize arrays via IDictionary<string, object>
It's common for me to use JSON.NET to deserialize json of unknown schema like this:
```
var expando = JsonConvert.DeserializeObject<ExpandoObject>(...)
if (expando is IDictionary<string, object> map)
{
    foreach (var kvp in map)
    {
        var propertyName = kvp.Key;
        switch (kvp.Value)
        {
            case string s: Console.WriteLine($"{propertyName}: {s}"); break;
            case IDictionary<string, object> obj: /* recurse...? */ break;
            case IList<IDictionary<string, object>> list: /* recurse in a loop...? */ break;
        }
    }
}
```

This fails with JSON.NET on webassembly but that's not Blazor's problem. I'm trying to do a similar thing with JsonUtil / SimpleJson and it sort of works but arrays come back null. so the following json:

```
{"array1":[{"key1":"value1"},{"key1":"value2"}]}
```

`JsonUtil.Deserialize(...json from above...)`

deserializes as if like this: `new Dictionary<string, object> { "array1": null }`

thus this unit test fails:

```
[Fact]
        public void SupportsDictionaryArrayDeserialization()
        {
            // Arrange/Act
            var result = JsonUtil.Deserialize<IDictionary<string, object>>("{\"array1\":[{\"key1\":\"value1\"},{\"key1\":\"value2\"}]}");

            // Assert
            var array1 = result["array1"] as IList<IDictionary<string, object>>;
            Assert.Equal("value1", array1[0]["key1"]);
            Assert.Equal("value2", array1[1]["key1"]);
        }
```

I have a PR which I will submit shortly to "fix" this issue, although I don't know if it's correct because it's possible this is desired behavior... If you want to pull it, great, otherwise I can just maintain a separate copy of SimpleJson such that it does support this behavior, since it's pretty well self contained.

chris-ray:
Added ability for `JsonUtil.Deserialize<IDictionary<string, object>>(json)` to properly deserialize arrays to support a scenario where you want to deserialize json for which you don't know the schema at compile time.

fixes #812

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
@@ -1497,6 +1497,12 @@ namespace SimpleJson
                             foreach (object o in jsonObject)
                                 list.Add(DeserializeObject(o, innerType));
                         }
+                        else if (type == typeof(object))
+                        {
+                            list = new List<IDictionary<string, object>>();
+                            foreach (object o in jsonObject)
+                                list.Add(DeserializeObject(o, typeof(IDictionary<string, object>)));
+                        }
                         obj = list;
                     }
                 }
--- a/test/Microsoft.AspNetCore.Blazor.Test/JsonUtilTest.cs
+++ b/test/Microsoft.AspNetCore.Blazor.Test/JsonUtilTest.cs
@@ -197,8 +197,20 @@ namespace Microsoft.AspNetCore.Blazor.Test
             // Arrange/Act
             var json = JsonUtil.Serialize(new WithCustomSerializer());
 
-            // Asssert
+            // Assert
             Assert.Equal("{\"key1\":\"value1\",\"key2\":123}", json);
+        }
+
+        [Fact]
+        public void SupportsDictionaryArrayDeserialization()
+        {
+            // Arrange/Act
+            var result = JsonUtil.Deserialize<IDictionary<string, object>>("{\"array1\":[{\"key1\":\"value1\"},{\"key1\":\"value2\"}]}");
+
+            // Assert
+            var array1 = result["array1"] as IList<IDictionary<string, object>>;
+            Assert.Equal("value1", array1[0]["key1"]);
+            Assert.Equal("value2", array1[1]["key1"]);
         }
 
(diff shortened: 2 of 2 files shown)


SteveSandersonMS:
Thanks for proposing this. I've looked into it, and unfortunately I don't think this is the desired way to proceed.

The trouble with this approach is that the items in the array aren't necessarily objects (i.e., can't necessarily be represented by an `IDictionary<string, object>`). For example, this change would cause cause an `InvalidCastException` for `JsonUtil.Deserialize<IDictionary<string, object>>("{\"array1\":[\"something\"]}")` (which previously wouldn't return a useful result, but at least wouldn't crash, making it it safe to have JSON of this form as part of a larger deserialization process). In general the `IDictionary<string, object>` contract implies that the dictionary entries can be anything, which could include numbers, strings, arrays, other dictionaries, and so on.

Instead of doing this, if you really want to work with an unknown JSON schema, an easier way to proceed is just to call `JsonUtil.Deserialize<object>(someJson)`. Then you can inspect the resulting type and perform the necessary runtime casts (e.g., to `IDictionary<string, object>` or `IList<object>` depending on what you receive at runtime).

For example, the following unit test would already pass without any implementation change:

```csharp
[Fact]
public void SupportsDictionaryArrayDeserialization()
{
    // Arrange/Act
    var result = JsonUtil.Deserialize<object>("{\"array1\":[{\"key1\":\"value1\"},{\"key1\":\"value2\"},\"another\"]}");

    // Assert
    var array1 = (result as IDictionary<string, object>)["array1"] as IList<object>;
    Assert.Equal("value1", (array1[0] as IDictionary<string, object>)["key1"]);
    Assert.Equal("value2", (array1[1] as IDictionary<string, object>)["key1"]);
    Assert.Equal("another", array1[2]);
}
```

I'm aware that SimpleJson's built-in handling for the specific `IDictionary<string, object>` case isn't hugely useful, but trying to cover all scenarios is the difference between SimpleJson and a full JSON library like Json.NET. At least there are workarounds like using `Deserialize<object>`.

chris-ray:
Fair and agreed. Thanks for investigating it, and keep up the good work, Steve.

Issue #479, opened 2018-04-03
HttpClient enhancements - binary data


SvenEV:
Currently `HttpClient` cannot be used to retrieve binary data such as images. The response content appears to get truncated at zero bytes when the HTTP response is returned from JS back to .NET. In some cases, requests just fails with "SCRIPT7017: Memory index is out of range" (Edge) or "RuntimeError: memory access out of bounds" (Chrome) without anything being returned.

I found a TODO regarding this issue in the code (see below). Anyway, I think this should be tracked and I hope this isn't a duplicate.

https://github.com/aspnet/Blazor/blob/a9c45fe78ac18b6bd524904b40896c249f9f09c0/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts#L32-L47

SteveSandersonMS:
Thanks @SvenEV - it's good to have an issue to track this. So, starting a list of possible future enhancements:

* Reading binary data
* Streaming responses

rockfordlhotka:
Also byte arrays. I ran into this last night, and changed my code to base64 encode the data, but it would be more efficient to just pass the binary byte array like on other platforms.

yowl:
Also posting binary data that is not from files, e.g. binary serializing DTOs as Json is not the most efficient for large datasets.

yowl:
And consuming    Content-Type: application/msbin1 web services would be helpful for large data transfer apps.

RemiBou:
- different model depending on response code (bad request won't send the same paylone as ok)

Don't know if it's in the scope of this issue

SteveSandersonMS:
@rockfordlhotka `byte[]` is now handled for both sending and receiving thanks to #815

gojanpaolo:
Hello, I would just like to ask if you someone could provide an example on how to use this newly added feature to upload files from the client and send it to the server? I'm completely new to web development so any help would be greatly appreciated. Thanks!

smartprogrammer93:
@gojanpaolo
An easy google search would lead you to this.

https://remibou.github.io/Upload-file-with-Blazor/
Try it out and let me know if you need help

gojanpaolo:
@smartprogrammer93 yes I found that. But it needs JS Interop. I thought this merged PR solved not having to use JS Interop during file upload?

smartprogrammer93:
@gojanpolo

[...]

PR #815, opened 2018-05-11
Add support for sending and receiving arbitrary HttpContent, refs #479 

Suchiman:
This adds support for sending arbitrary `HttpContent` by writing it to a `byte[]` first and then zero copy marshalling that as `Uint8Array` which is supported as request body.
It also supports receiving binary data by getting an `ArrayBuffer` from fetch, allocating a managed array of that size and then copying the response over.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
@@ -91,6 +91,12 @@ export const monoPlatform: Platform = {
     return mono_string(jsString);
   },
 
+  toUint8Array: function toUint8Array(array: System_Array<any>): Uint8Array {
+    const dataPtr = getArrayDataPointer(array);
+    const length = Module.getValue(dataPtr, 'i32');
+    return new Uint8Array(Module.HEAPU8.buffer, dataPtr + 4, length);
+  },
+
   getArrayLength: function getArrayLength(array: System_Array<any>): number {
     return Module.getValue(getArrayDataPointer(array), 'i32');
   },
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
@@ -1,29 +1,30 @@
 ﻿import { registerFunction } from '../Interop/RegisteredFunction';
 import { platform } from '../Environment';
-import { MethodHandle, System_String } from '../Platform/Platform';
+import { MethodHandle, System_String, System_Array } from '../Platform/Platform';
 const httpClientAssembly = 'Microsoft.AspNetCore.Blazor.Browser';
 const httpClientNamespace = `${httpClientAssembly}.Http`;
 const httpClientTypeName = 'BrowserHttpMessageHandler';
 const httpClientFullTypeName = `${httpClientNamespace}.${httpClientTypeName}`;
 let receiveResponseMethod: MethodHandle;
 
-registerFunction(`${httpClientFullTypeName}.Send`, (id: number, method: string, requestUri: string, body: string | null, headersJson: string | null, fetchArgs: RequestInit | null) => {
-  sendAsync(id, method, requestUri, body, headersJson, fetchArgs);
+registerFunction(`${httpClientFullTypeName}.Send`, (id: number, body: System_Array<any>, jsonFetchArgs: System_String) => {
+  sendAsync(id, body, jsonFetchArgs);
 });
 
-async function sendAsync(id: number, method: string, requestUri: string, body: string | null, headersJson: string | null, fetchArgs: RequestInit | null) {
+async function sendAsync(id: number, body: System_Array<any>, jsonFetchArgs: System_String) {
   let response: Response;
   let responseText: string;
 
-  const requestInit: RequestInit = fetchArgs || {};
-  requestInit.method = method;
-  requestInit.body = body || undefined;
+  const fetchOptions: FetchOptions = JSON.parse(platform.toJavaScriptString(jsonFetchArgs));
+  const requestInit: RequestInit = Object.assign(fetchOptions.requestInit, fetchOptions.requestInitOverrides);
+
+  if (body) {
+    requestInit.body = platform.toUint8Array(body);
+  }
 
   try {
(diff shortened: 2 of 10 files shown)


Suchiman:
It seems i've uncovered a bug where in the old implementation, `headersJson` contains
```json
[["Content-Type","text/plain; charset=utf-8"],["Content-Type","application/json"]]
``` but `content-type: application/json` gets actually sent.
In the new implementation, you get `Content-Type: text/plain; charset=utf-8, application/json` which breaks the `CanSendAndReceiveJson` unit test as the server now thinks he's getting `text/plain` and therefore responds with `415 Unsupported Media Type`.
As the new behavior seems to be more correct, i've adjusted the unit test.

**Comment by Suchiman** [comment]
Now i'm starting to get paranoid, adding support for receiving binary data, that didn't touch headers at all, made the response `Content-Type` header become lowercase, failing that same test again. I've made that Assert now compare case insensitive.

**Comment by Suchiman** [comment]
This is ready for review

File: src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts [hunk 1]
```diff
@@ -91,6 +91,12 @@ export const monoPlatform: Platform = {
     return mono_string(jsString);
   },
 
+  getByteArray: function getByteArray(array: System_Array<any>): Uint8Array {
```
SteveSandersonMS: For consistency with the other value-converting method (`toJavaScriptString`), could this be renamed to `toUint8Array`?

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
```diff
@@ -1,50 +1,76 @@
...
+    allocateArrayMethod = platform.findMethod(
+      httpClientAssembly,
+      httpClientNamespace,
+      httpClientTypeName,
+      'AllocateArray'
+    );
+  }
+
+  // allocate a managed byte[] of the right size
+  const dotNetArray = platform.callMethod(allocateArrayMethod, null, [platform.toDotNetString(responseData.byteLength.toString())]) as System_Array<any>;
```
SteveSandersonMS: Did you have to convert the length to a `string` because of a JS interop limitation? Fine if so. Just wondered whether you tried passing it directly as a number and receiving an `int` on the .NET side. Not sure whether that works yet.

SteveSandersonMS:
Thanks so much for this, @Suchiman! I'm impressed that you managed to figure out the details of the interop so well.

I've posted a few comments requesting updates, but hopefully we can include this in the 0.4.0 release if resolved soon :)

src/Microsoft.AspNetCore.Blazor.Browser/Http/BrowserHttpMessageHandler.cs
```diff
@@ -124,17 +126,34 @@ private static object GetDefaultCredentialsString()
             }
         }
 
-        // Keep in sync with TypeScript class in Http.ts
+        // Keep these in sync with TypeScript class in Http.ts
+        private class DotNetRequestInit
```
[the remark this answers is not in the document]
Suchiman: Sometimes the solution is easier than it seems, thanks

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
Suchiman: I've copied the trick from https://github.com/aspnet/Blazor/blob/dc532a3963a36597c2d4df907c9a686fd58266b7/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts#L71-L76
so it's prior art.

callMethod expects an `System_Object[]` but i suppose it would work in theory as it seems to allocates 32bit space on the stack for each argument, which seems a bit fishy https://github.com/aspnet/Blazor/blob/42e3848837be69a479944e55a8789ecc59a4f2cc/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts#L50-L64
Everything around `RenderTree*` friends is built with 64bit pointers in mind.

test/testapps/BasicTestApp/HttpClientTest/BinaryHttpRequestsComponent.cshtml
```diff
@@ -0,0 +1,67 @@
...
+@if (responseStatusCode.HasValue)
+{
+    <h2>Response</h2>
+    <p><div>Status:</div><span id="response-status">@responseStatusCode</span></p>
+}
+
+<span id="status-text">@statusText</span>
+
+@functions {
+    string uri = "http://localhost:46675/api/data";
```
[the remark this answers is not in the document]
Suchiman: I did orient myself at https://github.com/aspnet/Blazor/blob/42e3848837be69a479944e55a8789ecc59a4f2cc/test/testapps/BasicTestApp/HttpClientTest/HttpRequestsComponent.cshtml#L64 which also hardcodes an url which is then changed in the unit tests

samples/StandaloneApp/_ViewImports.cshtml
```diff
@@ -1,4 +1,6 @@
 ﻿@using System.Net.Http
+@using Microsoft.AspNetCore.Blazor
+@using Microsoft.AspNetCore.Blazor.Components
```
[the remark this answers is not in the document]
Suchiman: Weird, back then when i implemented it the project wouldn't compile without them but now it does 🤷‍♂️

Suchiman:
@SteveSandersonMS feedback incorporated and rebased onto dev.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
SteveSandersonMS: All the .NET code has to allow for running in x64 (because that will often be the case, e.g., in SSR or unit tests). But the JS-side code can assume 32-bit because the only time JS code can see the .NET memory directly is when running under WebAssembly, and the WebAssembly runtime is always 32-bit. Perhaps this will change in the future but it is the case presently.

SteveSandersonMS:
This is absolutely superb. Thanks so much, @Suchiman!

It's really great to be able to include this in 0.4.0. It's a valuable feature that a lot of people will benefit from, and the implementation looks ideal.

Issue #867, opened 2018-05-21
Ctrl/Shitf Click on internal application link doesn't open in a new browser tab/window


Andrzej-W:
Standard behaviour for Ctrl/Shift Click in Firefox/Chrome/Edge is to open link in a new browser tab or window. Unfortunately this does not work for internal links in Blazor application. Middle mouse click doesn't work either (it works like standard left click). Blazor also doesn't respect `target="_blank"` attribute for internal links.

I can right-click on internal link and select "Open in a new tab/window" from context menu - this does work.

**Why we need this behaviour?**
First it is needed for consistency. Second, in LOB applications people often have to see related data from different modules. To do this in the browser they can Ctrl Click on the link to open it in a new browser tab and then drag this tab to the second monitor - it is very convenient.

Suchiman:
On https://blazor-demo.github.io/ with Edge and `CTRL + Click` and `CTRL + Shift + Click` this does work for me. E.g. clicking on Counter opens a new page on https://blazor-demo.github.io/Counter

Andrzej-W:
You are right - in Edge Ctrl+Click works - a new browser **tab** is opened. Shift+Click doesn't work - link should be opened in a new browser **window**. Ctrl+Click and Shift+Click doesn't work as expected in Firefox and Chrome.

Suchiman:
Strange indeed, CTRL / Shift + Click does nothing on Firefox but middle mouse click does work for opening a new tab. Interestingly the only difference Shift makes for me, is that the new tab gets focus immediately vs. opening in the background.

Andrzej-W:
https://blazor-demo.github.io/ is based on Blazor 0.2.0. Middle click works here in all browsers. I have Blazor 0.3.0 on my computer and here middle click works in Edge and Chrome, but **doesn't work in Firefox**.
Firefox 60.0.1
Chrome 66.0.3359.181
Edge Microsoft EdgeHTML 16.16299, Microsoft Edge 41.16299.402.0

SteveSandersonMS:
Thanks for reporting it.

The fix should be a relatively simple tweak in `UriHelper.ts` around line 25 to check whether any `metaKey` flag or similar is present on the `event` and if so, skip the internal-link handling logic.

If anyone wants to submit a PR for it we'd be happy to take it!

RemiBou:
what's your expectation regarding tests for this PR ? E2E ?

PR #868, opened 2018-05-22
manages when users clicks on a link with the control button pushed. 

RemiBou:
- added E2E test for checking 2 windows opened after click with ctrl hold
- added E2E test for checking 1 window opened after click without ctrl hold

issue #867


RemiBou:
why is build failing on travis but not on appveyor ? Is there anything to know with selenium on macos ? Maybe I should wait until the windows is closed ?

SteveSandersonMS:
Thanks for this! I'll add some comments.

> why is build failing on travis but not on appveyor ? Is there anything to know with selenium on macos ? Maybe I should wait until the windows is closed ?

I don't know exactly why, but this is the first E2E test that tries to work with multiple windows, so maybe it is necessary to handle the cleanup code differently. The error message does look like a genuine issue rather than just a random Travis instability.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
```diff
@@ -22,7 +22,10 @@ registerFunction(`${registeredFunctionPrefix}.enableNavigationInterception`, ()
     const anchorTarget = findClosestAncestor(event.target as Element | null, 'A');
     if (anchorTarget) {
       const href = anchorTarget.getAttribute('href');
-      if (isWithinBaseUriSpace(toAbsoluteUri(href))) {
+        console.log(event);
+      if (isWithinBaseUriSpace(toAbsoluteUri(href)) && 
+          !event.ctrlKey)
```
SteveSandersonMS: I think it's not only ctrl that needs to be accounted for. There's also shift and meta, depending on the browser OS. Would you be able to determine the desired behavior for each such modifier key?

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/RoutingTest.cs
```diff
@@ -68,6 +69,40 @@ public void CanFollowLinkToOtherPage()
             Assert.Equal("This is another page.", app.FindElement(By.Id("test-info")).Text);
             AssertHighlightedLinks("Other", "Other with base-relative URL (matches all)");
         }
+        [Fact]
```
SteveSandersonMS: Incredibly minor point, but could you add a newline before each of the two new `[Fact]` methods? Thanks!

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
RemiBou: ok I'll try to find more details on this

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/RoutingTest.cs
```diff
@@ -68,6 +69,40 @@ public void CanFollowLinkToOtherPage()
...
+                .Click(button)
+                .Build()
+                .Perform();
+            
+            Assert.Equal(2, Browser.WindowHandles.Count);
+
+            Browser.SwitchTo().Window(Browser.WindowHandles.Last());
+            Browser.Close();
+            Browser.SwitchTo().Window(Browser.WindowHandles.First());
+            new Actions(Browser)
```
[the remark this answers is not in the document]
RemiBou: np

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/RoutingTest.cs
RemiBou: :)

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
```diff
@@ -22,7 +22,11 @@ registerFunction(`${registeredFunctionPrefix}.enableNavigationInterception`, ()
     const anchorTarget = findClosestAncestor(event.target as Element | null, 'A');
     if (anchorTarget) {
       const href = anchorTarget.getAttribute('href');
-      if (isWithinBaseUriSpace(toAbsoluteUri(href))) {
+      //if the user wants to user some specific browser/OS feature, we dont handle it and let the browser/OS 
+      var anyChangeBehaviorKeyHold = event.ctrlKey || event.shiftKey || event.altKey || event.metaKey;
+      if (isWithinBaseUriSpace(toAbsoluteUri(href)) && 
+          !anyChangeBehaviorKeyHold ) 
+      {
```
SteveSandersonMS: Really minor comments, but could you use `const` instead of `var` on line 26? Also lines 27 and 28 can easily go on the same line as they are short now.

SteveSandersonMS:
Thanks for the updates! Just added a couple more comments, plus hopefully an explanation for the Mac failure.

RemiBou:
can't find a way to make it working on mac os, 2 solutions :
- Keys.Command or Keys.Meta are not the good keys for mac os
- event.metaKey is not good on mac, then it's a  bit more complicated  (see https://stackoverflow.com/questions/3902635/how-does-one-capture-a-macs-command-key-via-javascript)

What do you think ? should I remove the test on mac, and we'll see later or add something eavier like on the SO link for managing the command key ?

RemiBou:
I search and it seems that angular and react do nearly the same thing 

https://github.com/angular/angular/blob/6c2c95851ae759cb149bc0868145f0b6611b61ee/aio/src/app/shared/location.service.ts

https://github.com/ReactTraining/react-router/blob/e6f9017c947b3ae49affa24cc320d0a86f765b55/packages/react-router-dom/modules/Link.js

I could add the _blank handling and other than left click ignore also. It doesn't explain why the E2E test on mac os fails though

RemiBou:
@SteveSandersonMS can you give me feedback here please ?

SteveSandersonMS:
Thanks for the update. We definitely need it to actually work on Mac. I just had a look, and `event.metaKey` did appear to correspond to the cmd key being held at the time of the click, at least in the version of Chrome I was using on macOS. Are you aware of there being specific browsers where this doesn't work?

RemiBou:
yes it passes ! it was just a stupid coding mistake. All this work for 50 LOC, E2E tests are depressing.

SteveSandersonMS:
Thanks very much @RemiBou! This is now merged.

RemiBou:
You're welcome, please add more tags like "good first issue" or "up for grabs" I'll help with great pleasure. keep the good work.

Issue #870, opened 2018-05-22
Blazor uses old version of .NET Core SDK


SIkebe:
Blazor 0.3.0 requires dotnet core SDK 2.1.300-preview2-008533 or later.
https://blogs.msdn.microsoft.com/webdev/2018/05/02/blazor-0-3-0-experimental-release-now-available/

However, Blazor itself uses 2.1.300-preview2-008367 to build.
https://github.com/aspnet/Blazor/blob/efc737b21492d5b74157b447256396d752a5d598/global.json#L1

Is there any reason to use prior version? Not RC1?

rynowak:
The ASP.NET Core repos will almost always use a prerelease SDK to build. This is how we flow arbitrary builds of the SDK and tools through our build system. Even is we did update this repo to target a specific version, that would be a temporary situation and we'd end up with a prerelease again very soon. 

Since Blazor is part of the ASP.NET Core project, we don't want to deviate from the tools and practices that the shipping repos use. Let us know if this is causing you problems in any specific way. 

You're right that in this case we haven't updated this repo in a while, I'll do that.

SIkebe:
Thanks for clarifying!

rynowak:
I don't really understand the question. What are you trying to do and why?

rynowak:
Closing this because it's updated and I've answered the original question. @ncarandini if you still have something you want to discuss relevant to Blazor feel free to open an issue or reply here.

ncarandini:
@rynowak, sorry, I'll try to clarify:
Using VS 2017 Preview (15.8.0 Preview 1.1) and CI Blazor (version 15.7.10293), after creating a new Blazor app the project will contain a `global.json` file that contain a reference to .NET Core SDK version `2.1.300-preview2-008533` that is not the latest, and you have well explained why.

If I delete the `global.json` file, the project will automagically use the latest installed (in my case, the version `2.1.300-rc1-008673`).

My question is: is this a good practice?

SIkebe:
@ncarandini It depends. IMHO, you might want to have your `global.json` and pin the SDK version to avoid unexpected behavior in future.

rynowak:
It's not in general a *good practice* to delete the global.json - if you're the only one working on your codebase then it probably won't matter what you do since you control your environment. The tooling about SDK versions is going to improve during 15.8 - and I hope we'll eventually be at a place where VS tells you if you have an older or mismatched SDK than that the global.json indicates.

[...]

PR #872, opened 2018-05-22
Update Korebuild

rynowak:
Related to #870

### DIFF
--- a/korebuild.json
+++ b/korebuild.json
@@ -1,11 +1,11 @@
 {
   "$schema": "https://raw.githubusercontent.com/aspnet/BuildTools/dev/tools/korebuild.schema.json",
-  "channel": "dev",
+  "channel": "release/2.1",
   "toolsets": {
     "visualstudio": {
       "required": false,
       "includePrerelease": true,
-      "minVersion": "15.0.26730.03",
+      "minVersion": "[15.0.26730.03, 15.9)",
       "requiredWorkloads": [
         "Microsoft.VisualStudio.Component.VSSDK"
       ]
--- a/build/VSIX.targets
+++ b/build/VSIX.targets
@@ -63,7 +63,6 @@
       <MSBuildArguments Include="
         $(VSIXProject);
         /t:Restore;
-        /m;
         /v:m;
         /p:Configuration=$(Configuration);
         /p:BuildNumber=$(BuildNumber);" />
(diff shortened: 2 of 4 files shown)


korebuild.json
```diff
@@ -1,11 +1,11 @@
 {
   "$schema": "https://raw.githubusercontent.com/aspnet/BuildTools/dev/tools/korebuild.schema.json",
-  "channel": "dev",
+  "channel": "release/2.1",
   "toolsets": {
     "visualstudio": {
       "required": false,
       "includePrerelease": true,
-      "minVersion": "15.0.26730.03",
+      "minVersion": "[15.0.26730.03, 15.8)",
```
rynowak: We need to put an upper bound on which versions of MSBuild we use to build the VSIX. The current preview of VS isn't compatible with this SDK for building **BUILDING THE VSIX AT THE COMMANDLINE ONLY**.

I'm going to check on this before merging this PR and see if there's another option. This is the setup that razor is using currently.

korebuild.json
```diff
@@ -1,11 +1,11 @@
 {
   "$schema": "https://raw.githubusercontent.com/aspnet/BuildTools/dev/tools/korebuild.schema.json",
-  "channel": "dev",
+  "channel": "release/2.1",
```
rynowak: Using the release version of the tools, which is going to track 2.1 and its patches. If we retarget at later builds, we can update then. I figured we don't want to use 2.2 yet.

NTaylorMullen:
> Blazor is using an ancient version of Korebuild, but the new ones in the release channel are incompatible with desktop MSBuild in 15.8p1. Ideas?

When you say incompatible what error are you getting?

rynowak:
```
C:\Users\Ryan\.dotnet\x64\sdk\2.1.300-rtm-008823\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.ta
rgets(199,7): error MSB4064: The "ProjectAssetsCacheFile" parameter is not supported by the "ResolvePackageAssets" task.
 Verify the parameter exists on the task, and it is a settable public instance property. [F:\git\aspnet\Blazor\tooling\M
icrosoft.VisualStudio.LanguageServices.Blazor\Microsoft.VisualStudio.LanguageServices.Blazor.csproj] [C:\Users\Ryan\.dot
net\buildtools\korebuild\2.1.0-rtm-15789\KoreBuild.proj]
```

NTaylorMullen:
@rynowak remove the `/m;` from the msbuild item group in both locations in your VSIX.targets. Like in [this](https://github.com/aspnet/Razor/commit/63af9221dc147c4931449e6ae37f27f39d314f1b#diff-ac80d3760c2af1ba56c6c1aa46d74fa4) commit.

rynowak:
Awesome! That did the trick.

PR #918, opened 2018-06-02
Fix base type of UIErrorEventArgs

danroth27:
I don't believe `UIErrorEventArgs` should derive from `UIProgressEventArgs`. Switching base type to `UIEventArgs` instead.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
@@ -248,7 +248,7 @@ interface UIDataTransferItem {
   type: string;
 }
 
-interface UIErrorEventArgs extends UIProgressEventArgs {
+interface UIErrorEventArgs extends UIEventArgs {
 }
 
 interface UIFocusEventArgs extends UIEventArgs {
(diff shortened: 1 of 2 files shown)


galvesribeiro:
Yeah that was a mistake, sorry about that. Just confirmed https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

Thanks for the fix!

galvesribeiro:
In fact, I don't remember why I didn't added the `message` field at least there...

src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs
```diff
@@ -168,7 +168,7 @@ public class UIDataTransferItem
     /// <summary>
     /// Supplies information about an error event that is being raised.
     /// </summary>
-    public class UIErrorEventArgs : UIProgressEventArgs
```
rynowak: This should have some properties https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
```diff
@@ -248,7 +248,7 @@ interface UIDataTransferItem {
   type: string;
 }
 
-interface UIErrorEventArgs extends UIProgressEventArgs {
```
rynowak: https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

rynowak:
@SteveSandersonMS - quick review please 😆

SteveSandersonMS:
@rynowak It looks like we're still not using the `UIErrorEventArgs` type. Shouldn't we add something under line 55 in `EventForDotNet.ts` to make it parse the event object as a `UIErrorEventArgs`?

galvesribeiro:
It should have an even handler...

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

rynowak:
> It should have an even handler...

Can you explain that in more detail? I'm not sure what that means

rynowak:
@SteveSandersonMS updated

PR #931, opened 2018-06-05
Refactoring SimpleJson.EatWhiteSpace

toddlang:
-Replacing arbitrary length IndexOf call with unrolled switch statement
-Reduces deserialization runtime on ~310KB test data from 13s to 5s.


src/Microsoft.AspNetCore.Blazor/Json/SimpleJson/SimpleJson.cs
```diff
@@ -1,4 +1,4 @@
-﻿//-----------------------------------------------------------------------
+//-----------------------------------------------------------------------
```
toddlang: I can't seem to make it stop thinking this line has changed.

SteveSandersonMS:
Thanks very much @toddlang!

PR #933, opened 2018-06-05
Add type checking for component parameters

rynowak:
Introduces a new primitive used by the compiler for type checking. Type
checking applies to component parameters when setting the value directly
and when using bind. This is nice because it also adds error checking
for bind.


rynowak:
> What do you think?

I'd like to see some data before spending time on microoptimizations 😆. Given that we don't really have a way to collect data right now that's impossible. 

I don't see why we can't trust the compiler in this case

PR #937, opened 2018-06-06
Don't preventDefault on events. Fixes #803

### LINKED ISSUE #803: Keydown event on input gets cancelled
Created a blazor project from default template. Then modifed the index.cshtml as the below code. Then type something in the input box. The browser console prints the pressed key, but the keyboard event seems cancelled. Nothing gets printed in the input field.
Also the `arg.Key` property gives me a string. Can I get a integer specifying as to which key was pressed (27 for escape key for example)

```
@page "/"
<input onkeydown="@HandleKeyDown" />
@functions{
    void HandleKeyDown(UIKeyboardEventArgs arg) {
        Console.WriteLine(arg.Key);
    }
}

### PR DESCRIPTION
**Pull request by SteveSandersonMS**
It seems that we started calling `event.preventDefault` on all .NET-handled events in this commit: https://github.com/aspnet/Blazor/commit/36f5c5898820899831447b3dbde5e8658cc0baaa

It's not explicitly stated why I added that line in that commit, but my guess now is that it was needed because we didn't yet have the "internal link click" handling logic that we now do (which does call `preventDefault` when it recognizes clicks on internal links).

So it looks like we can fix #803 and all related issues trivially just be removing the `event.preventDefault` call. It doesn't appear to affect anything else negatively, and conceptually should be the right thing to do because saying you want to handle an event is not the same as saying you want to prevent its default actions.

### DIFF
--- a/test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/ComponentRenderingTest.cs
+++ b/test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/ComponentRenderingTest.cs
@@ -100,6 +100,9 @@ namespace Microsoft.AspNetCore.Blazor.E2ETest.Tests
             Assert.Collection(liElements(),
                 li => Assert.Equal("a", li.Text),
                 li => Assert.Equal("b", li.Text));
+
+            // Textbox contains typed text
+            Assert.Equal("ab", inputElement.GetAttribute("value"));
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/ComponentRenderingTest.cs [hunk 1]
```diff
@@ -100,6 +100,9 @@ public void CanTriggerKeyPressEvents()
             Assert.Collection(liElements(),
                 li => Assert.Equal("a", li.Text),
                 li => Assert.Equal("b", li.Text));
+
+            // Textbox contains typed text
+            Assert.Equal("ab", inputElement.GetAttribute("value"));
```
SteveSandersonMS: Note that this assertion would fail if the `event.preventDefault` call was still present.

SteveSandersonMS:
Merging now so I can proceed to make an RC build, but CC @rynowak @javiercn in case either of you can think of any reason why the change in this PR is not the right thing to do.

rynowak:
Looks straightforward to me.

schotime:
Wouldn't this cause problems for things like onsubmit handlers for forms where it will also submit the form or for a tags follow the hrefs if there?

RemiBou:
@SteveSandersonMS I observed the problem raised by @schotime since 0.4

PR #938, opened 2018-06-06
Update Steve Sanderson's demo to latest NDC talk

mattferderer:
Steve recently did another NDC talk which is much more up to date. Suggesting that one be linked to instead of the older talk.


danroth27:
Thanks!

PR #944, opened 2018-06-07
When clicking <a> tag with no href, don't attempt navigation to "null". Fixes #943

### LINKED ISSUE #943: How to preventdefault in link onclick event?
After https://github.com/aspnet/Blazor/commit/93c1e37c530cfa0e724ce073df83d16168261c90 how can we prevent default action on link?

I have a pagination 
`                    @foreach (var exerc in Exercises)
                    {
                        <li class="page-item @( exerc.isSelected ? "active" : "")"><a class="page-link" onclick="@((__value) => { ChangeCurrentExercise(exerc.Id); })">@count</a></li>
                        count++;
                    }`
and after onclick it tries to route to null route.

SteveSandersonMS:
This PR will merge into `release/0.4.0` (and afterwards, I'll cherry-pick to `dev`).

The recent change to stop calling `preventDefault` on all events has caused a regression if you click on an `<a>` element that has no `href`. Our "internal link navigation" logic will try to read the `href`, gets the value `null`, and then because that's not anticipated it ends up being treated as string string `"null"` and we attempt to navigate to that as if it's a URL.

The fix is simply to do nothing if there's no `href` attribute. There is nothing we want to do in that situation.

**Broader note**: This is an example of how the removal of `preventDefault` might be breaking in a bunch of unanticipated ways. However I still think it's the right thing to do (both by default, and for the great majority of cases). When we hear of scenarios where it's necessary to preventDefault, we'll have to consider each of them separately, as well as consider adding a general mechanism for controlling whether events are preventDefaulted.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
@@ -19,9 +19,11 @@ registerFunction(`${registeredFunctionPrefix}.enableNavigationInterception`, () 
 
   document.addEventListener('click', event => {
     // Intercept clicks on all <a> elements where the href is within the <base href> URI space
-    const anchorTarget = findClosestAncestor(event.target as Element | null, 'A');
-    if (anchorTarget) {
-      const href = anchorTarget.getAttribute('href');
+    // We must explicitly check if it has an 'href' attribute, because if it doesn't, the result might be null or an empty string depending on the browser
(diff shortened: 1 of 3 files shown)


conficient:
Has this fix been lost somehow? If I create a `<a onclick="@myFn">text</a>` then `href='#'` is inserted and it navigates rather than handling the click event?

I note the related issue on aspnetcore: https://github.com/aspnet/AspNetCore/issues/5545

conficient:
Ah, found a work-around:
```
<a href='javascript:null' onclick="@myFn">text</a>
```
This makes the link enabled but doesn't navigate, and `onclick` is handled

PR #973, opened 2018-06-11
Add para [Parameter] Code Snippet

vertonghenb:
Should complete the idea of #965
Adds support for auto complete of para > tab > tab
```csharp
[Parameter] protected int MyParameter {get;set;}
````
just like prop > tab > tab
```csharp
public int MyProperty { get; set; }
````
A new VSIX project was created to separate the snippets from the templates. However I'm not sure if this is the way to go and or if we still need some build steps?


rynowak:
It would be preferrable to keep this in a single VSIX. Could you try adding the pkfdef to the existing VSIX? We already do this for the templates.

vertonghenb:
Reverted and added to the existing vsix. 
@rynowak  is the namespace of the snippet ok? (Blazor) or rather "Visual C#"?

vertonghenb:
Just gonna clean-up the PR (new one)

Issue #966, opened 2018-06-09
Built-in Json isn't compatible with the default serialization in ASP.NET


ericwj:
I found that JSON coming from an API controller that uses Entity Framework Core cannot handle the default serialization done by ASP.NET in a few cases.

* `sbyte`, `ushort`, `uint` and `ulong` do not deserialize (ArgumentException)
* `byte[]` content for e.g. `[Timestamp]` properties which are serialized as base64 do not deserialize
* `decimals` outside the range of `double` cq `long` do not deserialize

The first needs a single line changed, the second needs 2 added lines, the third needs two if else statements added.

Would you like me to submit a pull request to fix this?

SteveSandersonMS:
Thanks for letting us know @ericwj!

Longer term our intention is not to have the built-in JSON utilities, but rather to make use of an external package to handle JSON (de)serialization. We only have the built-in JSON utility as a short-term workaround for the lack of an ideal external package, but as soon as one becomes available we'll switch to it.

So if you're interested and you can keep the PR simple and minimal enough, yes please, we'd be happy to include these enhancements. Just bear in mind that it won't be in here forever as we'll be removing all the JSON code as soon as we can. Hope that's OK!

ericwj:
Yes I found Json.NET doesn't work at all. So I ripped your copy of SimpleJson and put it in my project. Then made the changes I needed. This is just to contribute them back.

PR #977, opened 2018-06-11
Support the full range of all basic types. Deserialize byte[] from base64.

ericwj:


ericwj:
#966

SteveSandersonMS:
Thanks @ericwj for contributing this! I appreciate that you've been thorough with this PR.

I wasn't expecting it to involve such extensive code changes. In #966, you mentioned:

> The first needs a single line changed, the second needs 2 added lines, the third needs two if else statements added.

Since we are specifically *not* trying to build and maintain an all-encompassing JSON library inside Blazor, we'd generally favour not adding or changing complex behaviour. Rather, if a change this sophisticated is needed, it would be better to publish it as a separate package that developers can use if they need.

Would you be interested in publishing your own fork of the JSON utils as a separate package?

Or is there any way you can reduce the scope of the changes just to the few very small tweaks originally mentioned in #966? If it really was just a couple of tiny changes we'd be happy to merge it. Of course, I understand that would mean covering a smaller range of functionality, which is explicitly our goal for the built-in JSON util which will be removed at some point anyway when superseded.

Suchiman:
If it's going to be removed anyway, whats the problem with reducing friction in the mean time?

SteveSandersonMS:
@Suchiman Just that it's taking away time from the real project goals. Every change to the JSON code involves careful review and may introduce new issues. We think we can provide better value to the community by focusing elsewhere.

@ericwj How would you feel about publishing a separate JSON util package? We'd definitely point people to it whenever they have more sophisticated requirements than the built-in package handles. If the community is keen to develop this further it would be great to decouple it.

Andrzej-W:
Decimal data type is crucial in any LOB application. I understand that Blazor is an experimental framework now and we will see a lot of changes in the future. Some of us want to test Blazor in real word scenarios and it is inconvenient when we don't have full support for basic data types. Please merge this commit.

ericwj:
I would probably revert this to the three lines. Maybe I can make the promised two if statements (the changes to ParseNumber) somewhat simpler either by just trying decimal or ulong first and falling back to long/double if success is false. That would (further) reduce performance because it would change the default type returned for all floating point numbers to decimal.

SteveSandersonMS:
@ericwj If that's your preference, that sounds OK.

Personally, I suspect it might be time for a community-led JSON package, presumably based on a fork of the SimpleJson code in this repo. If there are places where Blazor's API makes it hard or impossible to use an external JSON serializer, we would prioritise extending the APIs to make that easier. This work is needed in the long term anyway, since using third-party JSON libraries will always be a scenario.

ericwj:
Most people will default to getting Json.NET and find that it crashes however simple an object you try to deserialize.

Currently its probably the runtime that has some quirks causing this - if it weren't the case I wouldn't consider making these changes but use Json.NET (although its big and complex).

Besides that - if this isn't supposed to be consumed in the default case, then it shouldn't surface through the extensions methods on HttpClient or in any other place by default. It'd probably have to be internal.

But, just as ASP.NET includes Json.NET, it'd be my preference if Blazor would have something so basic inbox.

ericwj:
Its probably easiest to create a new pull request @SteveSandersonMS?

Would you like to have a [look at it](https://github.com/ericwj/Blazor/compare/feature/simplejson-min?expand=1) before I submit it?

If this isn't simple enough, the alternative of literally three lines will only add support for sbyte, ushort, uint and ulong but fail for any integers outside the range of long (half the range of ulong will be supported) and loose precision if the 54 bits of double aren't enough.

Daddoon:
Hello @ericwj,

For Json.net you may take a look and some test at my fork if you find it usefull:

[Json.NET.Aot](https://github.com/Daddoon/Json.NET.Aot)

It's based on SaladLab Json.Net 9.0.1 Aot compatibility library, initially for Unity3D, but this version is upgraded to .Net standard 2.0. I use it on UWP/Android/iOS (Xamarin) and Blazor.

SteveSandersonMS:
@owaits You can use a more powerful JSON library such as Json.NET. Just note that you need to configure the linker not to strip out `System.Linq.Expressions*` ([see docs](https://github.com/aspnet/Blazor.Docs/blob/master/docs/host-and-deploy/configure-linker.md)). This is a temporary issue that should be resolved in Blazor 0.7.0.

Longer term, we intend to swap out the default JSON library with a more powerful one.

danroth27:
At this point we think we'd like to take a different approach with the implementation. We're investigating using a different JSON implementation, so we'd prefer not to take more involved changes to the existing SimpleJson code base.  As such, I'm closing this issue.

Suchiman:
@danroth27 btw. is this expected to be the new JSON API comming to .NET Core 3?

danroth27:
That's the hope: https://github.com/dotnet/announcements/issues/90

Issue #982, opened 2018-06-12
Right clicking local anchor tags automatically navigates you instead of display a context menu


Worble:
### Title
Right clicking local anchor tags automatically navigates you instead of display a context menu

### Functional impact
Users cannot use any of the context menu items when right clicking anchor tags

### Minimal repro steps
Given that you have 2 pages defined, say "/" and "/home", right clicking this `<a href="/home">Home<a/>` on the "/" page will cause you to instantly navigate to home instead of displaying the usual browser context menu.

### Expected result
Viewing of the browser context menu

### Actual result
The user is navigated to the anchor as if they left clicked it

### Further technical details
 This does not happen for links outside of the scope of the router, for example `<a href=www.google.com>Google</a>` or `<a href=/images/image.jpg>Image</a>`

reaction1989:
@SteveSandersonMS i would look at this. Is there a way at the moment to debug the code in "Microsoft.AspNetCore.Blazor"?

chucker:
>Is there a way at the moment to debug the code in "Microsoft.AspNetCore.Blazor"?

Basically, no. See #14

SteveSandersonMS:
The bug in this case is in the TypeScript code which can be debugged using the browser’s built-in debugger. It will be a trivial fix.

PR #1010, opened 2018-06-15
Fix right mouse click on <a href""> in firefox

reaction1989:
#982


Suchiman:
Should this perhaps use [button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) instead of [which](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/which) which appears to be non standard?

Andrzej-W:
@reaction1989 In all web browsers I have tested middle click (mouse wheel click) opens link in a new tab. Unfortunately in Blazor middle click on internal link works exactly the same as left click. I think that your small change will fix this problem also. Thanks!

SteveSandersonMS:
Thanks, @reaction1989 and @Suchiman! This is now merged.

RemiBou:
maybe we could've updated eventHasSpecialKey with the new check "|| event.bouton === 0"

RemiBou:
eventHasSpecialKey is here for handling those cases when we don't want to override the browser behavior (ctrl click, shift click). The check button === 0 could be in this method as it's the same goal. Is that clear enough ?

Andrzej-W:
In my humble opinion eventHasSpecialKey is a utility function. Similarly isWithinBaseUriSpace is also utility function. Both functions together + outer `if` are used to decide if we want to perform internal navigation or not.

RemiBou:
fair enough

Issue #959, opened 2018-06-08
Out-of-process execution


SteveSandersonMS:
Blazor's architecture creates opportunities for other ways of hosting and running Blazor applications, including:

* In Electron
* In a Web Worker
* With server-side .NET code

For all three of these, we need a method for serializing the .NET/JS interactions so they can be pushed over a pipe (IPC, postMessage, WebSocket, etc.).

We'll break down this issue further as the design progresses.

Remaining work:
* [x] Get E2E tests working
* [ ] Edge support (#1131)
* [x] Templates

galvesribeiro:
@SteveSandersonMS Electron doesn't require any special treatment in this case. The standalone app can work just fine.

The only thing I think needs some love but would be potentially be an external package (Perhaps I would add it on BlazorExtensions) is the Eletron Main <-> Render process APIs. In other words, the _Electron SDK/APIs_ but for that to work, there must be a way to JS -> C# calls on non-static C# methods...

Do you think we need something else for Electron support?

SteveSandersonMS:
@galvesribeiro It's so the .NET part can run natively under CoreCLR instead of WebAssembly (so you immediately have access to 100% of the package ecosystem, debugging, JIT compilation, etc.).

galvesribeiro:
>  It's so the .NET part can run natively under CoreCLR instead of WebAssembly (so you immediately have access to 100% of the package ecosystem, debugging, JIT compilation, etc.).

Sorry, I'm not following... What you mean?

SteveSandersonMS:
I mean you can then have a normal native .NET Core process (not under Mono WebAssembly, instead just the native .NET Core runtime, a.k.a. CoreCLR) which controls an Electron renderer.

galvesribeiro:
Ahhh I see...

Well, perhaps `Blazelectron` is a good project start with! 💃

RichiCoder1:
You found this too: https://github.com/mattkol/Chromely

Paperthin render process that communicates with an out-of-process .net core process. Sound familiar :)?

stavroskasidis:
I love the `Blelectron` (or `Blazelectron` heh) idea. It provides us a solution for .Net/Cross-platform/Desktop apps (finally !!!). 

[...]

PR #1033, opened 2018-06-22
JavaScript interop v3

SteveSandersonMS:
This is one of the big pieces needed for #959. It changes how the JS interop works so it's factored out of the Blazor codebase itself, and makes "async" calls the mainstream lowest-common-denominator case (with sync and unmarshalled calls still available via certain casts if you really need that).

It's a monstrously big PR, but it does all have to happen in one go because once you change the underpinnings there are so many consequences. @rynowak, we can discuss whether or how you want to review this.

I'll write up a list of changes (especially the breaking API changes) in here later. For now I just want the CI run to start so am submitting this.


src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
```diff
@@ -254,3 +250,30 @@ function asyncLoad(url, onload, onerror) {
...
 }
+
+function attachInteropInvoker() {
+  const dotNetDispatcherInvokeMethodHandle = findMethod('Microsoft.JSInterop', 'Microsoft.JSInterop', 'DotNetDispatcher', 'Invoke');
+  const dotNetDispatcherBeginInvokeMethodHandle = findMethod('Microsoft.JSInterop', 'Microsoft.JSInterop', 'DotNetDispatcher', 'BeginInvoke');
+
+  DotNet.attachDispatcher({
+    beginInvokeDotNetFromJS: (callId, assemblyName, methodIdentifier, argsJson) => {
+      monoPlatform.callMethod(dotNetDispatcherBeginInvokeMethodHandle, null, [
+        callId ? monoPlatform.toDotNetString(callId.toString()) : null, // TODO: Implement a way to pass a number to a long without stringification
```
SteveSandersonMS: Am still planning to do this "TODO" item.

test/testapps/TestContentPackage/MyPrompt.cs
```diff
@@ -1,16 +1,14 @@
...
-        // Keep in sync with the identifier in the .js file
-        const string ShowPromptIdentifier = "TestContentPackage.showPrompt";
-
-        public static string Show(string message)
+        public static Task<string> Show(string message)
         {
-            return RegisteredFunction.Invoke<string>(
-                ShowPromptIdentifier,
+            return JSRuntime.Current.InvokeAsync<string>(
+                "TestContentPackage.showPrompt", // Keep in sync with identifiers in the.js file
```
SteveSandersonMS: After we merge this PR to `dev` and there is a new build on the dev feed, we need to remember to go back to the BlazorLib project template and make the equivalent change to this. Can't do so until there is a new build on the dev feed with the new APIs in it.

src/Microsoft.AspNetCore.Blazor.Browser.JS/package.json
```diff
@@ -4,13 +4,14 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "webpack",
+    "build": "webpack --mode development",
```
rynowak: Assuming this is temporary 😆

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/GlobalExports.ts
```diff
@@ -1,16 +1,20 @@
...
     navigateTo,
-    invokeDotNetMethod,
-    invokeDotNetMethodAsync
+
+    _internal: {
+      attachRootComponentToElement,
+      renderBatch,
+      http: httpInternalFunctions,
+      uriHelper: uriHelperInternalFunctions
+    }
```
rynowak: I can see that you're doing a relayer on some of the internals of `blazor.js` - I'm mostly ignoring these since I'm not sure I would have any meaningful feedback.

rynowak: So as I understand it - you're making everything was a registered function into a function that's reachable on some path from the root (`window`) since that's the new mechanism we're using to bind functions.

src/Microsoft.AspNetCore.Blazor.Browser/Interop/ArgumentList.cs
```diff
@@ -1,260 +0,0 @@
...
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-using System.Collections.Concurrent;
-using System.Linq;
-using System.Reflection;
-
-namespace Microsoft.AspNetCore.Blazor.Browser.Interop
-{
-    internal class ArgumentList
```
rynowak: RIP

src/Microsoft.AspNetCore.Blazor/Microsoft.AspNetCore.Blazor.csproj
```diff
@@ -1,7 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.JSInterop\Microsoft.JSInterop.csproj" />
+  </ItemGroup>
```
rynowak: It looks like the only reason for this reference is that ElementRef does custom serialization. Is that the case? Is this worth thinking harder about? (probably not) I can't imagine anyone will build a Blazor application without the interop layer.

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
```diff
@@ -0,0 +1,234 @@
...
+   */
+  export function attachReviver(reviver: JsonReviver) {
+    jsonRevivers.push(reviver);
+  }
+
+  /**
+   * Invokes the specified .NET public method synchronously. Not all hosting scenarios support
+   * synchronous invocation, so if possible use invokeMethodAsync instead.
+   *
+   * @param assemblyName The name of the .NET assembly containing the method.
```
rynowak: Do you think it's worth specifying in the docs that its the short assembly name (without key and version)? If you're coming from a .NET background you might assume that it's the full name

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
```diff
@@ -0,0 +1,234 @@
...
+     * @param resultOrExceptionMessage Either the operation result or an error message.
+     */
+    endInvokeDotNetFromJS: (asyncCallId: string, success: boolean, resultOrExceptionMessage: any): void => {
+      const resultOrError = success ? resultOrExceptionMessage : new Error(resultOrExceptionMessage);
+      completePendingCall(parseInt(asyncCallId), success, resultOrError);
+    }
+  }
+
+  function parseJsonWithRevivers(json: string): any {
+    return json ? JSON.parse(json, (key, initialValue) => {
```
rynowak: Interesting... I wasn't aware this feature exists 👍

galvesribeiro: Yeah, typescript language features are amazing... Each version it surprises me more! 👍

src/mono/driver_blazor.c
```diff
@@ -1,65 +1,77 @@
 static void*
-blazor_invoke_js (MonoString **exceptionMessage, MonoString *funcName, void* arg0, void* arg1, void* arg2)
+mono_wasm_invoke_js_marshalled (MonoString **exceptionMessage, void *asyncHandleLongPtr, MonoString *funcName, MonoString *argsJson)
 {
 	*exceptionMessage = NULL;
 	char *funcNameUtf8 = mono_string_to_utf8 (funcName);
-	void *jsCallResult = (void *)EM_ASM_INT ({
+	char *argsJsonUtf8 = argsJson == NULL ? NULL : mono_string_to_utf8 (argsJson);
+
+	MonoString *resultJsonMonoString = (void *)EM_ASM_INT ({
```
rynowak: So I am assuming that this block is 'inline assembly' in the form of JS.

S P I C Y

test/testapps/BasicTestApp/InteropTest/JavaScriptInterop.cs
```diff
@@ -12,42 +13,41 @@ public class JavaScriptInterop
...
-                state =>
-                {
-                    tcs.SetException(new InvalidOperationException("Threw an async exception!"));
-                },
-                null,
-                3000,
-                Timeout.Infinite);
-
-            return tcs.Task;
+            await Task.Delay(300);
```
rynowak: You can also use `Task.Yield()` here if you want to avoid a slowing down the test significantly. That's the documented way to actually force asynchrony

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
rynowak: Well I this case I think it's browser APIs - but yes, Typescript is amazing. 😻

rynowak:
Ok. So you really inflated the number of lines changed by moving mono.asm.js. T

hese are the kinds of shenanigans that put @pranavkm in first place in the MVC repo https://github.com/aspnet/Mvc/graphs/contributors - I don't approve #salty

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/ElementReferenceCapture.ts
```diff
@@ -1,12 +1,22 @@
...
   return `_bl_${referenceCaptureId}`;
-}
\ No newline at end of file
+}
+
+// Support receiving ElementRef instances as args in interop calls
+const elementRefKey = '_blazorElementRef'; // Keep in sync with ElementRef.cs
+DotNet.attachReviver((key, value) => {
+  if (value && typeof value === 'object' && value.hasOwnProperty(elementRefKey) && typeof value[elementRefKey] === 'number') {
+    return getElementByCaptureId(value[elementRefKey]);
```
rynowak: Should this be looking at the `key`? here? I suppose it won't be a problem because the top-value object is always an array?

src/Microsoft.JSInterop/DotNetDispatcher.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            var syncResult = InvokeSynchronously(assemblyName, methodIdentifier, argsJson);
+
+            // If there was no callId, the caller does not want to be notified about the result
+            if (callId != null)
+            {
+                // Invoke and coerce the result to a Task so the caller can use the same async API
+                // for both synchronous and asynchronous methods
+                var task = syncResult is Task syncResultTask ? syncResultTask : Task.FromResult(syncResult);
+                task.ContinueWith(completedTask =>
```
rynowak: Do you anticipate any problems if the continuation executes synchronously? I believe the existing callers of this method currently behave correctly

src/Microsoft.JSInterop/DotNetDispatcher.cs
```diff
@@ -0,0 +1,181 @@
...
+
+                    try
+                    {
+                        var result = TaskGenericsUtil.GetTaskResult(completedTask);
+                        jsRuntimeBaseInstance.EndInvokeDotNet(callId, true, result);
+                    }
+                    catch (Exception ex)
+                    {
+                        ex = UnwrapException(ex);
+                        jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, ex.ToString());
```
rynowak: Is there any use case here for passing the exception object? It seems little odd to me that this layer would decide how you want to treat exceptions.

src/Microsoft.JSInterop/IJSInProcessRuntime.cs
```diff
@@ -0,0 +1,20 @@
...
+    public interface IJSInProcessRuntime : IJSRuntime
+    {
+        /// <summary>
+        /// Invokes the specified JavaScript function synchronously.
+        /// </summary>
+        /// <typeparam name="T">The JSON-serializable return type.</typeparam>
+        /// <param name="identifier">An identifier for the function to invoke. For example, the value <code>"someScope.someFunction"</code> will invoke the function <code>window.someScope.someFunction</code>.</param>
+        /// <param name="args">JSON-serializable arguments.</param>
+        /// <returns>An instance of <typeparamref name="T"/> obtained by JSON-deserializing the return value.</returns>
+        T Invoke<T>(string identifier, params object[] args);
```
rynowak: Would about void-returning methods? Is the expectation that you'd use `object` as the type?

rynowak: or (`System.Void`)

src/Microsoft.JSInterop/JSRuntimeBase.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Invokes the specified JavaScript function asynchronously.
+        /// </summary>
+        /// <typeparam name="T">The JSON-serializable return type.</typeparam>
+        /// <param name="identifier">An identifier for the function to invoke. For example, the value <code>"someScope.someFunction"</code> will invoke the function <code>window.someScope.someFunction</code>.</param>
+        /// <param name="args">JSON-serializable arguments.</param>
+        /// <returns>An instance of <typeparamref name="T"/> obtained by JSON-deserializing the return value.</returns>
+        public Task<T> InvokeAsync<T>(string identifier, params object[] args)
+        {
+            // TODO: Auto-timeout
```
rynowak: I'd suggest against this if you don't have a clear need in mind.  

There are lots of cases where you don't want to have a timeout - "I want to be notified if X happens, but it might never happen".

If you **want** a timeout, then you want it to be cooperative - in which case you'd want to send it to the JS side as an actual argument - but I don't believe promises have a concept like cancellation - so again, you're building your own protocol. 

If you want a timeout but you're not going to make it cooperative (client cancellation) then what you'd want to cancel is any continuations you have scheduled, which is pretty easy to do in .NET.

src/Microsoft.JSInterop/JSRuntimeBase.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        /// <summary>
+        /// Begins an asynchronous function invocation.
+        /// </summary>
+        /// <param name="asyncHandle">The identifier for the function invocation, or zero if no async callback is required.</param>
+        /// <param name="identifier">The identifier for the function to invoke.</param>
+        /// <param name="argsJson">A JSON representation of the arguments.</param>
+        protected abstract void BeginInvokeJS(long asyncHandle, string identifier, string argsJson);
+
+        internal void EndInvokeDotNet(string callId, bool success, object resultOrExceptionMessage)
```
rynowak: With these begin-end pattern things, it's possible to have a bug where begin/end are called on different runtime instances - (if anything changes the instance of the runtime during execution). I don't think it's likely that anyone will do that - but just in case.

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
```diff
@@ -0,0 +1,234 @@
+// This is a single-file self-contained module to avoid the need for a Webpack build
+
+module DotNet {
+  (window as any).DotNet = DotNet; // Ensure reachable from anywhere
```
rynowak: Any reason why this isn't camelCase? just struck me as odd

src/Microsoft.JSInterop/JSRuntimeBase.cs
```diff
@@ -0,0 +1,83 @@
...
+                success,
+                resultOrExceptionMessage
+            }));
+        }
+
+        internal void EndInvokeJS(long asyncHandle, bool succeeded, object resultOrException)
+        {
+            if (!_pendingTasks.TryRemove(asyncHandle, out var tcs))
+            {
+                throw new ArgumentException($"There is no pending task with handle '{asyncHandle}'.");
```
rynowak: I imagine that this exception would only been seen by something like our interop transports - or in the case of mono it can bubble back to JS. We'll have to make sure to handle arbitrary exceptions in those cases unless we want the app to crash.

rynowak:
So I'm imagining that for server and B-lectron we'll have a 'channel' that does remote IO and is responsible for calling the same methods that are called by monoplatform.js. That's the location where we'll do things like call in the appropriate sync context and ensure that the runtime is initialized in the call-context. It seems simple enough.

Let me know if you had a different idea.

src/Microsoft.JSInterop/IJSRuntime.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.JSInterop
+{
+    /// <summary>
+    /// Represents an instance of a JavaScript runtime to which calls may be dispatched.
+    /// </summary>
+    public interface IJSRuntime
```
rynowak: Should we expose any sort of features/capabilities here? Did you plan to add that through more interfaces?

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoPlatform.ts
SteveSandersonMS: Changed my mind. The two approaches I considered both don't fit well.

 * Allocating the 8 bytes on the stack and passing directly as a raw value parameter: The existing code that prepares the params array on the stack assumes 4 bytes per param, and generalising it will come at a perf cost for every invocation, plus complicate the `Platform` interface considerably.
 * Passing as a boxed long: Mono doesn't currently expose any convenient APIs for allocating a heap object of an arbitrary type. And even if it did that's still a heap allocation, so the perf gain over passing as a string is not massive.

For now I'm keeping the "pass as string" behavior. It's not optimal for perf but it only affects this one code path. We can reconsider in the future if we decide it's consequential.

src/Microsoft.AspNetCore.Blazor.Browser/Interop/ArgumentList.cs
SteveSandersonMS: Yep, the updated equivalent to this is: https://github.com/aspnet/Blazor/pull/1033/files#diff-b80ce524aaadc1a948d2f5b50db36160R77

Instead of having to create a special wire format for args arrays and then a lot of .NET-side logic for parsing and extracting values, it was possible just to `JSON.serialize` the whole `args` array on the JS side with no preprocessing, and then do a two-phase deserialize on the .NET side. It shouldn't cost anything extra in perf as in effect it was doing this before but just not retaining the information it needed.

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
```diff
@@ -78,10 +85,19 @@ public void Dispose()
...
+                mono.InvokeUnmarshalled<int, RenderBatch, object>(
+                    "Blazor._internal.renderBatch",
+                    _browserRendererId,
+                    batch);
+            }
+            else
+            {
+                // When implementing support for an out-of-process JS runtime, we'll need to
+                // do something here to serialize and transmit the RenderBatch efficiently.
+                throw new NotImplementedException("TODO: Support BrowserRenderer.UpdateDisplay on other runtimes.");
```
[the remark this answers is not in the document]
SteveSandersonMS: Yes, that's possible! I strongly expect we'll be tweaking the layering a bit when/if we implement another Renderer, since there is probably some aspects of `BrowserRenderer` that we'd want to share across renderers. The exact layer boundaries probably have to move around a bit.

src/Microsoft.AspNetCore.Blazor.Browser/Services/ActivateMonoJSRuntime.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// use this directly, and it will be removed shortly.
+    /// </summary>
+    public class ActivateMonoJSRuntime
+    {
+        static ActivateMonoJSRuntime()
+        {
+            // Temporarily enable MonoWebAssemblyJSRuntime on this class constructor
+            // Later it will become part of the app startup and config mechanism
+            JSRuntime.SetCurrentJSRuntime(new MonoWebAssemblyJSRuntime());
+        }
```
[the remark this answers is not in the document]
SteveSandersonMS: Great, looking forwards to tying all this together.

src/Microsoft.AspNetCore.Blazor.Browser/Services/BrowserServiceProvider.cs
```diff
@@ -17,6 +17,12 @@ public class BrowserServiceProvider : IServiceProvider
     {
         private readonly IServiceProvider _underlyingProvider;
 
+        static BrowserServiceProvider()
+        {
+            // TODO: Remove once we make this part of the app startup mechanism
+            ActivateMonoJSRuntime.EnsureActivated();
```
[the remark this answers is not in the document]
SteveSandersonMS: Good point. Done.

The linker wasn't stripping it out before, but maybe an updated linker might be smart enough to do so. So I agree it's a good idea to `GC.KeepAlive` on a returned object to ensure it isn't stripped.

src/Microsoft.AspNetCore.Blazor/Microsoft.AspNetCore.Blazor.csproj
SteveSandersonMS: It's also to make the JSON utils available, as they are used in a few places in this project.

It might be that one day we can reference some other JSON utils from here and not reference `Microsoft.JSInterop`. For the time being I expect it's fine as-is.

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
SteveSandersonMS: Good point. Done.

src/mono/driver_blazor.c
SteveSandersonMS: Exactly: https://kripken.github.io/emscripten-site/docs/porting/connecting_cpp_and_javascript/Interacting-with-code.html#interacting-with-code-call-javascript-from-native

test/testapps/BasicTestApp/InteropTest/JavaScriptInterop.cs
SteveSandersonMS: Good tip - done.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/ElementReferenceCapture.ts
SteveSandersonMS: If I understand your question correctly then no, it shouldn't be looking at `key`. If we receive the following:

```
{"someElem":{"_blazorElementRef":123}}
```

Then we'd receive a call to the reviver with `key`=`"someElem"` and `value`=`{_blazorElementRef:123}`. The `key` doesn't matter; we just need to check that the `value` looks like an ElementRef, and return an `HTMLElement` instance obtained by reading the `_blazorElementRef` property from `value`.

src/Microsoft.JSInterop/DotNetDispatcher.cs
```diff
@@ -0,0 +1,181 @@
...
+            = new ConcurrentDictionary<string, IReadOnlyDictionary<string, (MethodInfo, Type[])>>();
+
+        /// <summary>
+        /// Receives a call from JS to .NET, locating and invoking the specified method.
+        /// </summary>
+        /// <param name="assemblyName">The assembly containing the method to be invoked.</param>
+        /// <param name="methodIdentifier">The identifier of the method to be invoked. The method must be annotated with a <see cref="JSInvokableAttribute"/> matching this identifier string.</param>
+        /// <param name="argsJson">A JSON representation of the parameters.</param>
+        /// <returns>A JSON representation of the return value, or null.</returns>
+        public static string Invoke(string assemblyName, string methodIdentifier, string argsJson)
```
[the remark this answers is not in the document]
SteveSandersonMS: Good point. Added the following comment to both:

```
// This method doesn't need [JSInvokable] because the platform is responsible for having
// some way to dispatch calls here. The logic inside here is the thing that checks whether
// the targeted method has [JSInvokable]. It is not itself subject to that restriction,
// because there would be nobody to police that. This method *is* the police.
```

src/Microsoft.JSInterop/DotNetDispatcher.cs
SteveSandersonMS: I don't think there will be any problem with that, because this method is `void`, so the caller can't be relying on receiving any synchronous output before the continuation is invoked.

Also since the other side of the interop boundary is getting the continuation as a `Promise` resolution, that already guarantees as per the `Promise` spec that resolutions will be delivered asynchronously. So in practice I don't think it will be possible to observe the callback arriving synchronously.

If anyone ever manages to create a scenario where the difference is observable, we can tweak this to force an async continuation.

src/Microsoft.JSInterop/DotNetDispatcher.cs
SteveSandersonMS: That's a good point. Agreed and updated.

src/Microsoft.JSInterop/DotNetDispatcher.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            // TODO: Consider looking first for assembly-level attributes (i.e., if there are any,
+            // only use those) to avoid scanning, especially for framework assemblies.
+            return GetRequiredLoadedAssembly(assemblyName)
+                .GetExportedTypes()
+                .SelectMany(type => type.GetMethods())
+                .Where(method => method.IsDefined(typeof(JSInvokableAttribute), inherit: false))
+                .ToDictionary(
+                    method => method.GetCustomAttribute<JSInvokableAttribute>(false).Identifier,
+                    method => (method, method.GetParameters().Select(p => p.ParameterType).ToArray())
```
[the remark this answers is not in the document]
SteveSandersonMS: I'm not too worried about that, as it will already throw with a clear message when you try to actually invoke it (e.g., "not a closed generic type"). If we try to anticipate all the cases here there might be a lot to consider.

src/Microsoft.JSInterop/IJSInProcessRuntime.cs
SteveSandersonMS: Yes, we have been using `<object>` for that (the compiler doesn't allow `void` or `System.Void`). I'm aware it's a bit clunky and we can add another overload for that if people turn out to want it.

src/Microsoft.JSInterop/JSRuntimeBase.cs
SteveSandersonMS: OK, I'll leave it for now, but will have a comment there that it might need to be reviewed.

My concern is that in the server-executed case, it's preferable not to be leaking memory if the JS-side code is just badly written and doesn't notify about completion (e.g., on errors, or if it's a void function and the developer hasn't thought about what they are doing). Plus, an individual user could deliberately change the JS-side code running in their browser so it doesn't notify completion even if the developer wrote their app correctly.

We might decide that a timeout is not the best solution even still. Maybe there should be a default per-JS-runtime limit on the number of pending tasks. This would limit it separately for each connected user.

src/Microsoft.JSInterop/JSRuntimeBase.cs
SteveSandersonMS: It is possible, that's true. But our work on sync context should make it easier to get this right than wrong in typical cases. In the even that a developer somehow still gets it wrong, it should be pretty detectable because there will be a lot of errors of the form "there is no pending task with ID x". If you have an idea for how we should improve this please let me know :)

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
SteveSandersonMS: I'm thinking of `DotNet` as being like a static class. It's conventional for class names to be PascalCase in JS, hence static-method-like APIs such as `Promise.race`, `WebAssembly.compile`, etc.

src/Microsoft.JSInterop/JSRuntimeBase.cs
SteveSandersonMS: It should be hard to trigger this, but if you somehow do, it implies something is pretty badly wrong.

In client-side execution, a global unhandled exception is a reasonable outcome. For server-side execution, the code that invokes `DotNetDispatcher.Invoke/BeginInvoke` should try/catch it and do something sensible with unhandled exceptions.

src/Microsoft.JSInterop/IJSRuntime.cs
SteveSandersonMS: Yes, the interfaces mechanism was working fine to distinguish cases where sync or unmarshalled interop is supported. Maybe that will be enough. If not we can figure out what extra capability advertisement would be appropriate when we have scenarios for it.

test/testapps/TestContentPackage/MyPrompt.cs
rynowak: Reminder - we need to do this 😆

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/ElementReferenceCapture.ts
rynowak: Ah ok, you are correct. Thanks for clarifying

src/Microsoft.JSInterop/JSRuntimeBase.cs
rynowak: > should try/catch it and do something sensible with unhandled exceptions.

I'm interested if you have more thoughts, but the only sensible thing I can think of would be to log, or optionally destroy the circuit (seems bad). We don't have the context of a user-operation when this happens because it happens in on resuming async.

test/testapps/TestContentPackage/MyPrompt.cs
SteveSandersonMS: Yep, done in 36dc3368

Awkwardly the new convention of making the JS functions globally-reachable via an expression of the form `window.something.somethingElse.etcetera` means there's no longer a simple way to auto-generate different per-project namespaces matching the project name in JS code. This is because AFAICT [`dotnet new` doesn't have a built-in way of saying you want to inject the "safe" version of the project name](https://github.com/dotnet/templating/issues/402#issuecomment-285605336) in specific places, so we can't use the project name in the JS code because it might contain dots, and our lookup logic interprets dots as property name separators.

Net result is that I've now hard-coded the namespace of the `showPrompt` example function as `exampleJsFunctions`, which I think is adequate to make the point that developers should change it to something that makes sense in their project.

The only way this would be a problem is if people actually start shipping NuGet packages in which the JS functions are still attached to a global called `exampleJsFunctions`, and then trying to use multiple such packages would lead to name clashes. But I think that's extremely unlikely and would obviously be a package authoring error.

galvesribeiro: Yeah, I would not bother with it... A comment in the docs saying that people should change to a meaningful (and safe) name is necessary for sanity...

PR #1052, opened 2018-06-28
Add RenderBatch reader abstraction

SteveSandersonMS:
This is also part of the out-of-process execution work. It separates the knowledge of `RenderBatch` (etc) memory layout from the logic that actually uses the information in the `RenderBatch` (etc).

With this change, we'll be able to continue with shared-memory interop for the in-proc scenario, and plug in a different `RenderBatch` reader for out-of-process hosting. It means we can use any serialization format we want for the `RenderBatch` without having to simulate the same memory layout, which in turn will allow for a more compact and easily-implemented representation.

**Implementation note**

The reason we have all these `xyzReader` APIs that act upon references to read properties (e.g., `frameReader.attributeName(frame)` instead of `frame.attributeName`) is so that we don't have to parse all the incoming data into a large graph of objects. We keep the incoming data in its original form (e.g., a single byte array) and walk through it without instantiating any child objects to represent different parts of it. This avoids the need for any up-front parsing (especially good for parts of the data structure we might turn out not to even look at) and dramatically lessens GC pressure.


src/Microsoft.AspNetCore.Blazor.Browser.JS/src/GlobalExports.ts
```diff
@@ -12,7 +14,7 @@ if (typeof window !== 'undefined') {
 
     _internal: {
       attachRootComponentToElement,
-      renderBatch,
+      renderBatch: (browserRendererId: number, batchAddress: Pointer) => renderBatch(browserRendererId, new SharedMemoryRenderBatch(batchAddress)),
```
SteveSandersonMS: Currently it's hard-coded here to use `SharedMemoryRenderBatch`, but later we'll have different startup logic depending on the hosting model.

PR #1055, opened 2018-06-29
Add sync context

rynowak:
This seems like 90% done to me, but I wanted to get eyeballs on it in case you had any feedback.

I know I at least want to add another test that verifies you can `await` and resume in the sync context. I also might need to add some more code to the synchronous code path to apply the sync context.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
@@ -1,0 +1,269 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Blazor.Server.Circuits
+{
+    [DebuggerDisplay("{_state,nq}")]
+    internal class CircuitSynchronizationContext : SynchronizationContext
+    {
+        private static readonly ContextCallback ExecutionContextThunk = (object state) =>
+        {
+            var item = (WorkItem)state;
+            item.SynchronizationContext.ExecuteSynchronously(null, item.Callback, item.State);
+        };
+
+        private static readonly Action<Task, object> BackgroundWorkThunk = (Task task, object state) =>
+        {
+            var item = (WorkItem)state;
+            item.SynchronizationContext.ExecuteBackground(item);
+        };
+
+        private readonly State _state;
+
+        public event UnhandledExceptionEventHandler UnhandledException;
+
+        public CircuitSynchronizationContext()
+            : this(new State())
+        {
+        }
+
+        private CircuitSynchronizationContext(State state)
+        {
+            _state = state;
+        }
+
+        public Task Invoke(Action action)
+        {
+            var completion = new TaskCompletionSource<object>();
+            Post(_ =>
+            {
+                try
+                {
+                    action();
+                    completion.SetResult(null);
+                }
+                catch (Exception exception)
+                {
+                    completion.SetException(exception);
+                }
+            }, null);
+
+            return completion.Task;
+        }
+
+        public Task InvokeAsync(Func<Task> asyncAction)
+        {
+            var completion = new TaskCompletionSource<object>();
+            Post(async (_) =>
+            {
+                try
+                {
+                    await asyncAction();
+                    completion.SetResult(null);
+                }
+                catch (Exception exception)
+                {
+                    completion.SetException(exception);
+                }
+            }, null);
+
+            return completion.Task;
+        }
+
+        public Task<TResult> Invoke<TResult>(Func<TResult> function)
+        {
+            var completion = new TaskCompletionSource<TResult>();
+            Post(_ =>
+            {
+                try
+                {
+                    var result = function();
+                    completion.SetResult(result);
+                }
+                catch (Exception exception)
+                {
+                    completion.SetException(exception);
+                }
+            }, null);
+
+            return completion.Task;
+        }
+
+        public Task<TResult> InvokeAsync<TResult>(Func<Task<TResult>> asyncFunction)
+        {
+            var completion = new TaskCompletionSource<TResult>();
+            Post(async (_) =>
+            {
+                try
+                {
+                    var result = await asyncFunction();
+                    completion.SetResult(result);
+                }
+                catch (Exception exception)
+                {
+                    completion.SetException(exception);
+                }
+            }, null);
+
+            return completion.Task;
+        }
+
+        // asynchronously runs the callback
+        public override void Post(SendOrPostCallback d, object state)
+        {
+            TaskCompletionSource<object> completion;
+            lock (_state.Lock)
+            {
+                if (!_state.Task.IsCompleted)
+                {
+                    _state.Task = Enqueue(_state.Task, d, state);
+                    return;
+                }
+
+                // We can execute this synchronously because nothing is currently running
+                // or queued.
+                completion = new TaskCompletionSource<object>();
+                _state.Task = completion.Task;
+            }
+
+            ExecuteSynchronously(completion, d, state);
+        }
+
+        // synchronously runs the callback
+        public override void Send(SendOrPostCallback d, object state)
+        {
+            Task antecedant;
+            var completion = new TaskCompletionSource<object>();
+
+            lock (_state.Lock)
+            {
+                antecedant = _state.Task;
+                _state.Task = completion.Task;
+            }
+
+            // We have to block. That's the contract of Send - we don't expect this to be used
+            // in many scenarios in Blazor.
+            //
+            // Using Wait here is ok because the antecedant task will never throw.
+            antecedant.Wait();
+
+            ExecuteSynchronously(completion, d, state);
+        }
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
```diff
@@ -0,0 +1,261 @@
...
+            item.Callback(item.State);
+        };
+
+        private static readonly WaitCallback BackgroundWorkThunk = (object state) =>
+        {
+            var context = (CircuitSynchronizationContext)state;
+            context.ExecuteBackgroundWork();
+        };
+
+        private State _state;
```
rynowak: This can be readonly

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        public event UnhandledExceptionEventHandler UnhandledException;
```
rynowak: I chose to make this pump unhandled exceptions on the async call path via an event. The idea is that whatever error handling strategy we need can be attached to this. 

Note that the typical call paths here will use `InvokeAsync` and so will get their errors in a normal way 😆 

You'd only get an unhandled exception if you threw from an async void method.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
```diff
@@ -0,0 +1,261 @@
...
+                return;
+            }
+
+            // Perf - using a static thunk here to avoid a delegate allocation.
+            ExecutionContext.Run(item.Context, ExecutionContextThunk, item);
+        }
+
+        private void ExecuteBackgroundWork()
+        {
+            Monitor.Enter(_state.Monitor);
```
rynowak: Note that this might block. I'm debating changing this to tasks instead of QUWI to avoid this, but it will be much much more complicated. 

There's a pattern used increasingly often to keep an instance field holding a task that represents the 'current' execution, and simply call `ContinueWith` and replace the task with the new task. This enforces ordering pretty nicely and doesn't require blocking - I'm sure this would invoke more allocations in some common cases though so there's a tradeoff. 

Overall I'm more worried about the blocking. Threads are a limited resource, allocations much more so.

rynowak: Note that the System.Web sync-context uses task-chaining.

rynowak:
Hmm, well I guess I have a compilation error

image

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
davidfowl: I actually came here to say don’t block and mimic the system.web context of anything.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
```diff
@@ -0,0 +1,269 @@
...
+            {
+                handler(this, new UnhandledExceptionEventArgs(ex, isTerminating: false));
+            }
+        }
+
+        private class State
+        {
+            public bool IsBusy; // Just for debugging
+            public object Lock = new object();
+            public Task Task = Task.CompletedTask;
```
rynowak: I played with using ValueTask here and it's not incredibly possible due to the fact that we want to avoid allocations when we execute synchronously - since the value task is returned at the end of the method. 

We'd basically end up inventing task completion source since it's needs to be a heap object - due to send being sync you need to be able to communicate with multiple pending work items.

I think this is the best we'll be able to do for now if we want to be non-blocking.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
```diff
@@ -0,0 +1,269 @@
...
+                    completion.SetException(exception);
+                }
+            }, null);
+
+            return completion.Task;
+        }
+
+        public Task<TResult> Invoke<TResult>(Func<TResult> function)
+        {
+            var completion = new TaskCompletionSource<TResult>();
```
rynowak: Note we still require a separate task/completion here due to the way we surface exceptions and return values.

src/Microsoft.AspNetCore.Blazor.Browser.JS/package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-6OaHAsknBA6M2gKszMZXunqFofGXCCk4UCXHMdzd3qtBpndSHuM2JgxBE9M3APyl/DlENt4FEe0C7mJwbcC/ZA==",
```
SteveSandersonMS: @rynowak Do you happen to know what triggered all the changes here? Did you regenerate the file from scratch?

Since there aren't any `package.json` changes in this PR, it may be that the `package-lock.json` is not needed.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
```diff
@@ -0,0 +1,269 @@
...
+            var handler = UnhandledException;
+            if (handler != null)
+            {
+                handler(this, new UnhandledExceptionEventArgs(ex, isTerminating: false));
+            }
+        }
+
+        private class State
+        {
+            public bool IsBusy; // Just for debugging
```
SteveSandersonMS: Could we rename this to `IsBusy_OnlyUsedForDebugging` or something, just to make sure it doesn't one day start getting used for other things without first verifying that it would be safe everywhere?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
```diff
@@ -0,0 +1,269 @@
...
+                return $"{{ Busy: {IsBusy}, Pending Task: {Task} }}";
+            }
+        }
+
+        private class WorkItem
+        {
+            public CircuitSynchronizationContext SynchronizationContext;
+            public ExecutionContext ExecutionContext;
+            public SendOrPostCallback Callback;
+            public object State;
```
SteveSandersonMS: I don't totally follow why this `State` property exists, as it seems like that only gets passed to `ExecuteSynchronously`, which can already get the same data from `_state`. Of course I might be wrong as it's hard to track execution from reading the PR.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
SteveSandersonMS: Sure. I wouldn't worry about this massively! In practice we'll be enqueuing work items when we receive inbound interop calls, which will then typically perform a render, serialize the result, and send an outbound notification, all of which will do more allocations (hopefully not loads, but some). One further `Task` allocation here isn't likely to be measurable.

test/Microsoft.AspnetCore.Blazor.Server.Test/Microsoft.AspnetCore.Blazor.Server.Test.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <PackageReference Include="xunit" Version="2.3.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Blazor.Server\Microsoft.AspNetCore.Blazor.Server.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Properties\" />
```
SteveSandersonMS: Probably not needed.

src/Microsoft.AspNetCore.Blazor.Browser.JS/package-lock.json
rynowak: I've never really be able to figure out what's going on with this file. I regularly clean/rebuild my repro. 1/20 times I forget and accidentally just include it in a PR.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
rynowak: There are cases where the runtime makes a copy of a sync context. We want all of those copies to share state. 

I'm not actually sure of the details of how and when this occurs - but this is a pattern that isolates us from it.

rynowak:
> Developers could kill their apps by returning some task that never completes, but I expect it's the same with MVC actions and so on.

Well, you could do that by blocking a thread - but not by returning say `Task.Delay(int32.MaxValue)`. We never await user code, we call it synchronously. If user code awaits some long-running thing, we'll execute up to the await and then go do something else 😆. From our point of view each await is potentially the end of a 'work item'.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitSynchronizationContext.cs
rynowak: Sure

PR #1061, opened 2018-06-29
Update versions to 2.1

rynowak:
Updates ASP.NET package versions to 2.1

Tweaks TFMs
- libraries = netstandard2.0
- exes/apps/tools = netcoreapp2.1
- unit tests = netcoreapp2.0
- e2e tests = netcoreapp2.1 (dependency on apps)

The exception to this is the Browser.JS project which depends on our
tool for its build. however this project just builds the js code so its
TFM doesn't really matter.


test/Microsoft.AspNetCore.Blazor.Build.Test/RuntimeDependenciesResolverTest.cs
```diff
@@ -67,17 +67,30 @@ public void FindsReferenceAssemblyGraph_ForStandaloneApp()
...
                 "System.Numerics.dll",
+                "System.Reflection.Emit.ILGeneration.dll",
+                "System.Reflection.Emit.Lightweight.dll",
+                "System.Reflection.Primitives.dll",
+                "System.Resources.ResourceManager.dll",
+                "System.Runtime.dll",
+                "System.Runtime.Extensions.dll",
                 "System.Runtime.Serialization.dll",
                 "System.ServiceModel.Internals.dll",
+                "System.Threading.dll",
```
rynowak: The set of references changed with netcoreapp2.1 - should we be concerned about this?

test/Microsoft.AspNetCore.Blazor.Test/Routing/RouteTableTests.cs
```diff
@@ -305,10 +305,21 @@ public TestRouteTableBuilder AddRoute(string template, Type handler = null)
...
+                {
+                    return new RouteTable(_routeTemplates
+                        .Select(rt => new RouteEntry(TemplateParser.ParseTemplate(rt.Item1), rt.Item2))
+                        .OrderBy(id => id, RouteTable.RoutePrecedence)
+                        .ToArray());
+                }
+                catch (InvalidOperationException ex) when (ex.InnerException is InvalidOperationException)
+                {
+                    // ToArray() will wrap our exception in its own.
+                    throw ex.InnerException;
```
rynowak: The behavior changed here and I'm not 100% certain why - it's possible that KoreBuild is now using netcoreapp2.1 to run tests.

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Server/BlazorHosted-CSharp.Server.csproj
```diff
@@ -8,12 +8,8 @@
     <LangVersion>7.3</LangVersion>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(Framework)' != 'netcoreapp2.0'">
-    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Blazor.Server" Version="$(TemplateBlazorPackageVersion)" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(Framework)' == 'netcoreapp2.0'">
-    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
+  <ItemGroup Condition="'$(Framework)' == 'netcoreapp2.1'">
```
SteveSandersonMS: Could we drop the `Condition` here now there's only going to be one choice for this?

test/Microsoft.AspNetCore.Blazor.Build.Test/RuntimeDependenciesResolverTest.cs
SteveSandersonMS: I'd like to know what the set of dependencies the default project ends up with after the linker runs.

 * Is it much bigger?
 * Does it try to use `System.Reflection.Emit.*.dll` for anything at runtime (which would be very slow on Mono WebAssembly, and won't work at all with AoT)? I guess this would be related to the ASP.NET change to DI that does some IL generation.

If it's not significantly bigger, and especially if it links out `System.Reflection.Emit.*.dll` completely, then I'm sure it would be fine.

If it is significantly bigger it raises more difficult questions.

test/Microsoft.AspNetCore.Blazor.Test/Routing/RouteTableTests.cs
SteveSandersonMS: That's strange. Probably not something to worry about though, as it's just changing the structure of the exceptions in the case of a fatal error. Nobody could be relying on the old behavior.

src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Server/BlazorHosted-CSharp.Server.csproj
rynowak: I left it here because it will just come back when we do the next big push. If you think it's important I can take it out.

SteveSandersonMS: OK, that makes sense. Reasonable to leave it then.

rynowak:
Doing a publish of `StandaloneApp` - comparing the size on disk of the `dist/_framework/bin` folder:

Before: 2.27 mb (12 files)
After: 2.28 mb (12 files)

The extra dependencies aren't showing up in the publish output.

I think the reason why this isn't happening is that the client applications all target netstandard2.0 not netcoreapp2.1. The IL emit feature for ASP.NET DI is behind an ifdef https://github.com/aspnet/DependencyInjection/blob/94b9cc9ace032f838e068702cc70ce57cc883bc7/src/DI/DI.csproj

rynowak:
@SteveSandersonMS - does this explanation make sense?

SteveSandersonMS:
Yes, that makes sense, and it's good to see there's no significant size penalty! Thanks for verifying this.

PR #1070, opened 2018-07-02
[WIP] ModelState/DataAnnotation library

uazo:
I've written a form validation library that use the standard ComponentModel.DataAnnotation and I've added a test page in StandaloneApp.

the library it's very simple and allow sharing models with dataannotations between blazor and backend (ref #17).

all the logic is in the ``ModelStateDictionary<T>`` class (so isn't in the form, as #970) and allow code like this

````
    var ModelState = new ModelStateDictionary<TestModel>(new TestModel())


        <div class="col-md-8 col-sm-8">
            <input type="text" bind="@ModelState[x => x.TextValue]" class="form-control" />
            <span class="text-danger">@ModelState.ValidationFor(x=> x.TextValue)</span>
        </div>

        <button onclick="@(() => this.UpdateModel())">Update model</button>


    protected virtual void UpdateModel()
    {
        this.ValidateModel();
        if (this.ModelState.IsValid)
        {
            this.ModelState.Update();
        }
    }
````

Since `<input>` tag values are always strings, ModelStateDictionary does an automatic type conversion of all values using `TypeConverter` (since it's available in netstandard2.0): putting an invalid value (like a non convertible string into number) will show an error automatically.
For now it's support short, int, long, datetime, float, double and nullable

ModelStateDictionary can update the Model via `Update(T model)` function.

The library support custom validation (with virtual `OnCustomValidateModel` and `AddModelError`).

It's a starting point (no xml comments or test cases), but seems work.


SteveSandersonMS:
I'm making a pass through some of the older PRs to see what we can clear up.

This one certainly looks interesting, but our hope would be that this type of validation library could be implemented as an external package rather than having to be embedded in the core. If you're still interested in proceeding with it, would you be able to see what would be needed to make it a separate package? If additional extensibility APIs are needed we can certainly consider adding them. Hope that's OK!

Issue #1071, opened 2018-07-02
Incorrect values in disposeComponent (BrowserRenderer.ts)


uazo:
Some strange value after last commit (I think #1052)

![browserrenderer ts error](https://user-images.githubusercontent.com/29201891/42177132-e663a630-7e2b-11e8-9d85-11e435585c8c.png)

Renderer.RenderInExistingBatch seems correct, but SharedMemoryRenderBatch.disposedComponentIdsEntry returns invalid values.

SteveSandersonMS:
Thanks for reporting this and for providing such a clear fix!

PR #1072, opened 2018-07-02
Fix for #1071

uazo:


SteveSandersonMS:
Thanks very much @uazo - glad you were able to track that down so quickly!

PR #1081, opened 2018-07-05
Update to Mono Linker d18d497

SteveSandersonMS:
Prerequisite for integrating with debugger, because the updated linker knows how to emit symbol files.


SteveSandersonMS:
Merged immediately because there's not really much to review here, and I need this change for some other work elsewhere. But @rynowak please feel free to review if you wish!

PR #1085, opened 2018-07-05
Initial debugger support

SteveSandersonMS:
This integrates with the debugger support that @kumpera has been working on. Although it looks like a big changeset, it's not really as big as it looks at first, because:

 1. This PR also includes a Mono upgrade, which touches many .dlls
 2. This PR also includes the Mono ws-proxy (debugging proxy) sources

### How to use

* Build a Blazor app in `Debug` configuration (i.e., the default for non-published apps)
* Run it in Chrome
* With the keyboard focus on the app (not in the dev tools, which you should probably close as it's less confusing that way), press the hotkey:
    * Shift+Alt+D on Windows/Linux
    * Shift+Cmd+D on macOS

This will open a debugger tab. After a couple of seconds, the debugger's *Sources* tab will show a list of the .NET assemblies in the app. You can expand these and find the `.cs`/`.cshtml` source files you want to debug. You can then set breakpoints, switch back to your app's tab, and cause the breakpoints to be hit. You can then single-step (F10) or resume (F8).

You have to be running Chrome with the debug port open. If you aren't doing so, it will detect that and give you instructions for fixing it.

### Caveats

**The debugger capabilities are very limited!** You can:

 * Single-step through the current method (F10) or resume (F8)
 * In the *Locals* display, observe the values of any local variables of type `int`/`string`/`bool`
 * See the call stack, including call chains that go from JS into .NET and vice-versa

... but for now that is all! You *cannot* step into child methods (i.e., F11), observe the values of any locals that aren't an `int`/`string`/`bool`, observe the values of any class properties or fields, hover over variables to see their values, evaluate expressions in the console, step across async calls, or basically anything else.

@kumpera has done some great work on tackling some of the hardest technical problems to enable source viewing, breakpoints, and stepping, but please be patient as completing the long tail of debugger features remains a significant task.

### Issues

After a minute left idle, the debugger will disconnect and you'll have to click "Reconnect" to continue. I know @kumpera was doing something to fix that. Not sure how close this is to being resolved.

### Not implemented

Any E2E test for launching and using the debugger. I started on one, but realised that WebDriver itself is using the Chrome debug port to automate the browser, so we can't also use it to control the debugger via ws-proxy. Maybe there is some way to hack both on at once, but it's not clear.


src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.props
```diff
@@ -11,7 +11,7 @@
 
   <PropertyGroup Label="Blazor build outputs">
     <MonoLinkerI18NAssemblies>none</MonoLinkerI18NAssemblies> <!-- See Mono linker docs - allows comma-separated values from: none,all,cjk,mideast,other,rare,west -->
-    <AdditionalMonoLinkerOptions>-c link -u link -t --verbose </AdditionalMonoLinkerOptions>
+    <AdditionalMonoLinkerOptions>-c link -u link -b true -t --verbose </AdditionalMonoLinkerOptions>
```
SteveSandersonMS: `-b true` means "emit PDBs"

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/MonoDebugProxyAppBuilderExtensions.cs
```diff
@@ -0,0 +1,178 @@
...
+    <h1>Unable to find debuggable browser tab</h1>
+    <p>
+        Could not get a list of browser tabs from <code>{debuggerTabsListUrl}</code>.
+        Ensure Chrome is running with debugging enabled.
+    </p>
+    <h2>Resolution</h2>
+    {GetLaunchChromeInstructions(appRootUrl)}
+    <p>... then use that new tab for debugging.</p>
+    <h2>Underlying exception:</h2>
+    <pre>{ex}</pre>
```
SteveSandersonMS: It would have been nice to use server-side Razor Pages to implement this basic built-in UI, but it seemed like using precompiled Razor views in a NuGet package involves conforming to Razor Pages file layout and routing conventions (e.g., having a `Pages` directory) which is not useful here.

We can maybe change this if we end up implementing more sophisticated built-in server-side UIs.

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/ws-proxy/Startup.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.AspNetCore.Routing;
+using System.Net.WebSockets;
+using System.Threading;
+using System.IO;
+using System.Text;
+
+namespace WsProxy
+{
+
+	internal class Startup {
```
SteveSandersonMS: I changed all the ws-proxy types to be `internal` instead of `public`.

src/mono/incoming/README.md
```diff
@@ -16,52 +16,63 @@ emcc -g4 -Os -s WASM=1 -s ALLOW_MEMORY_GROWTH=1 -s BINARYEN=1 -s "BINARYEN_TRAP_
 
 ## Sample
 
+For this we'll use the included sample web page in the sdk.
```
SteveSandersonMS: All the changes in ` src/mono/dist/unoptimized` are due to upgrading to a newer build of Mono WebAssembly needed for debugger support.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoDebugger.ts
```diff
@@ -0,0 +1,50 @@
...
+
+let hasReferencedPdbs = false;
+
+export function hasDebuggingEnabled() {
+  return hasReferencedPdbs && currentBrowserIsChrome;
+}
+
+export function attachDebuggerHotkey(loadAssemblyUrls: string[]) {
+  hasReferencedPdbs = loadAssemblyUrls
+    .some(url => /\.pdb$/.test(getAssemblyNameFromUrl(url, /* includeExtension */ true)));
```
rynowak: Should this just be a new function - `getFileNameFromUrl`? It really is a different thing...

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/ws-proxy/WsProxy.cs
```diff
@@ -0,0 +1,331 @@
...
+		}
+
+		public Task Pump (CancellationToken token)
+		{
+			current_send = null;
+			pending.RemoveAt (0);
+
+			if (pending.Count > 0) {
+				if (current_send != null)
+					throw new Exception ("WTF, current_send MUST BE NULL IF THERE'S no pending send");
```
SteveSandersonMS: If only we were allowed to write exception messages like this

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoDebugger.ts
SteveSandersonMS: Yes, I agree. I'll change that.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Platform/Mono/MonoDebugger.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (!hasReferencedPdbs) {
+        console.error('Cannot start debugging, because the application was not compiled with debugging enabled.');
+      } else if (!currentBrowserIsChrome) {
+        console.error('Currently, only Chrome is supported for debugging.');
+      } else {
+        launchDebugger();
+      }
+    }
+  });
+}
```
rynowak: I might have missed this when you and Rodrigo were discussing, what's the reason why we need to open a new thing?

SteveSandersonMS: Do you mean why we need to open a new tab? It's so we can have a debugger UI that's connected to the ws-proxy that intercepts all the commands and injects .NET info into the results. I'm not aware of any way to make the regular dev tools (on the same tab) connect to an external V8 protocol endpoint.

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/MonoDebugProxyAppBuilderExtensions.cs
```diff
@@ -0,0 +1,178 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Newtonsoft.Json;
+using WsProxy;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    internal static class MonoDebugProxyAppBuilderExtensions
+    {
+        public static void UseMonoDebugProxy(this IApplicationBuilder applicationBuilder)
```
rynowak: convention is to call this `app`

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/MonoDebugProxyAppBuilderExtensions.cs
rynowak: I think this is fine for now. We don't want to actually bring in MVC and the MVC view engine into this part of the codes.

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/MonoDebugProxyAppBuilderExtensions.cs
```diff
@@ -0,0 +1,178 @@
...
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                return $@"<p>Close this browser, then in a terminal window execute the following:</p>
+                          <p><strong><code>google-chrome --remote-debugging-port=9222 {appRootUrl}</code></strong></p>";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return $@"<p>Close this browser, then in a terminal window execute the following:</p>
+                          <p><strong><code>open /Applications/Google\ Chrome.app --remote-debugging-port=9222 {appRootUrl}</code></strong></p>";
+            }
```
rynowak: Is this the default or will users have to launch chrome this way every time? (Does VS do this for instance)?

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/MonoDebugProxyAppBuilderExtensions.cs
```diff
@@ -0,0 +1,178 @@
...
+                throw new InvalidOperationException("Unknown OS platform");
+            }
+        }
+
+        private static async Task<IEnumerable<BrowserTab>> GetOpenedBrowserTabs(string debuggerHost)
+        {
+            using (var httpClient = new HttpClient { Timeout = TimeSpan.FromSeconds(5) })
+            {
+                var jsonResponse = await httpClient.GetStringAsync($"{debuggerHost}/json");
+                return JsonConvert.DeserializeObject<BrowserTab[]>(jsonResponse);
```
rynowak: This should check the status code and then use `ReadAsAsync` https://www.fuget.org/packages/System.Net.Http.Formatting.Extension/5.2.3/lib/net/System.Net.Http.Formatting.dll/System.Net.Http/HttpContentExtensions#M%3ASystem.Net.Http.HttpContentExtensions.ReadAsAsync%60%601%28System.Net.Http.HttpContent%29

This package is part of ASP.NET Core as of 2.0

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/ws-proxy/Startup.cs
rynowak: OK. I'm largely just ignoring these for now.

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/MonoDebugProxyAppBuilderExtensions.cs
SteveSandersonMS: Unfortunately it's not the default. You can change your own VS config to launch Chrome with `--remote-debugging-port=9222` by editing the list of browsers in the "Browse With" dialog, but I'm not aware of a way of putting that config in `launchSettings.json`.

Long term we might want to ask about launching Chrome with that flag by default for Blazor project types.

For now people will just have to follow the instructions to open a new instance. Fortunately it's only once: for as long as that browser instance remains open, VS will open new tabs in it each time you Ctrl+F5.

src/Microsoft.AspNetCore.Blazor.Server/MonoDebugProxy/MonoDebugProxyAppBuilderExtensions.cs
SteveSandersonMS: Totally happy to change this later, but I'm not sure where the `ReadAsAsync` extension method really is.

The package you linked appears not to be owned by Microsoft: https://www.nuget.org/packages/System.Net.Http.Formatting.Extension/5.2.3. The most likely standard distribution package I can find is `System.Net.Http.Formatting`, but that doesn't appear to be in `netstandard2.0`.

Since I think the behaviour will be equivalent, I've left it using `JsonConvert`. But if you know of a `netstandard2.0` way of using `ReadAsAsync` please let me know!

DNF-SaS:
Hmm, permanent 404 after Shift-Alt-D .. :-(

PR #1086, opened 2018-07-05
Fix #954 - use weak typing for 'event handlers'

### LINKED ISSUE #954: Binding to events to a component that inherits from a c# class broke in 0.4.0
The following worked in 0.3.0:

DynamicElement is defined here: https://github.com/aspnet/Blazor/blob/stevesa/bootstrap-components-library/src/Microsoft.AspNetCore.Blazor.Bootstrap/DynamicElement.cs

```
<DynamicElement onclick="@onclick" ...>...</DynamicElement>

@functions {
  [Parameter] private Action onclick { get; set; }
}
```

in 0.4.0 an error on build:

```
Argument 1: cannot convert from 'System.MulticastDelegate' to 'string'
```

rynowak:
This change introduces a mechanism for bypassing type checking and then
uses for the 'event handlers'. The event handler tag helpers have some
ideosyncratic behaviors and rely on overloading at the render tree
builder level.


rynowak:
@SteveSandersonMS  - feel free to take a look if it interests you, I'm going to merge this since it seems like the simplest possible solution

PR #1090, opened 2018-07-06
Refactoring to prepare for remote rendering.

SteveSandersonMS:
This PR takes the parts of the `bots` branch that I think are clean and can exist separately from remote rendering, and will put them in `dev` so that the remaining remote rendering work becomes simpler.

- Prepare for building multiple entrypoint variants of the .js library
- Use async interop more consistently for rendering and event handling
- Add binary serializer for RenderBatch with tests

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
@@ -320,13 +320,7 @@ function countDescendantFrames(batch: RenderBatch, frame: RenderTreeFrame): numb
   }
 }
 
-function raiseEvent(event: Event, browserRendererId: number, componentId: number, eventHandlerId: number, eventArgs: EventForDotNet<UIEventArgs>) {
-  if (!raiseEventMethod) {
-    raiseEventMethod = platform.findMethod(
-      'Microsoft.AspNetCore.Blazor.Browser', 'Microsoft.AspNetCore.Blazor.Browser.Rendering', 'BrowserRendererEventDispatcher', 'DispatchEvent'
-    );
-  }
-
+async function raiseEvent(event: Event, browserRendererId: number, componentId: number, eventHandlerId: number, eventArgs: EventForDotNet<UIEventArgs>) {
   const eventDescriptor = {
     browserRendererId,
     componentId,
@@ -334,8 +328,9 @@ function raiseEvent(event: Event, browserRendererId: number, componentId: number
     eventArgsType: eventArgs.type
   };
 
-  platform.callMethod(raiseEventMethod, null, [
-    platform.toDotNetString(JSON.stringify(eventDescriptor)),
-    platform.toDotNetString(JSON.stringify(eventArgs.data))
-  ]);
+  await DotNet.invokeMethodAsync(
+    'Microsoft.AspNetCore.Blazor.Browser',
+    'DispatchEvent',
+    eventDescriptor,
+    JSON.stringify(eventArgs.data));
 }
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/Microsoft.AspNetCore.Blazor.Browser.JS.csproj
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/Microsoft.AspNetCore.Blazor.Browser.JS.csproj
@@ -26,7 +26,8 @@
 
   <Target Name="RunWebpack" AfterTargets="ResolveReferences" Inputs="@(WebpackInputs)" Outputs="dist\blazor.webassembly.js" DependsOnTargets="EnsureNpmRestored">
     <RemoveDir Directories="dist" />
-    <Exec Command="npm run build" />
+    <Exec Command="npm run build:debug" Condition="'$(Configuration)' == 'Debug'" />
+    <Exec Command="npm run build:production" Condition="'$(Configuration)' != 'Debug'" />
     <ItemGroup>
       <EmbeddedResource Include="dist\blazor.webassembly.js" LogicalName="blazor./blazor.webassembly.js" />
     </ItemGroup>
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/Microsoft.AspNetCore.Blazor.Browser.JS.csproj
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/Microsoft.AspNetCore.Blazor.Browser.JS.csproj
@@ -24,11 +24,11 @@
     <Exec Command="npm install" />
   </Target>
 
-  <Target Name="RunWebpack" AfterTargets="ResolveReferences" Inputs="@(WebpackInputs)" Outputs="dist\blazor.js" DependsOnTargets="EnsureNpmRestored">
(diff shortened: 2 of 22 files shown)


src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
```diff
@@ -320,22 +320,17 @@ function countDescendantFrames(batch: RenderBatch, frame: RenderTreeFrame): numb
...
 
-  platform.callMethod(raiseEventMethod, null, [
-    platform.toDotNetString(JSON.stringify(eventDescriptor)),
-    platform.toDotNetString(JSON.stringify(eventArgs.data))
-  ]);
+  await DotNet.invokeMethodAsync(
+    'Microsoft.AspNetCore.Blazor.Browser',
+    'DispatchEvent',
+    eventDescriptor,
+    JSON.stringify(eventArgs.data));
```
rynowak: Is the idea here to have these go though 'user-level' interop so that they don't require a platform-specific implementation?

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Blazor.Server
+{
+    // TODO: We should consider *not* having this type of infrastructure in the .Server
+    // project, but instead in some new project called .Remote or similar, since it
+    // would also be used in Electron and possibly WebWorker cases.
```
rynowak: Or if could even go in the main blazor assembly if the dependency profile is slim.

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
```diff
@@ -0,0 +1,235 @@
...
+    ///    so it's no disadvantage for the location info to be at the end.
+    ///  * We only serialize the data that the JS side will need. For example, we don't
+    ///    emit frame sequence numbers, or any representation of nonstring attribute
+    ///    values, or component instances, etc.
+    ///    
+    /// We don't have or need a .NET reader for this format. We only read it from JS code.
+    /// </summary>
+    internal class RenderBatchWriter : IDisposable
+    {
+        private readonly List<string> _strings;
```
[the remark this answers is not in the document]
SteveSandersonMS: I did consider that, but since I think we'll be very likely to want to put compression around this anyway, that will have exactly the same effect (well, it will do better, because it will de-dupe common substrings as well as complete ones). So assuming we do that, having extra logic here to do poor-person's compression might be undesirable.

If later it turns out we aren't going to compress the websocket traffic then yes I definitely think we should do that.

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
```diff
@@ -0,0 +1,235 @@
...
+    // TODO: We should consider *not* having this type of infrastructure in the .Server
+    // project, but instead in some new project called .Remote or similar, since it
+    // would also be used in Electron and possibly WebWorker cases.
+
+    /// <summary>
+    /// Provides a custom binary serializer for <see cref="RenderBatch"/> instances.
+    /// This is designed with both server-side and client-side perf in mind:
+    /// 
+    ///  * Array-like regions always have a fixed size per entry (even if some entry types
+    ///    don't require as much space as others) so the recipient can index directly.
```
rynowak: This doesn't seem true of https://github.com/aspnet/Blazor/pull/1090/files#diff-a60b475d488f645eaaa5af2613faaf8fR61

I'm following what you're doing in the code, but I'm not sure these comments are accurate.

rynowak: It might be worth listing which types and fixed size and which are variable size.

SteveSandersonMS: `ArrayRange<RenderTreeDiff>` does output a fixed size per entry (each entry is a single int, which acts as a lookup into a different table of info).

I guess the comment in a bit ambiguous. The part I mean is "fixed size" is the array data itself, not any other data that it points to from there.

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
```diff
@@ -0,0 +1,235 @@
...
+                case RenderTreeFrameType.Component:
+                    _binaryWriter.Write(frame.ComponentSubtreeLength);
+                    _binaryWriter.Write(frame.ComponentId);
+                    _binaryWriter.Seek(4, SeekOrigin.Current); // Padding
+                    break;
+                case RenderTreeFrameType.ComponentReferenceCapture:
+                    // The client doesn't need to know about these. But we still have
+                    // to include them in the array otherwise the ReferenceFrameIndex
+                    // values in the edits data would be wrong.
+                    _binaryWriter.Seek(12, SeekOrigin.Current); // Padding
```
rynowak: I think this might throw from non-seekable streams. Generally memory and file streams are seekable, and networking-related streams are not.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
SteveSandersonMS: Yes, unless there would be a meaningful perf gain from doing something more custom and low-level, it's more convenient to use the interop that's abstracted across platforms.

In this specific case I don't think there'd be any perf gain from going more low-level.

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
SteveSandersonMS: It could. Long term we'd want to avoid `Microsoft.AspNetCore.Blazor.dll` getting big since it's used on the client, but if it's just this one class we could put that in there.

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
SteveSandersonMS: Yes, it would throw in that case. Currently we only use this with `MemoryStream`, which is why it's currently OK.

Do you think it would be better to explicitly write zeros to the stream instead of seeking forwards like this?

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
SteveSandersonMS: Though I guess it should be linked out if not used, so maybe there's no disadvantage to putting it in there.

SteveSandersonMS:
Reminder to self: move `RenderBatchWriter` out of the `Bots` directory before merging this.

src/Microsoft.AspNetCore.Blazor.Server/Bots/RenderBatchWriter.cs
rynowak: Writing zeros would be better since it would be portable across all stream types, and it doesn't seem too complicated

PR #1092, opened 2018-07-07
Add startup for client/mono Blazor

rynowak:
This change adds a host builder, and the startup pattern for client-side
Blazor apps running in mono/wasm. This will help us align better with
server side Blazor.

### DIFF
--- a/samples/StandaloneApp/Program.cs
+++ b/samples/StandaloneApp/Program.cs
@@ -1,8 +1,7 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.AspNetCore.Blazor.Browser.Rendering;
-using Microsoft.AspNetCore.Blazor.Browser.Services;
+using Microsoft.AspNetCore.Blazor.Hosting;
 
 namespace StandaloneApp
 {
@@ -10,12 +9,11 @@ namespace StandaloneApp
     {
         public static void Main(string[] args)
         {
-            var serviceProvider = new BrowserServiceProvider(configure =>
-            {
-                // Add any custom services here
-            });
-
-            new BrowserRenderer(serviceProvider).AddComponent<App>("app");
+            CreateHostBuilder(args).Build().Run();
         }
+
+        public static IWebAssemblyHostBuilder CreateHostBuilder(string[] args) =>
+            BlazorWebAssemblyHost.CreateDefaultBuilder()
+                .UseBlazorStartup<Startup>();
     }
 }
--- a/samples/StandaloneApp/Startup.cs
+++ b/samples/StandaloneApp/Startup.cs
@@ -1,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Blazor.Builder;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace StandaloneApp
+{
+    public class Startup
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+        }
+
+        public void Configure(IBlazorApplicationBuilder app)
+        {
+            app.AddComponent<App>("app");
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Blazor.Browser/Builder/BlazorApplicationBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Browser/Builder/BlazorApplicationBuilderExtensions.cs
@@ -1,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Blazor.Components;
+
+namespace Microsoft.AspNetCore.Blazor.Builder
+{
+    /// <summary>
+    /// Provides extension methods for <see cref="IBlazorApplicationBuilder"/>.
+    /// </summary>
+    public static class BlazorApplicationBuilderExtensions
+    {
+        /// <summary>
+        /// Associates the component type with the application,
+        /// causing it to be displayed in the specified DOM element.
+        /// </summary>
+        /// <param name="app">The <see cref="IBlazorApplicationBuilder"/>.</param>
+        /// <typeparam name="TComponent">The type of the component.</typeparam>
+        /// <param name="domElementSelector">A CSS selector that uniquely identifies a DOM element.</param>
+        public static void AddComponent<TComponent>(this IBlazorApplicationBuilder app, string domElementSelector)
+            where TComponent : IComponent
+        {
+            app.AddComponent(typeof(TComponent), domElementSelector);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Blazor.Browser/Builder/IBlazorApplicationBuilder.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Browser/Builder/IBlazorApplicationBuilder.cs
@@ -1,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Blazor.Components;
+
+namespace Microsoft.AspNetCore.Blazor.Builder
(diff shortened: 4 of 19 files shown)


src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// <summary>
+        /// Runs the application.
+        /// </summary>
+        /// <param name="host">The <see cref="IBrowserHost"/> to run.</param>
+        /// <remarks>
+        /// Currently, Blazor applications running in the browser don't have a lifecycle - the application does not
+        /// get a chance to gracefully shut down. For now, <see cref="Run(IBrowserHost)"/> simply starts the host
+        /// and allows execution to continue.
+        /// </remarks>
+        public static void Run(this IBrowserHost host)
```
rynowak: This is what we discussed on the call this morning. Notice that the 'host' doesn't actually shut down in our case, we just start it and let it go.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            host.StartAsync().GetAwaiter().GetResult();
```
rynowak: This will return pretty much immediately.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostingEnvironment.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Blazor.Hosting
+{
+    internal class BrowserHostingEnvironment : IBrowserHostingEnvironment
+    {
+    }
```
rynowak: So this is one of the features of the web/generic host that we might want to parallel if we have any useful data to plug in here. Thoughts? If we don't have ideas I'll take it out.

src/Microsoft.AspNetCore.Blazor/Builder/BlazorApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// causing it to be displayed in the specified DOM element.
+        /// </summary>
+        /// <param name="app">The <see cref="IBlazorApplicationBuilder"/>.</param>
+        /// <typeparam name="TComponent">The type of the component.</typeparam>
+        /// <param name="domElementSelector">A CSS selector that uniquely identifies a DOM element.</param>
+        public static void AddComponent<TComponent>(this IBlazorApplicationBuilder app, string domElementSelector)
+            where TComponent : IComponent
+        {
+            app.AddComponent(typeof(TComponent), domElementSelector);
+        }
```
rynowak: We didn't get to a super high level of clarity about this earlier. I'm happy to move this to the browser library for now and we can revisit when/if it becomes necessary elsewhere

rynowak:
Note that I haven't done anything to the template yet. Waiting for feedback and then I'll add that on. 

My next step after this one is to start building out the server project.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHost.cs
```diff
@@ -0,0 +1,101 @@
...
+                }
+
+                throw;
+            }
+
+
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken = default)
```
SteveSandersonMS: I'm guessing this never actually happens in the browser case. Or is there any way the developer can trigger it?

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostBuilder.cs
```diff
@@ -0,0 +1,103 @@
...
+            var services = new ServiceCollection();
+            services.AddSingleton(_hostingEnvironment);
+            services.AddSingleton(_BrowserHostBuilderContext);
+            services.AddSingleton<IBrowserHost, BrowserHost>();
+            services.AddSingleton<IJSRuntime, MonoWebAssemblyJSRuntime>();
+
+            services.AddSingleton<IUriHelper, BrowserUriHelper>();
+            services.AddSingleton<HttpClient>(s =>
+            {
+                // Creating the URI helper needs to wait until the JS Runtime is initialized, so defer it.
```
SteveSandersonMS: Calling `GetBaseUri` on init was convenient when I did it the first time, but I appreciate it complicates things here. It will also complicate things in the server case.

No need to change this now, but later, we should consider having the base URI and current URI be something that gets passed into the .NET code as part of the "please boot up now" invocation.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostExtensions.cs
SteveSandersonMS: Makes sense.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostingEnvironment.cs
SteveSandersonMS: If our only scenario was browser hosting, then I think this would be pretty much definitely something we'd want eventually (e.g., to indicate development vs production environment).

But since we will have other scenarios too, I'm unsure what would end up on `IBrowserHostingEnvironment` as opposed to something more general and not browser specific.

Personally I'd err on the side of leaving this out until the use case becomes clearer, but if you feel there are some clear cases where we'd want a browser-specific hosting environment concept then it's fine with me.

src/Microsoft.AspNetCore.Blazor/Builder/BlazorApplicationBuilderExtensions.cs
SteveSandersonMS: Yes, it would be good to have it in the `.Browser` project because of it having DOM-specific concepts.

I hopeful the layering will pay off with some extra clarity when we get to server-side prerendering and unit test helpers, but that remains to be seen. We can merge the browser/nonbrowser layers together later if it turns out the distinction isn't helping.

samples/StandaloneApp/Program.cs
```diff
@@ -1,21 +1,19 @@
...
-            var serviceProvider = new BrowserServiceProvider(configure =>
-            {
-                // Add any custom services here
-            });
-
-            new BrowserRenderer(serviceProvider).AddComponent<App>("app");
+            CreateHostBuilder(args).Build().Run();
         }
+
+        public static IBrowserHostBuilder CreateHostBuilder(string[] args) =>
```
SteveSandersonMS: As a general thought about terminology, I wonder if we should say `IWebAssemblyHostBuilder` (etc) for all the concepts that are specific to running in the browser on WebAssembly.

That might be a clearer distinction to make rather than calling those "browser" apps, since even the remote-rendered ones run in the browser to some extent.

The "browserness" that I was meaning to capture in the `.Browser` project was more to do with rendering on a browser DOM, as opposed to rendering in memory for server-side prerendering or unit tests. I think that's orthogonal to whether the .NET part is running on the server/electron for remote rendering or under WebAssembly.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostingEnvironment.cs
rynowak: OK I'll take this out for now.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHost.cs
rynowak: No there's no way to trigger it right now. 

The purpose of this (and reason to leave it in) would be if we want to expose the ability for services to register that want to know the application is shutting down. This would be the place that we trigger that logic.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/BrowserHostBuilder.cs
rynowak: Yeah that would simplify things a good deal.

samples/StandaloneApp/Program.cs
rynowak: As discussed, I'll rename all of these

PR #1099, opened 2018-07-08
Removed trailing space from template survey link

jnm2:
Before:

After:


SteveSandersonMS:
Thanks for the suggestion, @jnm2!

I don't think we'd want to add the comment tag there in the template, as it will confuse people. Especially those less familiar with the technology.

I would prefer the solution of putting the whole `<a>` tag on a single line (even though it will be a long line). @jnm2, are you interested in changing it that way?

jnm2:
@SteveSandersonMS Absolutely, no problem!

SteveSandersonMS:
Thanks @jnm2! It's merged.

jnm2:
Thank you!

PR #1103, opened 2018-07-09
Change ElementRef IDs to be strings, not ints

SteveSandersonMS:
We don't want to use ints for elementref IDs for remote rendering, because:

1. Eventually we could run out of them. A few billion isn't enough.
2. Having a global static incrementing sequence discloses cross-user state (number of IDs assigned)

This PR changes the IDs to be strings.

Then for non-WebAssembly scenarios we use GUIDs. For WebAssembly we use an incrementing sequence of long values (with `.ToString()`), just because there's no need to go through all the GUID code paths. That's just an internal implementation detail though. Consumers just treat the IDs as opaque strings.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
@@ -153,7 +153,7 @@ export class BrowserRenderer {
         return this.insertFrameRange(batch, componentId, parent, childIndex, frames, frameIndex + 1, frameIndex + frameReader.subtreeLength(frame));
       case FrameType.elementReferenceCapture:
         if (parent instanceof Element) {
-          applyCaptureIdToElement(parent, frameReader.elementReferenceCaptureId(frame));
+          applyCaptureIdToElement(parent, frameReader.elementReferenceCaptureId(frame)!);
           return 0; // A "capture" is a child in the diff, but has no node in the DOM
         } else {
           throw new Error('Reference capture frames can only be children of element frames.');
--- a/src/Microsoft.AspNetCore.Blazor/PlatformInfo.cs
+++ b/src/Microsoft.AspNetCore.Blazor/PlatformInfo.cs
@@ -1,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.InteropServices;
+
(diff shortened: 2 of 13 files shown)


src/Microsoft.AspNetCore.Blazor/PlatformInfo.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Blazor
+{
+    internal static class PlatformInfo
+    {
+        public static bool IsWebAssembly { get; }
```
SteveSandersonMS: I know we might eventually have reasons to make this `public`, but it isn't required right now.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
```diff
@@ -153,7 +153,7 @@ export class BrowserRenderer {
         return this.insertFrameRange(batch, componentId, parent, childIndex, frames, frameIndex + 1, frameIndex + frameReader.subtreeLength(frame));
       case FrameType.elementReferenceCapture:
         if (parent instanceof Element) {
-          applyCaptureIdToElement(parent, frameReader.elementReferenceCaptureId(frame));
+          applyCaptureIdToElement(parent, frameReader.elementReferenceCaptureId(frame)!);
```
rynowak: Educational for me... what does the `!` mean in this context?

vertonghenb: @rynowak '!' means that you're sure the object will have some value (a.k.a, you're sure it's not going to be NULL) sometimes the compiler can complain if it's a nullable type. But if you're sure it's not going to be NULL in this context, you can overwrite it using '!'

rynowak: cool, thanks!

SteveSandersonMS: Yep, TypeScript treats types as non-nullable by default in `strict` mode, which have enabled.

In this case, `elementReferenceCaptureId` is declared as returning `string | null`, whereas `applyCaptureIdToElement` requires a `string`. Since we know we only go down this code path for the `FrameType.elementReferenceCapture` case, we know the `elementReferenceCaptureId` will be nonnull.

PR #1105, opened 2018-07-09
Avoid globally unique BrowserRenderer IDs

SteveSandersonMS:
I'm going through all teh codez to check for places where we're falsely assuming there's only a single user. For example, anywhere we use `static` to store per-user state. Any such code needs to be changed so it's safe for remote rendering. Fortunately there only appears to be two places where this applied, addressed by this PR and by #1103.

This PR is a partial step towards not letting users interfere with each other by dispatching events with somebody else's browserRendererId. The fix is not to have globally-unique IDs at all, but rather store the `BrowserRenderer` instances in user-scoped state. We still have browser renderer IDs, but they are per-circuit, which also has the benefit that we won't overflow the int value in the long run.

I can't yet really put in the user-scoped state because that depends on the work in #1092. Somehow when there are incoming JS interop calls, we need to set the correct `IJSRuntime` for that user, plus maybe we need an equivalent context to have the correct `IServiceProvider` for that user. Will discuss with @rynowak.

In the meantime this PR just simplifies and clarifies where the per-user storage needs to be plugged in.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRendererRegistry.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRendererRegistry.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRendererRegistry.cs
```diff
@@ -12,18 +12,26 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
     /// their associated component instances aren't GCed when events may still
     /// be received for them.
     /// </summary>
-    internal static class BrowserRendererRegistry
+    internal class BrowserRendererRegistry
     {
-        private static int _nextId;
-        private static IDictionary<int, BrowserRenderer> _browserRenderers
+        private int _nextId;
+        private IDictionary<int, BrowserRenderer> _browserRenderers
```
rynowak: We will likely need to make this thread-safe as well in the future.

SteveSandersonMS: AFAICT it already is, but please let me know if I'm missing something!

rynowak: What I mean is that using a `BrowserRendererRegistry` is not thread safe because `Dictionary<,>` is not thread safe.

SteveSandersonMS: The dictionary is only read/written from inside `lock` blocks. That should be sufficient, isn't it?

rynowak: Oh you're right, I'm not sure how I missed that sorry

PR #1108, opened 2018-07-09
trying to fix 1009 by changing the AssignToProperties method behavior

RemiBou:
So here is my other attempt on fixing this bug, I'll let the blazor team decide which way they prefer to go


RemiBou:
@SteveSandersonMS can you give a quick feedback on this PR ? If you think it needs more work please tell me. The build error doesn't seems to be related to the PR content but I'll dig it if you think I'm onto something.

RemiBou:
Ok I'll have a look at this. i'm not a big expert on avoiding allocations, but I'll try to dig it. Thanks for the feedback

RemiBou:
@SteveSandersonMS I changed the code a it so there is less allocation :
- got rid of WriteParameterAction (its creation showed a warning in clr heap analyzer that the compiler will emit a new class holding the property setter as a field) an use only IPropertySetter
- Add the default value in propertySetter
- use a local copy of the propertywriters list and remove the used one, then browse the remainings to set the default value
- the only warning I got in Clr Heap Allocation Analyzer seems to be the static fields, the exceptions and the loop line 77 but we don't have much choice here.

And I renamed the method to "SetParameterProperty" because of the attribute name.

What's the point of the generic PropertySetter here ? it's interface is not using generic parameters.

RemiBou:
@SteveSandersonMS ok the build is fixed after fixing the conflict with  PR #1545

RemiBou:
I'll fix the CI error this week, I don't understand how my change causes error on non-Windows E2E test but it's not failing on other PR so ...

RemiBou:
Hi all, do I need to keep this PR up-to-date with master, choose myself how to fix the CI (see previous comment), improve the code somehow, or should I close it ?

SteveSandersonMS:
Currently we’re focused on the code migration work. Once that’s done we can resume attending to PRs.

In the meantime if you want to simplify migrating this PR you could rebase it on master. Thanks!

RemiBou:
Ok sure I'll do that. If there is any stuff we can do for the migration do
not hesitate to add some up-for-grabs.

Rémi

Le dim. 25 nov. 2018 11:32, Steve Sanderson <notifications@github.com> a
écrit :

RemiBou:
Merge is done, the same tests are failing.

FYI the file move was not correctly handled by git, I had to reproduce all the changes on the new files. It might be a problem on bigger PR.

SteveSandersonMS:
Thanks again for this, @RemiBou! It's now moved to https://github.com/aspnet/AspNetCore/pull/4797, where it will be merged shortly.

PR #1116, opened 2018-07-11
[WIP] Blazor on the server

rynowak:
Putting this out there for early feedback. 

Things are basically in the right places, and rendering and event handling works. I still have some issues to work out with how our client code deals with HTTP clients.

I haven't made any serious attempts at error handling yet.

TODO:
- [x] Error Handling
- [x] Refactor Renderer
- [x] Refactor UriHelper
- [x] Remove UriHelper interop hack
- [x] Allow Server-Side startup to consume client-side startup
- [x] Tests For Server Side Sample

Planned to log issues for follow-up:
- [x] Allow server boot script to specify a non-default URL for startup #1130
- [x] Add object pool for serialization #1132
- [x] Make server boot script work with non-default pathbase #1130
- [x] TODO: Also support browsers that don't have TextDecoder (e.g., Edge) #1131
- [x] Run E2E Tests on Server Side #1129


src/Microsoft.AspNetCore.Blazor.Browser.JS/package.json
```diff
@@ -13,6 +13,8 @@
     "ts-loader": "^4.4.1",
     "typescript": "^2.9.2",
     "webpack": "^4.12.0",
-    "webpack-cli": "^3.0.8"
+    "webpack-cli": "^3.0.8",
+    "@aspnet/signalr": "^1.0.0",
+    "@aspnet/signalr-protocol-msgpack": "^1.0.0"
```
rynowak: The JS code is unchanged from your version, I didn't have to make any changes here though I expect I will have to integrate the rendererId change

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
```diff
@@ -18,6 +19,11 @@ public class BrowserRenderer : Renderer, IDisposable
     {
         private readonly int _browserRendererId;
 
+        /// <summary>
```
rynowak: My plan is still to create a subclass of `Renderer` for server-side execution. I haven't gotten to it yet 👍

src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorCircuitsApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.AspNetCore.Blazor.Browser.Rendering;
+using Microsoft.AspNetCore.Blazor.Server.Circuits;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Extension methods to configure an <see cref="IApplicationBuilder"/> for Blazor circuits.
+    /// </summary>
+    public static class BlazorCircuitsApplicationBuilderExtensions
```
rynowak: Did we get to a conclusion about branding for this stuff?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/Circuit.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Represents an active connection between a Blazor server and a client.
+    /// </summary>
+    public class Circuit
+    {
+        internal Circuit(CircuitHost circuitHost)
+        {
+
+        }
```
rynowak: Note that there's nothing here yet on purpose. This class is for user-code - `CircuitHost` is for us.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            try
+            {
+                await SynchronizationContext.Invoke(() =>
+                {
+                    Microsoft.JSInterop.JSRuntime.SetCurrentJSRuntime(JSRuntime);
+
+                    switch (methodIdentifier)
+                    {
+                        // Massive hack. Need a common system for dispatching calls within the context of
```
rynowak: I just copied your hack for now. We'll need to solve this or add interop for instances.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+            catch (Exception ex)
+            {
+                UnhandledException?.Invoke(this, new UnhandledExceptionEventArgs(ex, isTerminating: false));
+            }
+        }
+
+        public void Dispose()
+        {
+        }
```
rynowak: Note to self, this needs to dispose the service provider

src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+            var scope = _scopeFactory.CreateScope();
+            var renderer = new BrowserRenderer(scope.ServiceProvider);
+            var jsRuntime = new RemoteJSRuntime(client);
+            var synchronizationContext = new CircuitSynchronizationContext();
+
+            var circuitHost = new CircuitHost(scope, renderer, config, jsRuntime, synchronizationContext);
+
+            // Initialize per-circuit data that services need
+            (circuitHost.Services.GetRequiredService<IJSRuntimeAccessor>() as DefaultJSRuntimeAccessor).JSRuntime = jsRuntime;
+            (circuitHost.Services.GetRequiredService<ICircuitAccessor>() as DefaultCircuitAccessor).Circuit = circuitHost.Circuit;
```
rynowak: These interfaces provide access for scoped services to access the JS runtime and circuit via DI.

rynowak: Internal for now, but I expect we'd make them public.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteUriHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public void Initialize(string uriAbsolute, string baseUriAbsolute)
+        {
+            _baseUriStringWithTrailingSlash = ToBaseUri(baseUriAbsolute);
+            _baseUriWithTrailingSlash = new Uri(_baseUriStringWithTrailingSlash);
+
+            _uriAbsolute = uriAbsolute;
+
+            _jsRuntimeAccessor.JSRuntime.InvokeAsync<object>(_functionPrefix + "enableNavigationInterception");
+        }
```
rynowak: LMK what you think of these changes. In general you can't allow services to have a dependency on data without some kind of lifecycle. I opted to do the simplest thing which is to have a lifecycle with 2 steps 🤣

samples/HostedInAspNet.Client/wwwroot/index.html
```diff
@@ -7,6 +7,6 @@
 <body>
     <app>Loading...</app>
     <script src="customJsFileForTests.js"></script>
-    <script type="blazor-boot"></script>
+    <script src="_framework/blazor.server.js"></script>
```
SteveSandersonMS: To keep the WebAssembly version consistent with the server-executed version, I'm thinking we should change the WebAssembly startup mechanism to use a tag like:

    <script src="_framework/blazor.webassembly.js"></script>

... instead of the magic `<script type="blazor-boot">`. It would work by first issuing a request for a `_framework/appconfig.json` file that was generated during the build and contains the same information that we normally put onto the <script type="blazor-boot">.

This also has the benefit that we no longer need to "build" the `index.html` file; it just becomes a normal static file. In turn this means the developer can either serve the `index.html` from their client app or put the `<script>` tag in any server-rendered page.

@rynowak If you agree, I'll implement that in a separate PR.

samples/HostedInAspNet.Client/Program.cs
```diff
@@ -9,7 +9,7 @@ public class Program
     {
         static void Main(string[] args)
         {
-            new BrowserRenderer().AddComponent<Home>("app");
+            // Not used
```
SteveSandersonMS: Guessing this change isn't going to be merged in the end

samples/HostedInAspNet.Server/Startup.cs
```diff
@@ -9,21 +10,27 @@ namespace HostedInAspNet.Server
 {
     public class Startup
     {
-        // This method gets called by the runtime. Use this method to add services to the container.
-        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddBlazorCircuits();
```
SteveSandersonMS: `AddBlazorServerExecution`?

samples/HostedInAspNet.Server/Startup.cs
```diff
@@ -9,21 +10,27 @@ namespace HostedInAspNet.Server
...
+
+            services.AddHttpClient();
+            services.AddScoped<HttpClient>(s => s.GetRequiredService<IHttpClientFactory>().CreateClient());
```
SteveSandersonMS: For my education, why do we need these lines to add `HttpClient`?

samples/HostedInAspNet.Server/Startup.cs
```diff
@@ -9,21 +10,27 @@ namespace HostedInAspNet.Server
...
-        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();
             }
 
-            app.UseBlazor<Client.Program>();
+            app.UseBlazorOnTheServer(renderer =>
```
SteveSandersonMS: `UseServerSideBlazor`?

SteveSandersonMS: I expect you've considered this, but what about the possibility of changing this to just:

    app.UseServerSideBlazor<Client.Startup>();

... so that the nomination of root components is contained in the Blazor app code?

Also I'm not sure where the developer would configure per-circuit DI services, such as an `AppState` singleton. Do you envision this going in `ConfigureServices` in the server project? Is it relatively simple to scope something to a circuit?

If it was possible to have `app.UseServerSideBlazor<Client.Startup>();` here and then configure per-circuit DI services in the `ConfigureServices` method of the `Startup` class in the Blazor client lib, that would be extremely clear I think and would also make it simpler for developers to try switching from server-executed to WebAssembly-executed hosting models early in their project (until such time as they strongly take dependency on other aspects of a chosen hosting model).

src/Microsoft.AspNetCore.Blazor.Browser.JS/package.json
SteveSandersonMS: Great!

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Boot.Server.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as Environment from './Environment';
+import * as signalR from '@aspnet/signalr';
+import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
+import { OutOfProcessRenderBatch } from './Rendering/RenderBatch/OutOfProcessRenderBatch';
+import { internalFunctions as uriHelperFunctions } from './Services/UriHelper';
+import { renderBatch } from './Rendering/Renderer';
+
+function boot() {
+  const connection = new signalR.HubConnectionBuilder()
+    .withUrl('/_blazor')
```
SteveSandersonMS: I don't think I verified this works with non-root URLs (nonempty `PathBase`) so it's possible we might have to tweak something here.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Boot.Server.ts
```diff
@@ -0,0 +1,39 @@
...
+        }
+      });
+
+      connection.send(
+        'StartCircuit',
+        uriHelperFunctions.getLocationHref(),
+        uriHelperFunctions.getBaseURI()
+      );
+    })
+    .catch(err => console.error(err.toString()));
```
SteveSandersonMS: We might be better off without the `.toString()` part so that the dev console can display it as a structured object (so you can expand/collapse its properties, etc.).

src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorCircuitsApplicationBuilderExtensions.cs
SteveSandersonMS: I like the term "circuit" to represent the stateful aspect of a live-updating UI connection and would be totally happy to keep it.

For the services that people are configuring though, terms like `UseServerSideBlazor` and `AddServerSideBlazor` feel more natural to me, in the sense that it's what we keep saying by default in our discussions and pretty unambiguously reflect the type of UI mechanism we're offering.

src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorCircuitsApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// Temporary vague approximation to server-side execution needed so I can
+        /// build the rest of the interop.
+        /// </summary>  
+        public static IApplicationBuilder UseBlazorOnTheServer(
+            this IApplicationBuilder builder,
+            Action<BrowserRenderer> startupAction)
+        {
+            var endpoint = "/_blazor";
+
+            var factory = builder.ApplicationServices.GetRequiredService<CircuitFactory>() as DefaultCircuitFactory;
```
SteveSandersonMS: Just curious, do you anticipate a scenario where there would be any other type of circuit factory? Is this approach (as opposed to just hard-coding a reference to a particular concrete class here) mainly for consistency with other ASP.NET patterns?

SteveSandersonMS: Oh wait it looks like you're doing this as a way of supplying a preconfigured object to the SignalR hub. Is that right? If so makes sense.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/BlazorHub.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Blazor.Services;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Blazor.Server.Circuits
+{
+    internal class BlazorHub : Hub
+    {
+        private static readonly object CircuitKey = "circuit";
```
SteveSandersonMS: It's a minor point, but would the value `new object()` be preferable over `"circuit"` to make it less prone to outside interference?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/BlazorHub.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public override Task OnDisconnectedAsync(Exception exception)
+        {
+            CircuitHost.Dispose();
+            return base.OnDisconnectedAsync(exception);
+        }
+
+        public async Task StartCircuit(string uriAbsolute, string baseUriAbsolute)
+        {
+            var circuitHost = _circuitFactory.CreateCircuit(Context.GetHttpContext(), Clients.Caller);
```
SteveSandersonMS: Would `CreateCircuitHost` be a clearer name?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
```diff
@@ -0,0 +1,106 @@
...
+using Microsoft.AspNetCore.Blazor.Browser.Rendering;
+using Microsoft.AspNetCore.Blazor.Services;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.JSInterop;
+
+namespace Microsoft.AspNetCore.Blazor.Server.Circuits
+{
+    internal class CircuitHost : IDisposable
+    {
+        public event UnhandledExceptionEventHandler UnhandledException;
```
SteveSandersonMS: Do you know if there's any way we can wire this up to the SignalR hub so that, if there's an unhandled exception, we (1) send that info to the client so it can be displayed in the browser console and (2) terminate the SignalR connection?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
SteveSandersonMS: When we receive incoming JS interop calls to .NET, is there any way to access the scoped service provider? I expect there will need to be, e.g., for having an `AppState` scoped to a circuit.

If it is possible to do that, then `NotifyLocationChanged` can invoke some static method that accesses the scoped `RemoteUriHelper` to call its instance method.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    internal class DefaultCircuitFactory : CircuitFactory
+    {
+        private readonly IServiceScopeFactory _scopeFactory;
+
+        public DefaultCircuitFactory(IServiceScopeFactory scopeFactory)
+        {
+            _scopeFactory = scopeFactory ?? throw new ArgumentNullException(nameof(scopeFactory));
+
+            StartupActions = new List<(PathString path, Action<BrowserRenderer> config)>();
```
SteveSandersonMS: This looks good. It's nice to see this solution for how to configure multiple server-side Blazor apps while only having one top-level service provider.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+        public List<(PathString path, Action<BrowserRenderer> config)> StartupActions { get; }
+
+        public override CircuitHost CreateCircuit(HttpContext httpContext, IClientProxy client)
+        {
+            Action<BrowserRenderer> config = null;
+            for (var i = 0; i < StartupActions.Count; i++)
+            {
+                if (httpContext.Request.Path.StartsWithSegments(StartupActions[i].path))
+                {
+                    config = StartupActions[i].config;
```
SteveSandersonMS: Should we be running *all* the startup actions for a given path, or just the last one?

I'm unsure what the use case would be for having multiple startup actions for a given path. If there is only to be one, it would be good to enforce that by storing them in a dictionary.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
SteveSandersonMS: Wouldn't the `IJSRuntime` itself be a scoped DI service? If so maybe we wouldn't need a further `IJSRuntimeAccessor`.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/BlazorHub.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            var uriHelper = circuitHost.Services.GetRequiredService<IUriHelper>() as RemoteUriHelper;
+            if (uriHelper != null)
```
SteveSandersonMS: As it stands, it looks like there will always be a `IUriHelper` because it's set up in `AddBlazorCircuits`.

I can't think of any important reason why we should make it optional. But if we are having it mandatory, then:

 * Probably don't need to check for nullness here
 * We can probably do away with the `enableNavigationInterception` call from .NET to JS, and have `Boot.WebAssembly.ts` call it directly on startup

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteUriHelper.cs
SteveSandersonMS: The way it passes in `uriAbsolute` and `baseUriAbsolute` looks great.

As per comment above, I'm wondering if we could eliminate the `InvokeAsync ... enableNavigationInterception` and put that logic directly in `Boot.WebAssembly.ts` since it's not going to be optional.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private class RenderBatchFormatter : IMessagePackFormatter<RenderBatch>
+        {
+            public static readonly RenderBatchFormatter Instance = new RenderBatchFormatter();
+
+            // No need to accept incoming RenderBatch
+            public RenderBatch Deserialize(byte[] bytes, int offset, IFormatterResolver formatterResolver, out int readSize)
+                => throw new NotImplementedException();
+
+            public int Serialize(ref byte[] bytes, int offset, RenderBatch value, IFormatterResolver formatterResolver)
```
SteveSandersonMS: I toyed with using some kind of object pool for the underlying buffer here to avoid initializing multi-KB arrays on every interaction. Didn't do it at first because I was just focused on the functionality rather than that sort of perf optimization.

What do you think - is the cost of large object allocation big enough to warrant putting in an object pool here? Or do we decide later?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/Circuit.cs
SteveSandersonMS: I'm curious about what sort of thing would eventually go here.

masonwheeler:
The more I see of server-side Blazor, the less I like the idea.  The whole point of Blazor is to run the CLR in the browser.  Turning it into Yet Another Server Framework not only undermines that purpose, but it lowers the quality of Blazor by introducing network latency into DOM rendering where there was none before.

And yes, I understand that this is optional, but there are still three problems that it causes anyway:

1. The architecture still has to be able to support it, even though it's optional.  This means various breaking changes, including completely breaking JS interop and making everything necessarily async, regardless of whether or not the underlying JavaScript code requires asynchrony, because the communication with the server would require asynchrony even if you choose not to use it.  This is not in any way a good thing, and should be in and of itself a good enough reason to kill the idea.
2. Working on it takes scarce resources away from other things that ought to be a higher priority.  We still don't even have debugging, for example!  Not to mention the SignalR client that's been scheduled for the next release for every single release now.
3. If you make it possible for people to do, some people will do it.  Any server-side Blazor program someone writes is something that can't be used offline as a PWA, because it's tied to the server.

For about decade now, ever since the smartphone went mainstream, we've been moving in the right direction as an industry, moving computation off of servers as much as possible and back to local devices.  WASM gives us the ability to do so in the browser itself, at a greater scope than has ever been possible before.  Putting Blazor on the server would buck this very important trend when we should be encouraging it and becoming leaders in it.

Please reject this PR, and the fundamental concept it's based upon.  It's the wrong direction for Blazor and for the wider industry.

SteveSandersonMS:
Thanks for your feedback, @masonwheeler!

We know that server-side Blazor isn't going to be for everyone, or for every project. Our plan is to let people try out the server model in the 0.5.0 release and hopefully let us know how useful it is or isn't for the actual application scenarios people have in mind. If both models turn out to be useful, it might be a way of doubling the appeal of Blazor at little extra engineering cost, while also providing significant extra flexibility to developers. We just have to see when people start building with it.

On your specific concerns:

>  making everything necessarily async, regardless of whether or not the underlying JavaScript code requires asynchrony

Yes, you're right, it does force us to change that. However:

 * This change would be needed for a possible Electron model too, so there's a really big payoff from it
 * If you're going to use WebAssembly hosting exclusively, then you are still free to do synchronous interop! The updated interop APIs allow for either synchronous or async calls. You the developer just have to choose whether you want to support out-of-process execution (for server and electron), and if you do you need to limit yourself to async interop calls. If you are committed to WebAssembly hosting only, you're totally free to continue making sync calls.

> Working on it takes scarce resources away from other things that ought to be a higher priority. We still don't even have debugging, for example! Not to mention the SignalR client that's been scheduled for the next release for every single release now.

We think the payoff from this might be big, hence prioritising the investigation now.

On debugging, we do have initial support in the 0.5.0 release. We've made great progress with this in the last couple of weeks!

On SignalR, hopefully this is transitioning to be a community-led effort.

>  Any server-side Blazor program someone writes is something that can't be used offline as a PWA, because it's tied to the server.

Yes, and there are things that server-side Blazor apps can do trivially that client-side ones can't (e.g., talk directly to a database, or use NuGet packages that do things that probably will never work on the client such as GPU-accelerated computation). There are pros and cons with both hosting models, so we can best understand what's useful by letting people try each of them.

In summary, I agree with a lot of your points and we will take your overall position on board. Please continue to give us feedback about this and other aspects of Blazor, especially once you've tried out the server hosting model!

masonwheeler:
> This change would be needed for a possible Electron model too, so there's a really big payoff from it

Hmm... what's the value proposition there?  Because I'm having a hard time seeing one.

 * The purpose of Electron is to enable web developers to write desktop apps in their familiar technologies of JavaScript and HTML.
* The purpose of Blazor is to enable C# developers to write web apps in *their* familiar technologies and take JavaScript out of the picture as much as possible.
* Combining the two feels like Rube Goldberg's preferred solution for writing a desktop app in C# 😛 
* The only unique thing this brings to the picture is the ability to build the UI in HTML, but that feels like a step backwards compared to the power of virtually every platform's native widget set, or even existing cross-platform solutions such as Xamarin Forms.  (Which, unlike Electron, is not restricted to desktop systems.)

So I have to wonder, what benefit does Blazor + Electron bring?

> If you're going to use WebAssembly hosting exclusively, then you are still free to do synchronous interop!

That's a relief.  The initial discussion of this "feature" seemed to indicate othewise.

> Yes, and there are things that server-side Blazor apps can do trivially that client-side ones can't (e.g., talk directly to a database, or use NuGet packages that do things that probably will never work on the client such as GPU-accelerated computation).

Sure, but those are things that servers have always been able to do without Blazor.  This is what I mean by  "turning it into Yet Another Server Framework."  Executing C# in the browser is something new and exciting that advances the state of the art in web development.  Executing C# on the server... well... just *isn't.*

Andrzej-W:
Some people prefer client and some prefer server version. We have already seen arguments from both sides. Now it looks that server version is almost ready. Let's try it.

In the mean time I have a few questions about this technology.

1. Lets assume we have an input box with bind expression. We don't have any function attached to OnKeyDown or similar events. User types something into this input box. Is this fully handled in the browser or we have client-server communication after every key press? Hopefully we will have some communication only when variable/property have to be updated in OnChange event.

2. User resizes browser window or changes the orientation of mobile device. Is this fully handled in the browser or visible page fragment is rerendered on the server and send to the browser?

3. Web page is longer (higher) than browser window. What do you send to the browser? Full page or only visible part?

4. Probably related to the above. I have a fixed height `div` element with long table inside. I see scrollbar and I can scroll this table inside `div` element. Do you send whole table to the browser and scrolling is fully handled in the browser?

SteveSandersonMS:
@Andrzej-W Let’s talk about all that stuff once 0.5.0 is out, as it will all probably be quite clear once you see it. Plus I don’t want to derail this PR with a massive conversation about meta-issues. Let’s keep conversation on this PR about the code in the PR itself. Hope that’s ok!

Andrzej-W:
You are right @SteveSandersonMS , sorry.

samples/HostedInAspNet.Server/Startup.cs
rynowak: I'm really confused by this feedback, this is opposite of the discussion we had that lead to https://github.com/aspnet/Blazor/commit/128b38fce62bd8d495b7993099cb39aced601b60


-- 
> In both cases these would invoke the corresponding methods on Client.Startup (i.e., ConfigureServices and Configure) so that the developer can configure DI services there if they want. This would be useful if they wanted, for example, to register a scoped AppState when that type is internal to the Blazor library project.

In this case where is the service provider that's being configured? Is it the main ASP.NET Core service provider?

rynowak: Actually I missed something here. This doesn't look like anything else we've done but we could try it.

samples/HostedInAspNet.Server/Startup.cs
rynowak: ASP.NET Core doesn't add register an HttpClient as a service.

samples/HostedInAspNet.Server/Startup.cs
rynowak: 👍

samples/HostedInAspNet.Server/Startup.cs
SteveSandersonMS: I'm unsure which specific bit of discussion it's the opposite to (having just re-read the comments on #1092), but I fully recognise I might be forgetting something we covered verbally!

> In this case where is the service provider that's being configured? Is it the main ASP.NET Core service provider?

Yes, I think we're proposing to have only one service provider.

samples/HostedInAspNet.Client/wwwroot/index.html
rynowak: Yeah, this sounds great.  I'd suggest something like `_framework/blazorconfig.json` since appconfig is a common enough name (yes I realize its a separate directory).

samples/HostedInAspNet.Client/Program.cs
rynowak: FYI for this part, I just copied what you had already. I'll include a separate BlazorServerSide sample in the next iteration

samples/HostedInAspNet.Server/Startup.cs
SteveSandersonMS: Oh I see. So I guess we have a tradeoff between doing things in the more traditional server-centric manner (changing `FetchData.cshtml` to do `new HttpClient`) or retaining the Blazor WebAssembly-style model (getting `HttpClient` from the service provider).

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
```diff
@@ -18,6 +19,11 @@ public class BrowserRenderer : Renderer, IDisposable
...
+        /// Notifies when a rendering exception occured.
+        /// </summary>
+        public event EventHandler<Exception> OnException;
```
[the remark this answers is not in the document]
rynowak: I think that this will be fine. As mentioned, I haven't done **anything** for error handling yet.

src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorCircuitsApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Temporary vague approximation to server-side execution needed so I can
+        /// build the rest of the interop.
+        /// </summary>  
+        public static IApplicationBuilder UseBlazorOnTheServer(
+            this IApplicationBuilder builder,
+            Action<BrowserRenderer> startupAction)
+        {
+            var endpoint = "/_blazor";
```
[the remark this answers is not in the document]
rynowak: Yes, that's the idea.

samples/HostedInAspNet.Client/wwwroot/index.html
SteveSandersonMS: Great, I'll do that.

src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorCircuitsApplicationBuilderExtensions.cs
rynowak: Yes, there are lots of other less coupled ways of doing this as well, but I didn't want to start designing something complicated. `CircuitFactory` is internal anyway and I don't anticipate it becoming a customization point.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/BlazorHub.cs
rynowak: yeah it might be slightly better

rynowak: The advantage of this is that it's readable in the debugger. I might just give it a slightly more gross looking name.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/BlazorHub.cs
rynowak: sure

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteUriHelper.cs
SteveSandersonMS: Update: I mean `Boot.Server.ts`.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/BlazorHub.cs
rynowak: this check isn't for optionality, it's in case the UriHelper was replaced. I'm also ok with saying that's not supported and just making it a hard cast.

On `enableNavigationInterception` - I totally agree, that would simplify things.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/Circuit.cs
rynowak: Anything that user-code needs to be able to get from the circuit. So that would be things like persistance, or lifecyle notifications.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
rynowak: That's the plan. As stated, I've made no effort to do that.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
rynowak: We could try something like that. I was planning to make it possible to access the Circuit via async-local, which accomplises the same thing - basically what you suggested in your comment here.

rynowak: The tricky part btw is what to do in the client - does it invoke a different static method for serversideblazor?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
rynowak: This should be running one. We can make this a dictionary

src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
rynowak: Good idea. I'll add that those, but make them depend on the accessor. 

To put something in services that depends on *data* you have have a way to initialize it.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
rynowak: It will be worth it if it's LOH

samples/HostedInAspNet.Server/Startup.cs
SteveSandersonMS: When we get Mono WebAssembly to work with `new HttpClient()` then I guess the solution becomes clear (stop using DI for that).

rynowak: Actually getting the client from the services is now the idiomatic thing to do in web app 👍

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
SteveSandersonMS: What if we enhance the JS interop mechanism so that, if you're trying to do a JSON-marshalled call to a .NET method that takes a parameter of type `IServiceProvider`, then for that parameter we automatically supply the one that corresponds to:

 * Your circuit, for server-side execution
 * The whole Blazor application, for WebAssembly execution

I don't think this would break any other scenario, since it's not possible to supply an `IServiceProvider` instance through JSON-marshalled calls in any other way. Once the .NET method receives the `IServiceProvider`, it can acquire the circuit or IJSRuntime or whatever else from it.

Or if we wanted to make it nicer still, we could look at supporting a `[FromServices]` attribute that you can put on `[JSInvokable]` methods' parameters, and then we supply whatever type of thing from your current service provider.

src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorCircuitsApplicationBuilderExtensions.cs
rynowak: Fixed

samples/HostedInAspNet.Server/Startup.cs
rynowak: Fixed

samples/HostedInAspNet.Server/Startup.cs
rynowak: I changed this to just use an service provided by the server side app. This is more idomatic for a server side app than making an http call to yourself, and demonstrates how these models differ.

rynowak:
Lots of updates. This is still a rough cut - see the top post for TODO list.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
rynowak: I'm tracking this as something to log an issue for follow up. Let me know if you think it needs to be addressed immediately.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
rynowak: I've written what I think the error handling code will look like, but it is not doing what I expect. I'll get some advice about this tomorrow.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/CircuitHost.cs
rynowak: I'm not totally sure that makes sense as a feature of interop writ-large. It seems awfully tied to our purposes. The other option would be to just have the JS-side call different .NET methods depending on the environment. 

But we do have the general problem right now where everything we need to to do via interop needs to go through statics. 

We can either address it on the JS side by calling different methods, or on the .NET side by building more magic-statics. If we have the JS side call different methods we still need magic statics to some degree.

This gets much better if we implement interop for instances.

rynowak: I'm going to start by enabling navigation interception earlier in the handshake and see where this leads me.

rynowak:
OK, so a few issues that are remaining.

0. I rewrote the URI helper classes to be much more simpler. If you don't like what I've done let's have a discussion about what you want instead 😆. It will be somewhat difficult to rationalize the current design approach of `BrowserUriHelper` with what we want from the server-side version, so we might have to live with duplication if we don't have any bright ideas.
0a. I haven't really written any unit tests for the new stuff, waiting for you to do a round of feedback first.
1. I haven't deferred `enableNavigationInterception`. I feel like the code is pretty simple without doing this. Let me know what you think.
2. **UH OH** We have timing issues now with selenium. I had to skip the 'fetch data' test because of the following.

```
            Browser.FindElement(By.LinkText("Fetch data")).Click();
            Assert.Equal("Weather forecast", Browser.FindElement(By.TagName("h1")).Text);
```

The 'click' is now processed synchronously. This means that it's possible for the second line to find the *old* state of the page before we get the new render frames.

Even worse! The reference returned by `FindElement` can be invalidated before we are able to access the `Text` property which will cause it to throw an exception.

The solution to this is to have each navigation that we care about in tests have a selector that will be more unique so that tests are more resilient. However this means we won't be able to convert our existing tests to server-side and have them 'just work' because they rely on synchrony or timing in some cases.

I haven't done any work yet to try and bulk-convert tests, but I'd like to defer that until after this PR.

samples/ServerSideBlazor.Server/Startup.cs
```diff
@@ -0,0 +1,37 @@
...
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            // Use component registrations from the client startup.
+            app.UseServerSideBlazor<Client.Startup>();
+
+            // Enables serving files from the client project.
+            app.UseBlazor<App>();
```
SteveSandersonMS: It would be super cool if it's possible to merge these two (`UseServerSideBlazor` and `UseBlazor`), so the developer only needs to call `app.UseServerSideBlazor<Client.Startup>()`.

That is, to have `UseServerSideBlazor<T>` also call `UseBlazor<T>` internally. I *think* that would work fine because `UseBlazor<T>` only uses the `T` to identify the target project, so it would be file for it to be given the `Client.Startup` type for that generic param.

Then the developer is mostly just choosing between `UseBlazor` and `UseServerSideBlazor` depending on which hosting model they want. Of course the two methods may take different optional config info, but the symmetry would still feel really clean.

Are there any use cases where the developer would need to configure `UseBlazor` separately from `UseServerSideBlazor`? Is it maybe for the case where the developer wants *both* server and client-side hosting models together, with the final selection being made by which `<script>` tag they have in `index.html`?

samples/ServerSideBlazor.Server/Startup.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class Startup
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            // Adds the Server-Side Blazor services, and those registered by the client startup.
+            services.AddServerSideBlazor<Client.Startup>();
+
+            // Since Blazor is running on the server, we can use an application service
+            // to read the forecast data.
+            services.AddSingleton<WeatherForecastService, DefaultWeatherForecastService>();
```
SteveSandersonMS: This is really interesting. It seems like this addresses two different matters:

1. It's a pattern for providing some abstraction around data access (or other services) so that you could make a single Blazor app that runs either server-side *or* client-side, and inject a hosting-model-specific implementation during startup.
2. It's a way of getting around the project reference going in the Server->Client direction. In some ways it would feel more natural here if the reference was Client->Server so that the client app could directly just consume some classes from the server.

Point (1) sounds good. Not too many devs will need that level of flexibility, but it's good that it's there if needed.

Point (2) is a bit more contentious. I suspect at least in the early days we're going to get a lot of devs saying "why do I need to add these extra interfaces/ABCs just to consume services in my Blazor client app? it would be easier if it was just one project not two". Maybe in the long run it might push us towards making it possible to do it all in a single project, and somehow dealing with the compile-time complexity this causes for us.

In the short term, for concern (2), perhaps we would say "if it's a service you only need in your Blazor UI, then put it in the Blazor project" and "if you need the service both in the Blazor UI *and* to expose as a traditional JSON web API, then put it in a `.Shared` project and reference it from both `.Server` and `.Client`".

What do you think?

src/Microsoft.AspNetCore.Blazor.Server/Microsoft.AspNetCore.Blazor.Server.csproj
```diff
@@ -1,10 +1,18 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\Microsoft.AspNetCore.Blazor.Browser\Hosting\ConventionBasedStartup.cs" Link="Hosting\ConventionBasedStartup.cs" />
+    <Compile Include="..\Microsoft.AspNetCore.Blazor.Browser\Hosting\IBlazorStartup.cs" Link="Hosting\IBlazorStartup.cs" />
+    <Compile Include="..\Microsoft.AspNetCore.Blazor.Browser\Services\BrowserUriHelperInterop.cs" Link="Circuits\BrowserUriHelperInterop.cs" />
```
SteveSandersonMS: Since this project references `Microsoft.AspNetCore.Blazor.Browser` anyway, should we use `[InternalsVisibleTo]` to avoid the need for these special shared sources?

src/Microsoft.AspNetCore.Blazor.Server/Microsoft.AspNetCore.Blazor.Server.csproj
```diff
@@ -14,7 +22,12 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Blazor.Browser\Microsoft.AspNetCore.Blazor.Browser.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Blazor\Microsoft.AspNetCore.Blazor.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Hosting\" />
```
SteveSandersonMS: Not needed I think

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Boot.Server.ts
```diff
@@ -38,8 +40,18 @@ function boot() {
...
 
-function unhandledError() {
-  err => console.error(err);
+function unhandledError(err) {
+  console.error(err);
+
+  // Disconnect on errors.
+  //
+  // TODO: it would be nice to have some kind of experience for what happens when you're
+  // trying to interact with an app that's disconnected.
```
SteveSandersonMS: For RTM I agree, but I think having errors show up in the console is fine for 0.5.0.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
```diff
@@ -1,9 +1,8 @@
-import { platform } from '../Environment';
-import { MethodHandle, System_String } from '../Platform/Platform';
-const registeredFunctionPrefix = 'Microsoft.AspNetCore.Blazor.Browser.Services.BrowserUriHelper';
-let notifyLocationChangedMethod: MethodHandle;
 let hasRegisteredEventListeners = false;
 
+// Will be initialized once someone registers
+let notifyLocationChangedCallback: { assemblyName: string, functionName: string } | null;
```
SteveSandersonMS: Probably won't make any difference, but there's some potential for developer confusion here, which is really the TypeScript compiler's fault. The variable is typed as `somestuff | null` but its actual value will be `undefined` until initalized (which is neither `somestuff` nor `null`). I don't think the compiler should be allowing this.

Confusion would occur if someone writes elsewhere:

```
if (notifyLocationChangedCallback !== null) {
    // Do something assuming that notifyLocationChangedCallback is a "somestuff"
    // but it could fail because notifyLocationChangedCallback===undefined
}
```

Suggested fix:

```
let notifyLocationChangedCallback: { assemblyName: string, functionName: string } | null = null;
```

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
```diff
@@ -53,11 +54,13 @@ function performInternalNavigation(absoluteInternalHref: string) {
...
 async function handleInternalNavigation() {
-  await DotNet.invokeMethodAsync(
-    'Microsoft.AspNetCore.Blazor.Browser',
-    'NotifyLocationChanged',
-    location.href
-  );
+  if (notifyLocationChangedCallback) {
+    await DotNet.invokeMethodAsync(
+      notifyLocationChangedCallback!.assemblyName,
+      notifyLocationChangedCallback!.functionName,
```
SteveSandersonMS: You don't need the `!` modifiers on lines 59 and 60, because the compiler will have narrowed the type automatically through flow analysis.

Inside the `if (notifyLocationChangedCallback) { }` block, the type of `notifyLocationChangedCallback` will not have the `| null` part, since the block wouldn't be entered if it was `null`.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
SteveSandersonMS: Tracking it as a future enhancement sounds fine to me.

samples/ServerSideBlazor.Server/Startup.cs
SteveSandersonMS: @vertonghenb No that's not correct. There is only one service provider, so you only register things once. The reason we support having `Startup.cs` in both the server host project and the Blazor library project is so you can register services that are internal to both projects if you want (and generally encapsulate concerns in either project if you want).

Example: If you have a `UserRepository`, you'd register it just once, and it's available both to Blazor components and to (say) MVC actions.

SteveSandersonMS:
> I rewrote the URI helper classes to be much more simpler. If you don't like what I've done let's have a discussion about what you want instead 

Having the base class looks good! One bit I wonder if we can simplify further is avoiding the registration of "notifylocationchanged" callbacks, perhaps as follows.

In general, I'd be surprised if we can get away without providing some way of accessing the `IServiceProvider` in incoming JS interop calls. People will have application-specific needs for this, e.g., to mutate a scoped `AppState` in some way when some JS event occurs. And if we do that, then the `IUriHelper` situation becomes simpler, as we can put `NotifyLocationChanged` as a `static` on the URI helper base class and always call that one, not needing to register the callback target any more. This `static` can then access the `IUriHelper` from the service provider, cast to `UriHelperBase` and invoke its `SetAbsoluteUri`/`TriggerOnLocationChanged`.

I tried spiking this up (having another magic asynclocal in the `.Blazor` project called `ServicesAccessor.CurrentServices`) and it worked out fine. I know this has the drawback that it gives people a way of hurting their unit testability by doing static service locator patterns where they should be using DI injection elsewhere in their code, but we could limit it to be `internal` if we really want.

If you want my implementation of this, please let me know and I'll append it to this PR. Don't want to do so without your say-so because I don't want to trample on your work and of course you're easily able to implement it too.

Another whole approach would be to implement instance method calls in JS interop. Again that's something we'll end up doing inevitably, but I'm reluctant to do it right now because we still have ongoing discussions with the Mono folks about precisely how this is going to work, and if we go forwards unilaterally we might make it harder to work that out.

> I haven't really written any unit tests for the new stuff, waiting for you to do a round of feedback first.

Sounds fine

> I haven't deferred enableNavigationInterception

It looks fine as-is. We might want to reduce the number of interop round trips involved in initializing this - currently it's 3 but could be reduced to 1 if the `enableNavigationInterception` returned the current uri/baseUri. I did that in the spike I mentioned above and it looks OK, even though it's kind of conflating theoretically separate things.

> UH OH We have timing issues now with selenium

Yeah that is awkward, but there's a good chance we can make the existing tests work out OK. I think Selenium has some way of retrying assertions briefly if they fail, as a way of dealing with this scenario. Sometimes it's still necessary to tweak how the tests are written, but we can do it.

Let's worry about getting the whole suite of E2E tests running on server rendering after this PR goes in, as you suggested.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/RenderBatch/OutOfProcessRenderBatch.ts
```diff
@@ -0,0 +1,241 @@
+import { RenderBatch, ArrayRange, RenderTreeDiff, ArrayValues, RenderTreeEdit, EditType, FrameType, RenderTreeFrame, RenderTreeDiffReader, RenderTreeFrameReader, RenderTreeEditReader, ArrayRangeReader, ArraySegmentReader, ArraySegment } from './RenderBatch';
+
+// TODO: Also support browsers that don't have TextDecoder (e.g., Edge)
```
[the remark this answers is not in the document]
rynowak: Logged #1132

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
rynowak: Logged #1132

samples/ServerSideBlazor.Server/Startup.cs
rynowak: I can't really think of a case where you're doing server-side Blazor, but don't want to also serve files from that project. 

The only issues I can think of are more philosophical. If you called it multiple times with different startups you could get a big stack of file-serving middleware.  If you wanted to control the position of the various middleware, you can't since it's a single call. These are some of the reasons why we don't typically put multiple middleware behind a single method call. 

But that can be an issue for future us.

samples/ServerSideBlazor.Server/Startup.cs
rynowak: > What do you think?

I think that if this is the case the implementations would be different in client and server. For instance, I didn't create a client-side implementation of this repository, but if I had, I'm sure you can imagine how it would be different.

src/Microsoft.AspNetCore.Blazor.Server/Microsoft.AspNetCore.Blazor.Server.csproj
rynowak: I think this is roughly the same lvel of complexity. If we're going to start using IVT I would have designed a few other things differently 😆

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
rynowak: Sure

rynowak:
the failures here are down to flakiness and timing. I'm going to merge this and see what I can do to harden these tests a bit as a follow up.

PR #1123, opened 2018-07-12
Make index.html static again

SteveSandersonMS:
Previously, we "built" the `index.html` file by transforming it during the build, replacing `<script type='blazor-boot'>` with a machine-generated tag that contained the app's boot config.

This worked fine, but had the significant drawback that you have no way of launching a Blazor app from any other HTML file. For example, you couldn't render an HTML page on the server and start up a Blazor app from it, because you wouldn't have any way of transforming the `<script type='blazor-boot'>`.

With this PR, `<script type='blazor-boot'>` is replaced by a plain old normal script tag (`<script src='_framework/blazor.webassembly.js'></script>` that you can put in any page you want, whether it's a static `index.html` file, or something you're generating at runtime on the server.

This is a breaking change, in that `<script type='blazor-boot'>` will no longer work. Developers need to swap it for the new tag mentioned above. We need to emphasise this in the upgrade steps in release notes.


ebekker:
Would the naming be a little bit more polished if changed from `_framework/blazor.webassembly.js` to something like `blazor-framework/blazor.webassembly.js` or even just `blazor-framework/blazor.js`?

SteveSandersonMS:
@ebekker We use the `_framework/` prefix as a namespace for all the framework-controlled resources. This is to ensure that those URLs never clash with anything the developer is trying to use for the application.

src/Microsoft.AspNetCore.Blazor.Build/Core/BootJsonWriter.cs
```diff
@@ -0,0 +1,97 @@
...
+                entryPoint,
+                assemblyReferences,
+                embeddedContent,
+                linkerEnabled);
+            return Json.Serialize(data);
+        }
+
+        private static string GetAssemblyEntryPoint(string assemblyPath)
+        {
+            using (var assemblyDefinition = AssemblyDefinition.ReadAssembly(assemblyPath))
```
rynowak: meta: do we need cecil for this? is the reason why we need to do this because otherwise we'd need to load the assembly?

src/Microsoft.AspNetCore.Blazor.Build/Core/BootJsonWriter.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    throw new ArgumentException($"The assembly at {assemblyPath} has no specified entry point.");
+                }
+
+                return $"{entryPoint.DeclaringType.FullName}::{entryPoint.Name}";
+            }
+        }
+
+        private static string Normalize(string outputPath) =>
+            Path.Combine(Path.GetDirectoryName(outputPath), Path.GetFileName(outputPath).ToLowerInvariant());
```
rynowak: Not sure if this is part of an earlier version, but how does this work on a case-sensitive filesystem?

SteveSandersonMS: That's a very good point! It looks like it *wouldn't* work on a case-sensitive filename.

I retained this logic from the earlier `IndexHtmlWriter`. I could try to ask why the `Normalize` step was added there (wasn't me) but TBH it looks like it might just have been for aesthetic reasons. Without it, the build works fine but logs lines like this:

```
Writing boot data to: C:\Users\username\Some\Dir\StandaloneApp\obj\Debug\netstandard2.0\blazor/blazor.boot.json
```

... so maybe it was just to deal with the weirdness of the slashes.

I've now replaced the `Normalize` method with just `Path.GetFullPath(outputPath)`. This corrects the slashes (and would resolve relative paths or directory traversal segments) without interfering with anything else.

src/Microsoft.AspNetCore.Blazor.Build/Core/BootJsonWriter.cs
rynowak: Yeah, loading assemblies in tooling, or at build time is just a fail. We shouldn't try. The `System.Reflection.Metadata` APIs are pretty low level and hard to use, hence Mono.Cecil still is the best option for many of these kinds of tasks.

PR #1128, opened 2018-07-13
Implementation of JS->.NET interop instance method calls

SteveSandersonMS:
This is in case we decide the best way to deal with accessing services from JS calls in #1116 is by having support for calling instance methods.

@rynowak Would be interested in your preference on whether to use this approach, or the `ServicesAccessor.CurrentServices` asynclocal approach.


SteveSandersonMS:
Note that this only allows for instance method calls when going JS -> .NET. It does handle both sync and async cases, which it really just gets for free from the underlying support with static method calls.

There is some danger of memory leaks with this. If the developer doesn't remember to call `.dispose` on the object handle they receive in JS (or if a visitor chooses to modify the JS running in their browser so it doesn't call `dispose`), then the .NET side will be leaking. One mitigation is that it's contained within the `JSRuntimeBase`, so in the server case, we still clean up everything held by that runtime instance when the user disconnects. But that doesn't take away the short-term leak.

galvesribeiro:
Finally! 👏

src/Microsoft.JSInterop/DotNetDispatcher.cs
```diff
@@ -32,7 +32,7 @@ public static string Invoke(string assemblyName, string methodIdentifier, string
             // the targeted method has [JSInvokable]. It is not itself subject to that restriction,
             // because there would be nobody to police that. This method *is* the police.
 
-            var syncResult = InvokeSynchronously(assemblyName, methodIdentifier, argsJson);
+            var syncResult = InvokeSynchronously(assemblyName, methodIdentifier, /* targetInstance */ null, argsJson);
```
rynowak: if you like this you can just named parameters 😆

rynowak:
I have a high level idea about this that I think is pretty S P I C Y.

1. We merge the server-side blazor PR as-is because I doesn't depend on these details
2. We add instance-based interop (this PR)
3. We create a first class js API for interacting with the service provider.

So this means you can write JS code like `blazor.services.getService("my.service.type")`.

We still need to a mechanism for an 'arbitrary caller' to get the service provider, but we make it internal and use it to build this mechanism. Then we push people towards instance-based interop. Thoughts?

src/Microsoft.JSInterop/DotNetDispatcher.cs
SteveSandersonMS: Yeah, I just always feel weird about it when it's not the final param, because then you have to name the rest even when their names are obvious. But still maybe that's better.

SteveSandersonMS:
Maybe, I don't know. It introduces some extra complications, e.g. having to specify the fully-qualified type name, which might involve generic args and would need to include the assembly name as well if we're to avoid scanning all loaded assemblies.

We would still need to have the same APIs for calling instance methods on arbitrary objects, because although the "call methods on DI services" use case is covered without it, other use cases exist. For example, the developer has a "todo list" where they want to invoke methods on specific todo items when a JS event occurs.

> We merge the server-side blazor PR as-is because it doesn't depend on these details

I'm fully in favour of merging #1116 using its existing solution and not waiting to solve this other stuff. We can figure out how to streamline it further later (e.g., using instance-method interop).

src/Microsoft.JSInterop/DotNetObject.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.JSInterop
+{
+    /// <summary>
+    /// Wraps a JS interop argument, indicating that the value should not be serialized as JSON
+    /// but instead should be passed as a reference.
+    ///
+    /// To avoid leaking memory, the reference must later be disposed by JS code.
+    /// </summary>
+    public struct DotNetObject
```
rynowak: Is there a benefit to making this a struct? It looks like it's always going to be boxed.

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
```diff
@@ -231,4 +231,48 @@ module DotNet {
...
+        methodIdentifier,
+        JSON.stringify(args));
+    }
+
+    public dispose() {
+      const promise = invokeMethodAsync<any>(
+        'Microsoft.JSInterop',
+        'DotNetDispatcher.ReleaseDotNetObject',
+        this._id);
+      promise.catch(error => console.error(error));
```
rynowak: Should we have some kind of global sink for errors that js swallows? I'm sure ultimately they would all end up in the console, but I'm wondering if its worth centralizing?

src/Microsoft.JSInterop/DotNetDispatcher.cs
```diff
@@ -129,6 +129,51 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
         public static void EndInvoke(long asyncHandle, bool succeeded, object resultOrException)
             => ((JSRuntimeBase)JSRuntime.Current).EndInvokeJS(asyncHandle, succeeded, resultOrException);
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="dotNetObjectId"></param>
+        /// <param name="methodIdentifier"></param>
+        /// <param name="argsJson"></param>
+        /// <returns></returns>
```
rynowak: Needs actual docs 😆

rynowak:
So I think regardless of how we resolve the looming issues with the service provider, we'll want this change. We would do it eventually, it might as well be now.

src/Microsoft.JSInterop/DotNetObject.cs
SteveSandersonMS: It will always be boxed when it's being passed to `TrySerializeKnownTypes`, but doesn't have to be boxed at other times.

Originally I was going to make it a reference type because then we could preserve the same box, but since it turns out we're not storing `DotNetObject` anywhere, it was not clear that there'd be any advantage.

Overall I'm not seeing any case where it would make any difference either way, so I erred on the side of not forcing extra heap objects if there are any cases where developers want to make, say, an array or list of them.

If you feel there's any reason to prefer having it as a reference type I'm happy to change it.

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
SteveSandersonMS: I'm not sure. A developer can already register their own top-level uncaught exception handler if they want to (`window.onerror`). If it turns out there are reasons for adding further APIs around this we can do so.

src/Microsoft.JSInterop/DotNetDispatcher.cs
SteveSandersonMS: Yes, and tests! I only had an hour to put this PR together earlier and wanted it to show up before you started your work today.

src/Microsoft.JSInterop/JavaScriptRuntime/src/Microsoft.JSInterop.ts
SteveSandersonMS: Actually I think code here is wrong, and should do something like:

    promise.catch(error => throw error);

.... to make sure it does hit `window.onerror`. I might be misremembering the details so will have to actually try it out.

SteveSandersonMS:
@rynowak In retrospect, my "quick" implementation was little more than a proof of concept. After implementing all the cases (passing .NET objects by reference in args and return values in both sync and async calls to instance and static methods in both directions) and adding tests, there's a *lot* more going on. Please feel free to re-review if you want, or trust that the tests are extensive enough, or let me know if there's anything you prefer to discuss.

src/Microsoft.JSInterop/DotNetDispatcher.cs
```diff
@@ -101,16 +126,26 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
+            var serializerStrategy = runtime.ArgSerializerStrategy;
             for (var i = 0; i < suppliedArgsLength; i++)
             {
-                suppliedArgs[i] = serializerStrategy.DeserializeObject(
-                    suppliedArgs[i], parameterTypes[i]);
+                if (parameterTypes[i] == typeof(JSAsyncCallResult))
+                {
+                    // For JS async call results, we have to defer the deserialization until
+                    // later when we know what type it's meant to be deserialized as
+                    suppliedArgs[i] = new JSAsyncCallResult(suppliedArgs[i]);
```
rynowak: I'm not totally sure I grok what's going on with this type. How is this different from using `object` ? What are the scenarios where you'd use this? Do we expect library code to need this?

src/Microsoft.JSInterop/DotNetObjectRef.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Microsoft.JSInterop
+{
+    /// <summary>
+    /// Wraps a JS interop argument, indicating that the value should not be serialized as JSON
+    /// but instead should be passed as a reference.
+    ///
+    /// To avoid leaking memory, the reference must later be disposed by JS code or by .NET code.
+    /// </summary>
+    public class DotNetObjectRef : IDisposable
```
rynowak: Should this have a finalizer?

Is this class just infrastructure? I'm not sure I understand yet how this is used

src/Microsoft.JSInterop/DotNetDispatcher.cs
SteveSandersonMS: A fair question - it was definitely not made clear in the code. Hopefuly I've now clarified this in 233c392, but please let me know if you think that still doesn't suffice.

src/Microsoft.JSInterop/DotNetObjectRef.cs
rynowak: Yeah I supposed the finalizer wouldn't help in this case because its rooted by the runtime

src/Microsoft.JSInterop/JSAsyncCallResult.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.JSInterop
+{
+    /// <summary>
+    /// Represents an intermediate value while handling an async call to JavaScript.
+    /// </summary>
+    public readonly struct JSAsyncCallResult
```
rynowak: What's the benefit from making this a struct? won't it always be boxed

SteveSandersonMS: Good point. Changed it.

SteveSandersonMS:
Merged via #1153

PR #1139, opened 2018-07-16
Fix Blazor tooling on 15.8

rynowak:
/cc @SteveSandersonMS

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/ITagHelper.cs
+++ b/src/Microsoft.AspNetCore.Blazor/ITagHelper.cs
@@ -1,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Blazor/ITagHelper.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.TagHelpers
+{
+    // This is an unfortunate workaround due to https://github.com/aspnet/Razor/issues/2482
+    // The Razor tooling looks for a type with exactly this name and will prevent tag helper
+    // discovery if it is not found.
+    internal class ITagHelper
```
rynowak: this is internal so it shouldn't be visible to anyone.

SteveSandersonMS:
@rynowak Thanks for sorting this out!

PR #1146, opened 2018-07-18
Add HTML Block rewriting

rynowak:
/cc @SteveSandersonMS - handing this off for you to do the runtime piece.

Rewrites HTML-only subtrees of the IR in an HTML 'block' node with simple string content. 

This will rely on new runtime/rendertree support in order to work E2E.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/MarkupString.cs
+++ b/src/Microsoft.AspNetCore.Blazor/MarkupString.cs
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-
 namespace Microsoft.AspNetCore.Blazor
 {
     /// <summary>
@@ -10,16 +8,19 @@ namespace Microsoft.AspNetCore.Blazor
     /// </summary>
     public struct MarkupString
     {
-        private readonly string _value;
-
         /// <summary>
         /// Constructs an instance of <see cref="MarkupString"/>.
         /// </summary>
         /// <param name="value">The value for the new instance.</param>
         public MarkupString(string value)
         {
-            _value = value ?? string.Empty;
+            Value = value;
         }
+
+        /// <summary>
+        /// Gets the value of the <see cref="MarkupString"/>.
+        /// </summary>
+        public string Value { get; }
 
         /// <summary>
         /// Casts a <see cref="string"/> to a <see cref="MarkupString"/>.
@@ -30,6 +31,6 @@ namespace Microsoft.AspNetCore.Blazor
 
         /// <inheritdoc />
         public override string ToString()
-            => _value;
+            => Value ?? string.Empty;
     }
 }
--- a/src/Microsoft.AspNetCore.Blazor/MarkupString.cs
+++ b/src/Microsoft.AspNetCore.Blazor/MarkupString.cs
@@ -1,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Blazor
+{
+    /// <summary>
+    /// A string value that can be rendered as markup such as HTML.
+    /// </summary>
+    public struct MarkupString
+    {
+        private readonly string _value;
+
+        /// <summary>
+        /// Constructs an instance of <see cref="MarkupString"/>.
+        /// </summary>
+        /// <param name="value">The value for the new instance.</param>
+        public MarkupString(string value)
+        {
+            _value = value ?? string.Empty;
+        }
+
+        /// <summary>
+        /// Casts a <see cref="string"/> to a <see cref="MarkupString"/>.
+        /// </summary>
+        /// <param name="value">The <see cref="string"/> value.</param>
+        public static explicit operator MarkupString(string value)
+            => new MarkupString(value);
+
+        /// <inheritdoc />
+        public override string ToString()
+            => _value;
+    }
+}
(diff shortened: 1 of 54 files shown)


rynowak:
FYI the test failures are of course expected right now. The rendering tests throw because the `RenderTreeBuilder` throws.

If you rename the new method on `RenderTreeBuilder` you'll have to regen the baselines for the runtime codegen tests.

SteveSandersonMS:
Thanks @rynowak! This looks perfect.

I've rebased on master (hence the changes to the commit history here now) and will proceed with the runtime bits.

SteveSandersonMS:
Still to do:

* Support `UpdateMarkup` edit frames in browser-side code
* Stop assuming that markup frames content will represent exactly 1 HTML element. Needs to support arbitrary ranges of markup.

Both are needed for `IHtmlString` support.

SteveSandersonMS:
OK all done now.

@rynowak We should discuss the existence of the new `MarkupString` type (and why I didn't use `HtmlString` or `IHtmlContent` from the HTML abstractions package), but other than that it should be ready to go.

src/Microsoft.AspNetCore.Blazor/MarkupString.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        private readonly string _value;
+
+        /// <summary>
+        /// Constructs an instance of <see cref="MarkupString"/>.
+        /// </summary>
+        /// <param name="value">The value for the new instance.</param>
+        public MarkupString(string value)
+        {
+            _value = value ?? string.Empty;
```
rynowak: Why this? Is it ever meaningful to have null markup string?

For instance the following is true. `((MarkupString)null).ToString() == ""`

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
```diff
@@ -68,13 +68,19 @@ public void CloseElement()
...
-        public void AddMarkupContent(int sequence, string textContent) =>
-            // TODO stevesa implement me
-            throw null;
+        /// <param name="markupContent">Content for the new markup frame.</param>
+        public void AddMarkupContent(int sequence, string markupContent)
+        {
+            if (markupContent == null)
+            {
+                throw new ArgumentNullException(nameof(markupContent));
+            }
```
rynowak: Should we check non-null here? `new MarkupContent(string)` is happy to accept null. IMO we should permit both or neither

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
```diff
@@ -96,9 +96,10 @@ public void SupportsElementsAsStaticBlock()
         [Fact]
         public void CreatesSeparateMarkupFrameForEachTopLevelStaticElement()
         {
-            // The JavaScript-side rendering code relies on markup blocks representing
-            // exactly one self-contained element. This test is to show that we don't
-            // coalesce together markup blocks that happen to be next to each other.
+            // The JavaScript-side rendering code does not rely on this behavior. It supports
+            // inserting markup frames with arbitrary markup (e.g., multiple top-level elements
+            // or none). This test exists only as an observation of the current behavior rather
+            // than a promise that we never want to change it.
```
rynowak: Do you think it's desirable that we change this in the compiler? It would be straightforward to do.

rynowak:
> We should discuss the existence of the new MarkupString type (and why I didn't use HtmlString or IHtmlContent from the HTML abstractions package), but other than that it should be ready to go.

My assumption is that you wanted to avoid the allocation and the extra references. In particular I think WebEncoders is pretty heavy-weight - this is why I didn't suggest it from the beginning.

rynowak:
:shipit: from me. I'm not allowed to make it approved since I created the PR.

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
SteveSandersonMS: Perf-wise yes, the ideal would be for adjacent literal markup frames (and any adjacent literal text frames) to be coalesced. 

If you think this is a simple addition that would be neat, but I think it’s also a good improvement as-is.

Note that we shouldn’t treat text literals as markup unless they are coalesced into a markup frame, because text nodes on their own are much cheaper to render than markup frames.

SteveSandersonMS:
> My assumption is that you wanted to avoid the allocation and the extra references. In particular I think WebEncoders is pretty heavy-weight - this is why I didn't suggest it from the beginning.

Yes, avoiding unnecessary allocations was one reason, the other was that simply referencing `Microsoft.AspNetCore.Html.Abstractions` suddenly exposes a bunch of new APIs that aren't really applicable. Finally the name `HtmlString` implies the markup has to be HTML, which is not the case for this feature - SVG is also supported.

src/Microsoft.AspNetCore.Blazor/MarkupString.cs
SteveSandersonMS: Good point.

I've now changed it so that `MarkupString` can hold the value `null`.

 * If you `.ToString()` such an instance, you get `string.Empty` just like HtmlAbstractions' `HtmlString`.
 * If you `renderTreeBuilder.AddContent` such an instance, it creates a markup frame whose value is `string.Empty`, just like if you `AddContent((string)null)`.

So the net effect is that nulls continue to be treated like `string.Empty`, but now it's possible to observe that `null` was stored in the `MarkupString` to avoid losing that info.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
SteveSandersonMS: Good point. It now just null-coalesces with `string.Empty` like we do for regular `string` values.

PR #1150, opened 2018-07-18
Polyfill for TextDecoder

### LINKED ISSUE #1131: Address TODO: Also support browsers that don't have TextDecoder (e.g., Edge)
Follow up item from #1116

This is a TODO in the Blazor server boot script. We use TextDecoder where it's available, but don't have any support yet for browsers without it.

Suchiman:
I was toying around with server side blazor and the first culprit i run into was the lack of Edge support 😉 
I've extracted the least amount of Polyfil necessary from https://github.com/samthor/fast-text-encoding/blob/master/text.js#L137 which worked for my experiments and still uses the native TextDecoder in other browsers.
Not sure if this is acceptable or how / where i need to attribute this.

Fixes #1131


Suchiman:
Can it be that the tests are flaky? I pushed, got 2 failed tests, ran it locally, got one failed test, tested the current master HEAD locally, tests passed, reapplied my patch, tests passed again, force pushed a comment change to retrigger CI, got 4 test failures?!

SteveSandersonMS:
Thanks @Suchiman for suggesting this.

Yes, you're right: the E2E tests are currently flaky since we put in the ones for server execution. This is because the browser UI updates occur asynchronously in this mode, and not all the tests allow for that. In #1153 I'm working on making them all robust.

Your selection of polyfill here looks good. However we do need to do something about attribution, which I will figure out. Also I think we'll probably use this polyfill internally in the Blazor library and not attach it globally to `window` because we don't want to interfere with any other polyfill the developer is trying to use. So if it's OK with you, I will reorganize some aspects of this PR before I merge it!

Suchiman:
> So if it's OK with you, I will reorganize some aspects of this PR before I merge it!

Of course

SteveSandersonMS:
As discussed, moving this over to #1168 so I can make some additions.

PR #1153, opened 2018-07-19
Run E2E tests under both WebAssembly and server-side execution modes

SteveSandersonMS:


SteveSandersonMS:
This PR is based on the as-yet-unmerged #1128, because it depends on a couple of bugfixes in that PR. That's why it says I've modified 50+ files. In reality this PR will be more compact once #1128 is merged into `master`. So if you want to see what's really changed, look at the commits *after* the first one here.

SteveSandersonMS:
@rynowak I'm duplicating your comments here because, since I rebased, GitHub considers them out of date:

> [About JSAsyncCallResult] I'm not totally sure I grok what's going on with this type. How is this different from using object ? What are the scenarios where you'd use this? Do we expect library code to need this?

A fair question - it was definitely not made clear in the code. Hopefuly I've now clarified this in 233c3929cf7c036c8ddf350f908b448ec58427d4, but please let me know if you think that still doesn't suffice.

> [About DotNetObjectRef] Should this have a finalizer? Is this class just infrastructure? I'm not sure I understand yet how this is used

This class is the public API for indicating that you want to pass something by reference, rather than by JSON serialization. Any .NET object can be wrapped in `DotNetObjectRef`, and then when it's used in a JS interop call as an arg or return value, the JS side receives a handle that it can later use to invoke instance methods on, or to pass back to .NET as JS interop args or return values in the opposite direction.

I don't think it would benefit from having a finalizer, because it doesn't have any non-.NET resources to release. It's explictly rooted in the `JSRuntimeBase` until you `.Dispose()` it (either from .NET code or from the JS side).

Later, if we have a symmetrical equivalent `JSObjectRef` counterpart, that *would* have a finalizer so that we could automatically notify the JS side that it can stop tracking that instance if the .NET side forgets about it.

And if JS had a concept of a finalizer, we would use it to trigger disposal of `DotNetObjectRef` instances. But JS has no such concept, so it's entirely the developer's own responsibility to dispose explicitly.

SteveSandersonMS:
Here's how this PR attempts to deal with the inherent timing issues in the server-execution E2E tests:

* Have enabled Selenium's *implicit wait* feature with a timeout of 1 second
    * This means that whenever the test code does a `FindElement`, Selenium attempts it immediately, and if there is no matching element, it waits for 1 second then tries again.
    * This doesn't slow the code if it was going to succeed anyway, but does slow it a little in cases where we do need to wait. It isn't polling - it's literally just waiting 1 second.
    * This approach deals with a lot of cases where we perform some action then try to assert about an element that's appeared because of the action
* Have defined some new assertion methods on a static class called `WaitAssert`
    * These look like Xunit assertions (e.g., instead of `Assert.Contains`, we have `WaitAssert.Contains`), but work by polling until the assertion becomes true.
    * They take a `Func<T>` instead of a `T` to define the "actual" value, because the polling logic has to keep evaluating the "actual" until it obtains a passing value
    * If it times out (after 1 second) then it throws the same exception that the underlying Xunit assertion does.
    * The idea is to use `WaitAssert` assertions immediately after actions such as clicks and navigations because it's possible that the new value will appear after an observable delay. It's *not* necessary to use `WaitAssert` for all assertions - only immediately after some action that mutates the UI.

I don't know yet how much of a long tail of instability issues we're going to have to deal with. I'm confident that any individual E2E test can be made to work robustly using the above two techniques, but it remains to be seen how many will need to be edited.

I hope to merge this PR as soon as I've made enough E2E test tweaks to get a few green runs in a row.

Issue #1069, opened 2018-07-02
Combine MVC razor pages with Blazor pages


floreseken:
I guess lots of us are now developing MVC applications and at some point want to convert those to Blazor.

I think I would be convenient if Blazor supports some sort of migration path by making it possible to combine Blazor and MVC pages. This will make it possible to gradually migrate your app, page by page.

Currently if there is no route registered with a Blazor component, Blazor will throw an exception, if the router would then try to just load the url I think we are almost there.. Would this be feasible?

I've got some sort of workaround running with iframes in a razorpage loading the MVC page, but that just feels dirty..

All ideas welcome..

GoranHalvarsson:
Yes, good point.

Andrzej-W:
I was also thinking about adding Blazor to my ASP.NET Core MVC application but who should be responsible for rendering common page elements like main application menu, footer, etc.? Probably we have to do this in MVC and Blazor. I think in non trivial web application serious refactoring will be necessary to reuse the same code in MVC and Blazor. There is yet another problem - application state. We should save/restore state before/after each transition between both technologies. Currently Blazor is not good for SEO - it is not acceptable for most people.

I'm not sure I still want to combine these two technologies in one application. Instead I would like to have server side rendering #24  (do not confuse this with Blazor working on the server as demonstrated recently by Steve!)

floreseken:
Ok. I think I figured this out. If you doe this in your startup:

` app.Map("/blazor", child => { child.UseBlazor<Client.Program>(); });`

and modify your blazor `index.html` like this:

`    <base href="/blazor/" />`

It works like I intend, I'm able to link and navigate from blazor to mvc pages that not reside in the /blazor folder.
I'll have to do more testing if everything works.. but I guess so..

vertonghenb:
@floreseken  any update on the testing of this?

floreseken:
Yes, works fine.

monu3012:
please tell me how to use blazor page in asp.net with mvc

tomRedox:
@monu3012 I wrote up how to get server-side Blazor components working on an MVC app here https://stackoverflow.com/a/56245667/1879019, which might give you some ideas.

Issue #979, opened 2018-06-12
Add UrlHelper.NavigateTo overload that forces a full page reload


SteveSandersonMS:
Inspired by the scenario in #962.

If you want to trigger a file download by navigating to an endpoint where the server will return something with a `Content-Disposition` header, or if you somehow are combining multiple client-side apps, you'd sometimes want navigation to bypass the usual "internal navigation" logic.

vertonghenb:
I'm thinking about submitting a PR for this.

I believe that overloading the `NavigateTo` method with a `bool isWithinBaseUriSpace` (defaults to `true`) should suffice. Right road to take?

Issue #962, opened 2018-06-08
Handle FileResult from controller service


Paradoxon101:
Currently there is no way to evaluate a FileResult in the browser or to save it via a download dialog.

SteveSandersonMS:
I'm not certain what you're looking for.

If you're trying to trigger a file download with a "Save As" dialog, that's normally handled by server-side code. Specifically, if the server returns the `Content-Disposition` header, this makes the browser show a "Save As" dialog. Also you need to make the browser perform a normal navigation to the URL that returns this content, rather than by using `HttpClient`.

Paradoxon101:
That's exactly what I wanted to do.

Server:
        [HttpGet]
        [Route("download/{downloadId}")]
        public FileResult Download(int downloadId)
        {
            return this.File(bytes, "application/zip", fileName);
        }
		
Client:
@functions {
    void Download()
    {
        UriHelper.NavigateTo("/download/1");
    }
}

Could the bug be in the UriHelper?

chucker:
What exactly is the bug? Does the browser navigate, but then not ask the user to download? Is `Content-Disposition` being set in the response?

Paradoxon101:
I used this on server:
System.Net.Mime.ContentDisposition cd = new System.Net.Mime.ContentDisposition
            {
                FileName = fileName,
                Inline = false  // false = prompt the user for downloading;  true = browser to try to show the file inline
            };

The browser displays the new url in the address bar, but nothing happens. If I now put the focus in the address line and press Enter, the download starts.

SteveSandersonMS:
`UriHelper.NavigateTo` performs internal navigation if you point it to a URL that could be handled by client-side routing. It doesn't know that you wanted the browser to actually do a server-side navigation in this specific case.

To cover this use case, I've filed #979.

In the meantime, as a workaround you could use JS interop to trigger a full-page load to the target URL. The JavaScript code would be something like `location.href = 'download/123';`.

vertonghenb:
@SteveSandersonMS True, but what if the Route `download/123` is protected by the `[Authorize]` attribute and needs a `HTTP Header` like `Authorization Bearer YOUR_TOKEN`

dan-drews:
@vertonghenb Were you able to solve this problem? I am running into the same issue. Entirely API driven application that generates a PDF. I am using a bearer token for this.

smartprogrammer93:
[...]

PR #1154, opened 2018-07-19
Overload UriHelper to forceLoad the page even if it's not a Blazor defined Route.

vertonghenb:
**PR issued to solve**
* Add UrlHelper.NavigateTo overload that forces a full page reload #979
* Handle FileResult from controller service #962 
* Combine MVC razor pages with Blazor pages #1069 
* Helps with Routing enhancements #293

Basically when you have a MVC Server that serves the Blazor Client and on the other hand sends File Downloads (for example). It was not possible to load the MVC Uri outside of Blazor's internal routing. 
With this PR, it's now possible to use the following block of code to load a Uri outside of Blazor's routing:

```csharp
@inject Microsoft.AspNetCore.Blazor.Services.IUriHelper UriHelper
<button onclick=@GoToMvcRoute>Navigate to Controller Action!</button>
<button onclick=@DontGoToMvcRoute>Don't Navigate to Controller Action!</button>

@functions {
    void GoToMvcRoute()
    {
        UriHelper.NavigateTo("/controller/action", forceLoad: true);

    }
    void DontGoToMvcRoute()
    {
        UriHelper.NavigateTo("/controller/action");

    }
}
````

**Tests**
Not really sure how to test this in the current suite, maybe add an additional AspNetHosted project with 
a server that also serves via MVC? I simply tested on "my machine" to make sure it's loading the page. 

**Caveats**
The mapping of the controllers still have to happen on Startup.cs (which is normal):
````csharp
            app.UseMvcWithDefaultRoute();
            app.UseBlazor<Client.Program>();
````
**Impact**
No impact on current usages of IUriHelper (default parameter)

If you know a better name for the parameter, I'm all ears! 
* byPassInternalRouting?

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
@@ -40,11 +40,11 @@ function enableNavigationInterception(assemblyName: string, functionName: string
 }
 
 export function navigateTo(uri: string, forceLoad: boolean) {
+  const absoluteUri = toAbsoluteUri(uri);
+
   if (forceLoad) {
     location.href = uri;
-  }
-  const absoluteUri = toAbsoluteUri(uri);
-  if (isWithinBaseUriSpace(absoluteUri)) {
+  } else if (isWithinBaseUriSpace(absoluteUri)) {
     performInternalNavigation(absoluteUri);
   } else {
     location.href = uri;
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
@@ -39,7 +39,10 @@ function enableNavigationInterception(assemblyName: string, functionName: string
   window.addEventListener('popstate', handleInternalNavigation);
 }
 
-export function navigateTo(uri: string) {
+export function navigateTo(uri: string, forceLoad: boolean) {
+  if (forceLoad) {
+    location.href = uri;
+  }
   const absoluteUri = toAbsoluteUri(uri);
   if (isWithinBaseUriSpace(absoluteUri)) {
     performInternalNavigation(absoluteUri);
--- a/src/Microsoft.AspNetCore.Blazor.Browser/Services/BrowserUriHelper.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Browser/Services/BrowserUriHelper.cs
@@ -47,14 +47,14 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Services
         }
 
         /// <inheritdoc />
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by chucker** [comment]
```csharp
/// <param name="forceLoad">Indicator to force load the URI, even if it's not a blazor route.</param>
```

I definitely find `forceLoad` too vague.

A few candidates:

* `skipRouteValidation`? I find it clearer, if a little awkward. (Sort of similar to your suggestion of `byPassInternalRouting`.)
* `useJavaScriptRouting`? This is perhaps a little misleading, but makes it clearer that 1) the normal routing has nothing to do with JS (and implicitly with Blazor instead), and 2) this is the traditional browser-side stuff.

kiegh:
Not to be nitpicky, but if "byPassInternalRouting" is used, could we correctly name it "bypassInternalRouting" bypass is one word.

Andrzej-W:
Where can we use enhancement like this? Probably there are two places: `UriHelper.NavigateTo` and HTML `<a>` element. If so, the best name in my opinion is `serverSideUrl` or simply `serverUrl`.

For completeness `NavLink` component `Blazor\src\Microsoft.AspNetCore.Blazor\Routing\NavLink.cs` should be enhanced also. Then we need some changes in Blazor router. Unfortunately I have no idea what have to be changed and where exactly.

vertonghenb:
I was also thinking to simply not use the `NavigateTo `and create a new property
`Location` with a getter and setter which triggers a JS.Interop call like: 
```js
location.href = uri;
```

This is basically what we're doing setting the location, since it is a hard reload of the page.

vertonghenb:
@SteveSandersonMS any opinion on this? So I can prepare it for the 0.5.0 is you'd like.

SteveSandersonMS:
@vertonghenb Thanks for contributing this! We won't get it in in time for 0.5.0 (there's quite a bit to review and some API decisions to make) but we should be able to merge to `master` soon after 0.5.0 ships. Hope that's OK!

vertonghenb:
Sure, if the API decision is made, i'll assist further.

andras-ferencz:
From my personal experience, `location.href = "xyz";`, depending on the browser, build, security settings, if you're running on localhost or production, and how pissed off the cosmic space gnomes are at you on a particular day will occasionally outright not work. Using `location.reload()` might be a better cross-experience solution like so.

```HTML
<a onclick=@(()=>Navigate("/login")) href="#">Login</a>
```
```js

        window.Refresh = (a) => {
            location.reload();
            return true;
        }
```
```C#
async void Navigate(string uri)
{
    UriHelper.NavigateTo(uri);
    await JSRuntime.Current.InvokeAsync<bool>("Refresh");
}
```

Paradoxon101:
Anything new yet? Will it be included in version 0.7.0?

SteveSandersonMS:
Thanks again for this, @vertonghenb! It's now moved to https://github.com/aspnet/AspNetCore/pull/4786, where it will be merged shortly.

PR #1157, opened 2018-07-19
Fixed exporting platform to window object

thild:
Recently versions are not exporting platform to window object correctly.
When GlobalExports.ts gets called, platform from Enviroment.ts is undefined and gets set only when Environment.setPlatform(monoPlatform) is called on boot.


SteveSandersonMS:
Thanks for the suggestion!

I've restored the older behavior in a simpler way here: https://github.com/aspnet/Blazor/commit/9a4951f1e6b7e4a77ecae04cdac8b4b129b5f9ac

Long term, I hope people won't be relying on `Blazor.platform` very often. The newer JS interop APIs are no longer on `Blazor.platform`. They are now on `DotNet` (e.g., `DotNet.invokeMethodAsync`), as will be covered in updated docs.

PR #1196, opened 2018-07-26
fix dispose components & EventHandler in OutOfProcess Render

uazo:


uazo:
You are welcome.
Same issue as #1071  #1072 (identical, really identical!)

danroth27:
@uazo Could you please provide simple code that reproduces the issue? We want to make sure we understand what is getting fixed here.

SteveSandersonMS:
Thanks again! This is now moved to https://github.com/aspnet/AspNetCore/pull/4596 and will be merged imminently.

Issue #1203, opened 2018-07-26
Bootstrap menu not rendering properly in Blazor 0.5


MarkStega:
The following code in my navmenu.cshtml fails to render properly. Attached are two images. navbar-properlyrendered.png & navbar-blazorrendered.png show the odd styling (notice the layout of the three links) in Blazor. 

The proper rendering is in a static html page with the bootstrap css/js imports. The Blazor rendering is from a project derived from the template produced Blazor client.

```html
<nav class="navbar navbar-expand-md bg-dark navbar-dark">
    <a class="navbar-brand" href="#">Navbar</a>
    <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#collapsibleNavbar">
        <span class="navbar-toggler-icon"></span>
    </button>
    <div class="collapse navbar-collapse" id="collapsibleNavbar">
        <ul class="navbar-nav">
            <li class="nav-item">
                <a class="nav-link" href="#">Link</a>
            </li>
            <li class="nav-item">
                <a class="nav-link" href="#">Link</a>
            </li>
            <li class="nav-item">
                <a class="nav-link" href="#">Link</a>
            </li>
        </ul>
    </div>
</nav>
```

NAVBAR-PROPERLY-RENDERED
![navbar-properly rendered](https://user-images.githubusercontent.com/634561/43286679-cec4be56-90f0-11e8-9f3f-eec2389c869b.png)

NAVBAR-BLAZOR-RENDERED
![navbar-blazor rendered](https://user-images.githubusercontent.com/634561/43286678-ceb8ebbc-90f0-11e8-9ad0-457005cc81b0.png)

MarkStega:
The lack of response to the hamburger menu is due to the lack of the bootstrap required js code.
In index.html I replaced
`link href="css/bootstrap/bootstrap.min.css" rel="stylesheet"`
with
`link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css"
script src="https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"
script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.0/umd/popper.min.js"
script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.1.0/js/bootstrap.min.js"`

There are less than & greater than symbols starting each line, I haven't figured out how to show html source.

[...]

Issue #1198, opened 2018-07-26
Issue on htmlblock render


uazo:
Consider this (please note: all in 1 line without space)
```
<NavLink class="nav-link" href="fetchdata"><span class="oi oi-list-rich" aria-hidden="true"></span> Fetch data</NavLink></li>
```

extension will compile with

```
builder.OpenComponent<Microsoft.AspNetCore.Blazor.Routing.NavLink>(48);
builder.AddAttribute(49, "class", "nav-link");
builder.AddAttribute(50, "href", "fetchdata");
builder.AddMarkupContent(51, "<span class=\"oi oi-list-rich\" aria-hidden=\"true\"></span>");
builder.AddAttribute(52, "ChildContent", (Microsoft.AspNetCore.Blazor.RenderFragment)((builder2) => {
    builder2.AddContent(53, " Fetch data");
}
));
builder.CloseComponent();
```

that's incorrect because add an attribute after a content.

SteveSandersonMS:
Fixed with 8c452ff71ad1c59af0115747794a313d91ace789

We'll apply a more long-term fix when we reintroduce the markup optimization later.

rynowak:
I've merged the PR #1209 that addresses the root cause of this

Issue #1193, opened 2018-07-26
0.5.0 is decoding HTML entities


guardrex:
I have angle bracket HTML entities in a Blazor client-side app page ...

```html
&lt;h1 style="font-style:@@_headingFontStyle"&gt;@@_headingText&lt;/h1&gt;
```

... it would render that content out with the encoded angle brackets. The browser would render that as text ...

```text
<h1 style="font-style:@_headingFontStyle">@_headingText</h1>
```

I just did an 0.4.0 -> 0.5.0 update. It's apparently decoding those `&lt;` and `&gt;` into actual angle brackets. The browser is now rendering that content as HTML.

<img width="481" alt="capture" src="https://user-images.githubusercontent.com/1622880/43234650-e0ead786-9042-11e8-8271-2860cabc0ae9.PNG">

conficient:
Confirmed, just found the same thing when I upgraded my project BlazorChatSample to 0.5.0.

Source: https://github.com/conficient/BlazorChatSample/blob/master/BlazorChatSample.Client/Pages/Index.cshtml this has `<code>invokeAsync&lt;object&gt;</code> to invoke..` in the HTML but on the browser is rendered as
```
<code>invokeAsync<object> to invoke .... <p></p>
```
See https://blazorchatsample.azurewebsites.net/ for example

guardrex:
```
&#x0003C;h1 style="font-style:@@_headingFontStyle"&#x0003E;@@_headingText&#x0003C;/h1&#x0003E;
```
```
&#60;h1 style="font-style:@@_headingFontStyle"&#62;@@_headingText&#60;/h1&#62;
```
... lead to the same result.

@danroth27 @SteveSandersonMS I can update the "Build your first" experience sample app (it doesn't use HTML entities), but I can't easily update the doc's "Common" sample app to 0.5.0 until this is resolved or I go with images as a workaround. The content explains what the sample code is doing, and I don't want to lose that presentation. If this will take a while to fix, then what I'll probably do is screenshot the 0.4.0 app's output and use images in the sample. Since that's a time-consuming workaround, I'll wait until I hear back from you on this before I proceed.

SteveSandersonMS:
Thanks for the report! And sorry for the inconvenience.

The earliest we could feasibly fix this would be in a 0.5.1 update possibly next week but more likely the week after.

guardrex:
[...]

PR #1207, opened 2018-07-27
Temporarily disable HtmlBlockPass

SteveSandersonMS:
We've received a number of bug reports about 0.5.0 having regressions for markup that people are commonly using (#1204, #1203, #1198, #1193). It looks like each of these relate to the new optimization in 0.5.0 that represents static HTML as markup frames.

It's not yet totally decided, but one possible way forwards for us would be:

* Via this PR, temporarily disable the compiler part of markup blocks and release as 0.5.1 tomorrow
   * This wouldn't stop people from using `MarkupString` (etc) as the runtime bits would be left enabled. I don't think any of the bugs above relate to the runtime bits anyway.
   * People just trying to render HTML as normal like before would no longer be affected.
* We also collect precise repro cases for each of the above issues
* We fix any `HtmlBlockPass` issues or any runtime issues and re-enable it in 0.6.0 (or 0.5.2 if we decide to do a further patch).

We'll make a final call on this tomorrow morning. cc @rynowak @danroth27

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorExtensionInitializer.cs
@@ -79,7 +79,9 @@ namespace Microsoft.AspNetCore.Blazor.Razor
             builder.Features.Add(new EventHandlerLoweringPass());
             builder.Features.Add(new RefLoweringPass());
             builder.Features.Add(new BindLoweringPass());
-            builder.Features.Add(new HtmlBlockPass());
+
+            // Temporarily disabled for 0.5.1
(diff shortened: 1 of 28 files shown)


rynowak:
I think the best thing to do is to revert the HtmlBlockPass and release a patch. We should collect the bugs and reintroduce it in master once we have fixes.

SteveSandersonMS:
@rynowak Not certain whether you mean reverting the entire commit that introduced the feature (3c4f7b5). I'd rather not revert that entire commit, because that would also remove the `MarkupString` feature, which is not affected by the reported issues and is a feature we've already advertised as available in docs and the blog post.

The changes in this PR are a more targeted removal of just the compiler part, so people can still use `MarkupString` if they want.

Let me know if you have any concerns with this, but I'll merge it shortly so we can proceed with the release.

Issue #821, opened 2018-05-11
Add onkeypress with throttling behaviour to binding


Andrzej-W:
Currently when we use `<input bind="@myVariable" />` variable is updated when input field looses focus (onchange event is called). Popular requirement in many applications is to implement incremental search. To do this we need:

- binding which updates variable after every input field change
- throttling - this is necessary if we want to do incremental search on the server or we have a big collection to filter; we should be able to specify a delay in milliseconds and our variable should be updated only after user stops typing for the specified period of time

I believe that throttling can be useful for other events also, for example mouse move, mouse wheel scrolling, etc.

SteveSandersonMS:
We already support binding to the `input` event (e.g., `bind-value-oninput=@myField`).

As for the throttling, that's an interesting question. For the short term at least I'd not want to complicate the event system with baking that in, at least not until there are more requests for it and we see the scenarios more broadly.

You can implement throttling yourself either by literally writing the code that throttles inside your event handler, or by using a library like Rx that has a built-in implementation. I know it's tricky to prevent redundant renders with this approach so we can keep considering it in the future if more people are requesting it often.

Andrzej-W:
@SteveSandersonMS unfortunately your solution doesn't work. This code
```
<input bind-value-oninput="@FieldValue" />
<p>@FieldValue</p>
@functions{
    string FieldValue { get; set; }
}
```
throws exception visible in console:
```
 SCRIPT5022: System.InvalidCastException: Specified cast is not valid.
  at Microsoft.AspNetCore.Blazor.Components.BindMethods+<>c__DisplayClass7_0.<SetValueHandler>b__0 (Microsoft.AspNetCore.Blazor.UIEventArgs _) <0x1c0c618 + 0x00014> in <0d9c96ffc9704aa9b088cbbe138126c6>:0 
  at Microsoft.AspNetCore.Blazor.Components.EventHandlerInvoker.Invoke (Microsoft.AspNetCore.Blazor.UIEventArgs e) <0x1c0c288 + 0x0008c> in <0d9c96ffc9704aa9b088cbbe138126c6>:0 
[...]

PR #1234, opened 2018-07-31
Marshal oninput events as UIChangeEventArgs

sebbert:
Blazor does handle the oninput event, but it is marshalled as a regular UIEventArgs. This means that we cannot access the new value of the input element from inside our oninput handler. This changes the type of `oninput` events to `UIChangeEventArgs`, so that we can access the new value from our event handler.

Addresses #821


SteveSandersonMS:
Thanks for this, @sebbert!

I also want to add an E2E test so moving this over to #1475.

Issue #1179, opened 2018-07-24
Incorrect OutputPath when using $(MSBuildThisFileDirectory)


SlowLogicBoy:
I have `Directory.Build.props` file:
```xml
<Project>
    <PropertyGroup>
        <OutputPath>$(MSBuildThisFileDirectory)out\bin</OutputPath>
        <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
    </PropertyGroup>
</Project>
```
When I build Blazor app I get:
```
C:\Users\User\.nuget\packages\microsoft.aspnetcore.blazor.build\0.4.0\targets\Blazor.MonoRuntime.targets(11,5): 
    warning MSB3026: 
        Could not copy 
"C:\Users\User\.nuget\packages\microsoft.aspnetcore.blazor.build\0.4.0\targets\..\tools\mono\asmjs\mono.asm.js" 
        to 
"d:\Code\Git\MyApp\src\Client\WebApp\d:\Code\Git\MyApp\out\bin\netstandard2.0\dist/_framework/asmjs/mono.asm.js". 
    Beginning retry 1 in 1000ms. The filename, directory name, or volume label syntax is incorrect : 
'd:\Code\Git\MyApp\src\Client\WebApp\d:\Code\Git\MyApp\out\bin\netstandard2.0\dist\_framework\asmjs'  
[d:\Code\Git\MyApp\src\Client\WebApp\WebApp.csproj]
```
Output Path seems to be generated incorrectly:
`d:\Code\Git\MyApp\src\Client\WebApp\d:\Code\Git\MyApp\out\bin\netstandard2.0\dist\_framework\asmjs`

**Workaround:**
I added additional `Directory.Build.props` file in Blazor project directory:
```xml
<Project>
    <!-- Reference to my main Directory.Build.props -->
    <Import Project="..\..\..\Directory.Build.props" />
    <PropertyGroup>
        <!-- Override OutputPath with relative path -->
        <OutputPath>..\..\..\out\bin</OutputPath>
    </PropertyGroup>
</Project>
```

SteveSandersonMS:
[As per docs](https://msdn.microsoft.com/en-us/library/bb629394.aspx), the `OutputPath` value must be relative to the project directory:

> OutputPath ... Specifies the path to the output directory, relative to the project directory, for example, "bin\Debug".

[...]

PR #1271, opened 2018-08-06
Use $(TargetDir) instead of $(ProjectDir)$(OutputPath)

natemcmaster:
Resolves https://github.com/aspnet/Blazor/issues/1179

Prevents errors when `$(OutputPath)` is an absolute path.


SteveSandersonMS:
This looks better - thanks @natemcmaster!

PR #1278, opened 2018-08-07
Automate authenticode code-signing using Roslyn sign tool

### LINKED ISSUE #1269: Enable code-signing for VSTS builds
Follow-up to #1251. Blocked on https://dotnet.visualstudio.com/internal/_workitems/edit/61

We should enable code-signing to run automatically on VSTS builds. One less thing to hand-crank.

natemcmaster:
This automates authenticode code-signing using the Roslyn's SignTool (which is currently being ported to Arcade) and MicroBuild (the *real* MicroBuild, not the aspnet namesake).

The biggest change here is that obj and bin folders are relocated. This is part of the upcoming convergence of KoreBuild and Arcade. It was easier to do this now to make code-signing work better with the SignTool.

Changes:
* Change project layout to match the expectations of Arcade and sign-tool
* Add signtool config file to configure OPC, NuGet, and Authenticode signing
* Fix a bug when BaseIntermediateOutputPath is set to an absolute path
* Split the build definition into ci-public and ci-official. ci-public does not code-sign.

~~TODO: flip from 'test' to 'real' signing once ticket 79507967 is approved.~~

Resolves #1269


.gitignore
```diff
@@ -1,9 +1,8 @@
 .vs/
-bin/
-obj/
```
rynowak: Is it really a good idea to remove these from gitignore? I get that we're not using these folders as part of the official build, but this just seems like it could cause annoyances.

.vsts/templates/blazor-build.yml
```diff
@@ -1,31 +1,49 @@
...
   - task: PublishBuildArtifacts@1
     displayName: Upload artifacts
     condition: and(succeeded(), eq(variables['system.pullrequest.isfork'], false))
     inputs:
-      pathtoPublish: artifacts/
+      pathtoPublish: artifacts/$(BuildConfiguration)/
       artifactName: artifacts-$(BuildConfiguration)
       artifactType: Container
+
+  - ${{ parameters.afterBuild }}
```
rynowak: FYI I'm totally #YOLOing on these yaml files. Not only do I have a meh reaction to yml, I don't know what kinds of things they do in the context of VSTS.

build/SignToolData.json
```diff
@@ -0,0 +1,246 @@
...
+    },
+    {
+      "certificate": "NuGet",
+      "strongName": null,
+      "values": [
+        "packages/*.nupkg"
+      ]
+    }
+  ],
+  "exclude": [
```
rynowak: This just seems super fraught with peril. Do we have to manually update this file when we add an assembly or add a dependeny? This sounds like a nightmare.

natemcmaster: Yes and yes. The Arcade team is working on a way to automatically generate these, but that feature isn't ready yet. So for now, you will need to manually specify how to handle new dependencies.

.vsts/templates/blazor-build.yml
natemcmaster: YAML docs: https://github.com/Microsoft/vsts-agent/blob/master/docs/preview/yamlgettingstarted.md

This `${{ }}` syntax is explained here: https://github.com/Microsoft/vsts-agent/blob/master/docs/preview/yamlgettingstarted-templateexpressions.md

.gitignore
natemcmaster: I removed this intentionally because I was hoping it would elevate the visibility of problems with misconfigured obj and bin paths.

natemcmaster: Also, if it wasn't clear from the PR description, the obj and bin folders share a common root under `artifacts/`

SteveSandersonMS:
Thanks @natemcmaster for proceeding with this.

From a quick scan, there's a huge amount going on that I don't immediately understand, a few bits of which I'm suspicious about. I'll need to go through it carefully. Could you please hold off on merging until I explicitly mark this as approved?

src/Microsoft.AspNetCore.Blazor.Cli/Server/Startup.cs
```diff
@@ -69,7 +69,7 @@ private static string FindClientAssemblyPath(IApplicationBuilder app)
         {
             var env = app.ApplicationServices.GetRequiredService<IHostingEnvironment>();
             var contentRoot = env.ContentRootPath;
-            var binDir = FindClientBinDir(contentRoot);
+            var binDir = AppContext.BaseDirectory;
```
SteveSandersonMS: I don't think this is right. The `BaseDirectory` would refer to the CLI tool, not the application you're launching. Or if you've verified this correctly picks out the correct `binDir` for the app being launched (in both debug and release modes) please let me know!

natemcmaster:
Sure, no problem

SteveSandersonMS:
As a more general inquiry, is the expectation that all `aspnet` repos would get similar changes to this, or do the others get it from "Universe"? Or is there any possibility that some of the nontrivial stuff from here (like `sign.proj`) will be able to go into the SDK rather than being put in each repo?

src/Microsoft.AspNetCore.Blazor.Cli/Server/Startup.cs
natemcmaster: Good point. This happens to work in our tests because we're using Microsoft.AspNetCore.Blazor.Cli as a ProjectReference. For those using this tool as  DotNetCliToolReference, BaseDirectory points to the wrong place.

To fix this properly, you'll need to [implement a technique](https://natemcmaster.com/blog/2016/12/26/project-evalutation-cli-tool/) we use in dotnet-watch and others to scrape project information.

I'll revert this and apply a different workaround to get tests passing.

natemcmaster:
The rest of the aspnet repos will get some changes w.r.t. project layout, but they won't have sign.proj. Sign.proj is temporary while we wait for dotnet/arcade#226

.gitignore
rynowak: It does seem clear, I just expect that we're going to find a case in the future where this happens and is-annoying.

.vsts/templates/blazor-build.yml
rynowak: I don't want to know.. that would cause me to forget something else 😆

build/SignToolData.json
rynowak: what's the ETA on this being reasonable?

samples/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Workaround for #261 which requires the bin and obj directory be in the project directory
+bin/
+obj/
```
rynowak: Can we just leave this alone in the .gitignore instead? I don't see how this helps.

natemcmaster: sure

build/SignToolData.json
natemcmaster: https://github.com/dotnet/arcade/issues/226 and https://github.com/dotnet/arcade/issues/396 are tracking work to manage this data better. These are currently in the August milestone which wraps up 8/23.

natemcmaster:
🆙 :date: i think I've addressed your feedback.

build/SignToolData.json
SteveSandersonMS: Out of interest, how did you obtain the list of assemblies here?

build/arcade.props
```diff
@@ -0,0 +1,37 @@
...
+<Project>
+
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+
+    <!--
+        Project output layout.
+
+        This customizes the locations to obj and bin directories, according
```
SteveSandersonMS: Not a big deal, but seems like this comment got cut short.

src/Microsoft.AspNetCore.Blazor.Templates/.gitignore
```diff
@@ -1,3 +1,5 @@
 # We only track the .template.config.src items in source control
 # The .template.config files are generated on build
 content/**/.template.config/
+bin/
+obj/
```
SteveSandersonMS: Looks like the two changes here aren't needed any more.

src/Microsoft.AspNetCore.Blazor.Templates/content/Directory.Build.props
```diff
@@ -0,0 +1,4 @@
+<!-- Intentional does not import parent Directory.Build.props file to isolate templates projects from the rest of the repo -->
```
SteveSandersonMS: nitpick: Intentional -> Intentionally

Here and in src/Microsoft.AspNetCore.Blazor.Templates/content/Directory.Build.targets

build/SignToolData.json
natemcmaster: SignTool produced an error about unrecognized binaries in packages. On local builds, signtool runs a test and skips the signing part, so if new packages appear or new binaries in packages appear as you make changes, local runs of `build.cmd` will fail.

src/Microsoft.AspNetCore.Blazor.Templates/content/Directory.Build.props
natemcmaster: Good catch. I fixed the grammar errors and incomplete code comments.

PR #1280, opened 2018-08-07
Fix for builds when the app assembly has spaces on it's name.

### LINKED ISSUE #1279: Project Build fails when startup assembly has spaces.
## Bug Description
It seems that the 0.5.1 templates fails to build when the solution name has spaces. The build task only gets a portion of the solution folder name as it's arguments for the CLI. The specific error is the ```_GenerateBlazorBootJson``` task on the ```write-boot-json``` part.
## Repro steps
 - Create a project with a solution name that has spaces e.g. ```Some Spaced name```
 - Build the solution

The fix is to add quotes on the ```@(IntermediateAssembly)``` for the Exe task.

RyoukoKonpaku:
Fixes #1279


SteveSandersonMS:
Thanks @RyoukoKonpaku!

PR #1286, opened 2018-08-09
Reenable markup blocks

rynowak:
There were two problems here:
- We weren't incrementing child count inside child content - fixed in #1290
- We weren't dealing with HTML entities correctly (fixed in this PR)

### DIFF
--- a/test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/ComponentRenderingTest.cs
+++ b/test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/ComponentRenderingTest.cs
@@ -173,9 +173,9 @@ namespace Microsoft.AspNetCore.Blazor.E2ETest.Tests
             Assert.Equal(
 @"
 <!--!-->
-<div><p>Some-Static-Text</p></div>
-<div id=""bar""><span>More-Static-Text</span></div>
-<div id=""baz""><span>Some-Dynamic-Text</span><span>But this is static</span></div>
+<!--!--><div><p>Some-Static-Text</p></div>
+<!--!--><div id=""bar""><span>More-Static-Text</span></div>
+<div id=""baz""><span>Some-Dynamic-Text</span><!--!--><span>But this is static</span></div>
 
 ",
                 appElement.FindElement(By.CssSelector("#foo")).GetAttribute("innerHTML"));
@@ -189,9 +189,9 @@ namespace Microsoft.AspNetCore.Blazor.E2ETest.Tests
             Assert.Equal(
 @"
 <!--!-->
-<div><p>Some-Static-Text</p></div>
-<div id=""bar"">&lt;span&gt;More-Static-Text&lt;/span&gt;</div>
-<div id=""baz""><span>Some-Dynamic-Text</span><span>But this is static</span></div>
+<!--!--><div><p>Some-Static-Text</p></div>
+<!--!--><div id=""bar"">&lt;span&gt;More-Static-Text&lt;/span&gt;</div>
+<div id=""baz""><span>Some-Dynamic-Text</span><!--!--><span>But this is static</span></div>
 
 ",
                 appElement.FindElement(By.CssSelector("#foo")).GetAttribute("innerHTML"));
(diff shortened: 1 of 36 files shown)


test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/ComponentRenderingTest.cs
```diff
@@ -156,6 +156,47 @@ public void CanRenderChildComponents()
             Assert.Equal("somevalue", styledElement.GetAttribute("customattribute"));
         }
 
+        // Verifies we can render HTML content as a single block
+        [Fact]
+        public void CanRenderChildContent_StaticHtmlBlock()
+        {
+            var appElement = MountTestComponent<HtmlBlockChildContent>();
+            Assert.Equal("<p>Some-Static-Text</p>",
+                appElement.FindElement(By.CssSelector("#foo")).GetAttribute("innerHTML"));
```
SteveSandersonMS: Nitpick: `By.CssSelector("#foo")` could be `By.Id("foo")`

I don't honestly mind whether you change this

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/ComponentRenderingTest.cs
```diff
@@ -156,6 +156,47 @@ public void CanRenderChildComponents()
...
+            var appElement = MountTestComponent<HtmlMixedChildContent>();
+            Assert.Equal(
+@"
+<!--!-->
+<!--!--><div><p>Some-Static-Text</p></div>
+<!--!--><div id=""bar""><span>More-Static-Text</span></div>
+<div id=""baz""><span>Some-Dynamic-Text</span><!--!--><span>But this is static</span></div>
+
+",
+                appElement.FindElement(By.CssSelector("#foo")).GetAttribute("innerHTML"));
```
SteveSandersonMS: It's great to have this extra E2E coverage.

It's a bit odd to see the `<!--!-->` details show up in the test though (and below). Elsewhere we've made a series of assertions about the DOM contents rather than snapshotting its HTML representation. It's not like it's a major issue, but the `<!--!-->` markers are a separate implementation detail and could change independently of the child content feature.

SteveSandersonMS: A further rationale for preferring not to rely on the per-character HTML representation is that if we later want to enable cross-browser tests, we'll find that different browsers pick different (but also valid) HTML representations. For example, having different whitespace around linebreaks, different attribute ordering, or differing rules about when to entity-encode a given character.

rynowak:
As I understand it, libraries are supposed to reference 2.1.0 not one of the specific patches. This is because of some trickery they did in the SDK that I don't understand @natemcmaster - what's the right thing to do?

natemcmaster:
> libraries are supposed to reference 2.1.0 not one of the specific patches.

In general, yes, I would recommend this to make your library compatible with as many 2.1 projects as possible. This goes hand in hand with the second recommendation -- always install the latest shared frameworks. You can get early access builds of 2.1.3 here: https://github.com/dotnet/versions/tree/master/build-info/dotnet/product/cli/release/2.1.3

rynowak:
What does that mean for builds though - we need Korebuild to pick up 2.1.3 and use it to run stuff.

natemcmaster:
The version of the shared runtime which KoreBuild installs is controlled by the code in this repo. It looks like this repo is configured to install just Microsoft.NETCore.App.

https://github.com/aspnet/Blazor/blob/a0c0978f3cc8db8596c8c20dae6198783cd075db/build/repo.props#L32-L35

If you want newer versions of the ASP.NET Core runtime, add or change repo.props to include 
```xml
 <ItemGroup> 
   <AspNetCoreRuntime Include="$(AspNetCoreRuntimeVersion)" /> 
 </ItemGroup> 
```

rynowak:
I decided to take your advice on the tests, this will be cleaner than more maintainable. Thanks!

rynowak:
Will send a follow up to address the the fix for the Kestrel version

PR #1291, opened 2018-08-10
Install ASP.NET Core runtime

rynowak:
This will make sure the build script includes the ASP.NET Core runtime
when running on local .NET. The effect of this is that our test projects
and apps will 'roll forward' unto the newest runtime without us
hardcoding it.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -4,9 +4,12 @@
   </PropertyGroup>
   <PropertyGroup Label="Package Versions"><!-- Used only in development when running the template contents directly from source -->
     <AspNetCorePackageVersion>2.1.0</AspNetCorePackageVersion>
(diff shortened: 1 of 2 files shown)


natemcmaster:
Where is AspNetCoreRuntimeVersion defined? I don't see it in the code.

rynowak:
OK, I think I missed something then, am I supposed to define it and set it to the version I want? I assumed this was defined somewhere magic like https://github.com/aspnet/Blazor/blob/master/build/repo.props#L33

natemcmaster:
No magic. In fact, it looks like MicrosoftNETCoreApp20PackageVersion and MicrosoftNETCoreApp21PackageVersion are undefined too. Because MSBuild is awesome, I think it's just not adding a new item because those variables are empty

rynowak:
mmmmk. So those probably belong in dependencies. props then

PR #1296, opened 2018-08-10
Merge sibling nodes during markup block rewrite

rynowak:
This change adds the ability to merge sibling nodes when possible during
markup block rewriting. We retain that invariant that each markup block
is a valid chunk of markup containing properly nested tags.

We still haven't done any work to remove whitespace yet, so most of the
cases where this comes into play right now will merge an element with
its surrounding whitespace.

### DIFF
--- a/test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/HtmlBlockPassTest.cs
+++ b/test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/HtmlBlockPassTest.cs
@@ -65,6 +65,76 @@ namespace Microsoft.AspNetCore.Blazor.Razor
         }
 
         [Fact]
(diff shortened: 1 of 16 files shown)


test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/HtmlBlockPassTest.cs
```diff
@@ -75,7 +145,7 @@ public void Execute_RewritesHtml_CSharpInAttributes()
   </head>
 </html>");
 
-            var expected = NormalizeContent(@"<div>foo</div>");
+            var expected = NormalizeContent("<div>foo</div>\n  ");
```
rynowak: These tests changed because the newline that follows the element has been merged in.

PR #1307, opened 2018-08-13
E2E benchmarks

SteveSandersonMS:
Adds a new project, `Microsoft.AspNetCore.Blazor.E2EPerformance`. When run it displays a benchmark runner that works by launching a real Blazor app in an iframe and scripting various perf test cases.

The intent is that:

(1) We can publish this app to pure static files, and hand a copy to the Mono WebAssembly team who can then try running it against newer builds of Mono WebAssembly (by replacing mono.wasm/mono.js), so as to quanity the effects of optimisations
(2) We can track the figures for different versions of Blazor (which includes the effects of moving to newer versions of Mono). We intend to record "official" timings using the ASP.NET perf test rig so as to be comparing with consistent hardware over time.

We can add further benchmark cases over time if we want, but the simple set in `icrosoft.AspNetCore.Blazor.E2EPerformance/wwwroot/benchmarks/*.js` will suffice for now.


SteveSandersonMS:
Note: I haven't yet figured out if/how this should be involved in CI runs.

rynowak:
It would be good if we could use it to get a pass/fail in our CI runs. Like our other benchmark tests have a special 'verification' mode that just runs them once to verify they aren't broken.

benchmarks/Microsoft.AspNetCore.Blazor.E2EPerformance/Pages/RenderList.cshtml
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    protected override void OnAfterRender()
+    {
+        BenchmarkEvent.Send("Finished rendering list");
+    }
+
+    static IEnumerable<WeatherForecast> GenerateForecasts(int count)
+    {
+        var rng = new Random();
```
rynowak: I'm not sure if its concerning that we're using the RNG as part of this benchmark. I don't think Random generally contributes too much perf overhead.

SteveSandersonMS: Just to be sure, I've now removed the RNG from here.

Andrzej-W:
Just for reference:
```c#
Random rnd = new Random();
            long sum = 0;
            int n = 100000;
            for (int ctr = 1; ctr <= n; ctr++)
            {
                int number = rnd.Next(0, 101);
                sum += number;
            }
```
On my computer (Intel Core i7 6700K, 4.2GHz turbo) this takes:
70ms - Firefox
85ms - Edge
215ms - Chrome
My phone Huawei Mate 9 Pro:
525ms - Chrome

Issue #1132, opened 2018-07-13
Add object pool for ServerSide RenderFrame serialization


rynowak:
Follow up item from #1116

We should consider pooling the byte arrays/streams from the render frame writer.

SteveSandersonMS:
Reopening because we had to bring back the temporary buffer in #1666.

PR #1321, opened 2018-08-16
Server rendering: shrink traffic

### LINKED ISSUE #1219: Experiment: Blazor Client vs Server performance
Default Blazor application is relatively simple, but for those interested I have just published two versions:

http://blazorserwer.azurewebsites.net/ (blazorserver name is already taken)
http://blazorclient.azurewebsites.net/

Both applications are in the same S1 plan in Western Europe data centre. Blazor version 0.5.1.

What is worth testing?
1. Click on `Home` and `Counter` menu. In Blazor client version there is no internet transmission at all. In server version `click` is transmitted to the server and server responds with updated page. About 1.5 - 2 KB.
2. You can also click on `Click me` button in `Counter` page. Yet again no transmission in client version and about 600 bytes in server version on every click.
3. Select `Fetch data`. In client version data is transmitted from the server in JSON format - about 5 KB. In server version we receive 26 KB. I have increased number of items to 50.

In my opinion server version looks responsive. It is normal that some transmission is required in experiment number 1 and 2. **But I think that in experiment number 3 server version transmits to much data.** If I understand correctly server has to send whole table in HTML format. This table is very simple:
```html
    <table class="table">
        <thead>
            <tr>
                <th>Date</th>
                <th>Temp. (C)</th>
                <th>Temp. (F)</th>
                <th>Summary</th>
            </tr>
        </thead>
        <tbody>
                <tr>
                    <td>7/28/2018</td>
                    <td>28</td>
                    <td>0</td>
                    <td>Freezing</td>
                </tr>
                ~~~~
```
Single table row has about 65-70 characters. Multiply this by 50, add ~200 bytes for page header and first paragraph, and we are well below 5 KB. Why do you need 26 KB to send it.

SteveSandersonMS:
Fixes #1219, and if we include the RecyclableMemoryStreamManager part, also #1132

This PR is a response to #1219. The goal is to make server-rendered Blazor send less data over the wire. In #1219 we found that changing the stock FetchData page to return 500 forecasts caused it to send about 240KB over the websocket connection on its initial render, which is far more than the equivalent HTML.

This PR implements two optimisations:

1. When `RenderBatchWriter` builds its string table, it now de-duplicates the strings. This reduces the 500-forecasts size from 243504 bytes to 137569 bytes (56% of original).
2. The serialized render batch data is now gzip-compressed. This further reduces the 500-forecasts size from 137569 bytes to 7705 bytes (3% of original).

So in summary, **this PR reduces the bandwidth requirement in the large-table scenario by 97%**.

For smaller payloads the percentage reduction would be less but really it's the bigger payloads that need optimisation most anyway.

**CR notes:** I'm a bit on the fence about whether to include the `Microsoft.IO.RecyclableMemoryStream` part. [It claims to make a big difference in eliminating large-object-heap bottlenecks](http://www.philosophicalgeek.com/2015/02/06/announcing-microsoft-io-recycablememorystream/), but I haven't got any measurements of that in our case. There's probably nothing harmful about using it, since it's been around for a while and pretty well battle-tested in some critical scenarios such as Bing. I lean towards including this unless @rynowak has concerns.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
@@ -4,6 +4,7 @@
 using MessagePack;
 using MessagePack.Formatters;
 using Microsoft.AspNetCore.Blazor.Rendering;
+using Microsoft.IO;
 using System;
 using System.IO;
 using System.IO.Compression;
@@ -17,6 +18,9 @@ namespace Microsoft.AspNetCore.Blazor.Server.Circuits
     /// </summary>
     internal class RenderBatchFormatterResolver : IFormatterResolver
     {
+        private static RecyclableMemoryStreamManager recyclableMemoryStreamManager
+            = new RecyclableMemoryStreamManager();
+
         public IMessagePackFormatter<T> GetFormatter<T>()
             => typeof(T) == typeof(RenderBatch) ? (IMessagePackFormatter<T>)RenderBatchFormatter.Instance : null;
 
@@ -30,14 +34,10 @@ namespace Microsoft.AspNetCore.Blazor.Server.Circuits
 
             public int Serialize(ref byte[] bytes, int offset, RenderBatch value, IFormatterResolver formatterResolver)
             {
-                // TODO: Instead of directly using MemoryStream here and in CompressStream below,
-                // consider using https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream
-                // That automatically provides pooling for the underlying buffers and avoids using
-                // the large object heap where possible. The logic here is already written to
-                // avoid calling memoryStream.GetBuffer() or memoryStream.ToArray(), so it should
-                // be possible to use RecyclableMemoryStream as a drop-in replacement.
-
-                using (var memoryStream = new MemoryStream())
+                // WARNING: Be careful never to execute .GetBuffer() or .ToArray() on this memory
+                // stream we get from recyclableMemoryStreamManager. If you do that, you'll silently
+                // eliminate all the benefit of RecyclableMemoryStream.
+                using (var memoryStream = recyclableMemoryStreamManager.GetStream($"{nameof(RenderBatchFormatter)}.{nameof(Serialize)}"))
                 using (var renderBatchWriter = new RenderBatchWriter(memoryStream, leaveOpen: false))
                 {
                     renderBatchWriter.Write(value);
@@ -56,7 +56,11 @@ namespace Microsoft.AspNetCore.Blazor.Server.Circuits
 
             private static Stream CompressStream(Stream input)
             {
-                var output = new MemoryStream();
+                // WARNING: Be careful never to execute .GetBuffer() or .ToArray() on this memory
+                // stream we get from recyclableMemoryStreamManager. If you do that, you'll silently
+                // eliminate all the benefit of RecyclableMemoryStream.
+                var output = recyclableMemoryStreamManager.GetStream($"{nameof(RenderBatchFormatter)}.{nameof(CompressStream)}");
+
                 using (var gzipStream = new GZipStream(output, CompressionLevel.Optimal, leaveOpen: true))
                 {
                     input.Seek(0, SeekOrigin.Begin);
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
@@ -6,6 +6,7 @@ using MessagePack.Formatters;
 using Microsoft.AspNetCore.Blazor.Rendering;
 using System;
 using System.IO;
+using System.IO.Compression;
 
 namespace Microsoft.AspNetCore.Blazor.Server.Circuits
 {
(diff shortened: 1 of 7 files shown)


stefan-schweiger:
@SteveSandersonMS I'm not sure if it's the best idea to use gzip by default.

1. I can see performance issues if there is compression of dynamic content for every interaction
2. I don't know if this applies in this situation since I'm not entirely familiar with the network technology behind SSR, but usually you should avoid using GZIP when transferring data over encrypted connections because of [BREACH and CRIME](https://www.helpnetsecurity.com/2016/08/11/compression-oracle-attacks-https/)

SteveSandersonMS:
@stefan-schweiger Thanks for raising those points.

1. The tradeoff of CPU time versus bandwidth is something that application developers will need to choose for themselves. We can make this possible to turn on or off. I strongly suspect most will want it on, so that's likely to be the default.
2. I'll follow up with our security people to make sure we're taking the right kinds of mitigations, and that this is a reasonable default. It's not uncommon to do HTTP compression within HTTPS (e.g., this is what ASP.NET's `UseResponseCompression` enables).

Andrzej-W:
BREACH and CRIME are relatively old. Attacker has to observe traffic between client and server. This one is newer and can be done with JavaScript only (the attacker does not have to observe the network traffic)
https://tom.vg/papers/heist_blackhat2016.pdf
Interesting reading. It is really important to consult this problem with security people.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
```diff
@@ -4,8 +4,10 @@
 using MessagePack;
 using MessagePack.Formatters;
 using Microsoft.AspNetCore.Blazor.Rendering;
+using Microsoft.IO;
```
rynowak: Asking just in in case: this needs approval approval from @Eilon  - even if its OSS and from Microsoft.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
```diff
@@ -29,15 +34,54 @@ public RenderBatch Deserialize(byte[] bytes, int offset, IFormatterResolver form
 
             public int Serialize(ref byte[] bytes, int offset, RenderBatch value, IFormatterResolver formatterResolver)
             {
-                using (var memoryStream = new MemoryStream())
+                // WARNING: Be careful never to execute .GetBuffer() or .ToArray() on this memory
+                // stream we get from recyclableMemoryStreamManager. If you do that, you'll silently
+                // eliminate all the benefit of RecyclableMemoryStream.
+                using (var memoryStream = recyclableMemoryStreamManager.GetStream($"{nameof(RenderBatchFormatter)}.{nameof(Serialize)}"))
```
rynowak: what's the string for here? leak tracking?

Could we put this in a constant or something static to avoid allocating a million of these?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
```diff
@@ -29,15 +34,54 @@ public RenderBatch Deserialize(byte[] bytes, int offset, IFormatterResolver form
...
+                    input.CopyTo(gzipStream);
+                    gzipStream.Flush();
+                    return output;
                 }
-                    
+            }
+
+            private static int CopyStreamToMessagePack(Stream stream, ref byte[] bytes, int offset)
+            {
+                var buffer = new byte[32768];
```
rynowak: Could we use `ArrayPool<>` here? It's fairly easy to use

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
```diff
@@ -29,15 +34,54 @@ public RenderBatch Deserialize(byte[] bytes, int offset, IFormatterResolver form
...
+                int read;
+
+                stream.Seek(0, SeekOrigin.Begin);
+                while ((read = stream.Read(buffer, 0, buffer.Length)) > 0)
+                {
+                    offset += MessagePackBinary.WriteBytes(ref bytes, offset, buffer, 0, read);
```
rynowak: does this method take ever replace the reference to `bytes`? I'm not sure I understand why the library requires `ref` in this case.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
```diff
@@ -29,15 +34,54 @@ public RenderBatch Deserialize(byte[] bytes, int offset, IFormatterResolver form
...
+                // stream we get from recyclableMemoryStreamManager. If you do that, you'll silently
+                // eliminate all the benefit of RecyclableMemoryStream.
+                var output = recyclableMemoryStreamManager.GetStream($"{nameof(RenderBatchFormatter)}.{nameof(CompressStream)}");
+
+                using (var gzipStream = new GZipStream(output, CompressionLevel.Optimal, leaveOpen: true))
+                {
+                    input.Seek(0, SeekOrigin.Begin);
+                    input.CopyTo(gzipStream);
+                    gzipStream.Flush();
+                    return output;
```
rynowak: Given that some folks have concerns about GZip, is it worth making it optional?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
SteveSandersonMS: Sounds good - will look into it.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
SteveSandersonMS: It’s the weird but performant way MessagePack APIs work to minimise GC churn. 

Yes, the reference can get replaced by MessagePack internally (when we call their Write method) if it needs to expand the underlying buffer or swap it for a different one when it runs out of space.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
SteveSandersonMS: For the CPU/bandwidth I agree there are legitimate reasons to go either way, but I’d like to see clearer measurements to prove the CPU cost could ever be prohibitive. We can always add more options later. It’s hard to see in what scenarios people would realistically turn this off. 

As for security, I’d like to get a really clear conclusion from our discussions with security people. As it stands I expect we will be able to completely eliminate the entire class of compression oracle attacks, as we have so much more control than in the more general HTTP response case.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
SteveSandersonMS: Actually I do think there's an issue here, but it's with how I'm updating `offset`. Looking into it.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchFormatterResolver.cs
SteveSandersonMS: Updated logic will eliminate the need for this temporary buffer.

SteveSandersonMS:
OK, so after consultation with security specialists, we're going to go in a different direction with this and avoid using compression by default. We won't be able to achieve the 97% size reduction suggested here, but should be about to get an 85% (ish) reduction without needing compression or deduplication of potentially sensitive data.

Issue #1330, opened 2018-08-17
Make server-rendering WebSocket traffic more compact


SteveSandersonMS:
This follows on from #1321. After consultation with security specialists, we've decided not to rely on gzip by default. Instead we intend to shrink the traffic by using more compact but fixed representations so as to avoid any potential issues.

Strategies:

1. Represent numbers as LEB128 instead of 4-byte ints. Also remove padding from edit frames and render tree frames.
  * This reduces the 500-item weather forecast test case from ~246500 bytes to ~135000 bytes
2. Deduplicate *only* HTML element names, attribute names, and pure whitespace
  * This reduces it further to ~40200 bytes
3. Stop writing out lists of array indices. These simplified things for the deserializer, but we can just make the deserializer do more work.
  * This reduces it further to ~3400 bytes

So in total, without relying on compression, we can reduce the traffic to about 14% of the original.

The proposal is to do only point (2) in 0.6.0, because it's very cheap to implement. The others will go on the backlog.

Andrzej-W:
Maybe you should also consider class names in point 2. Typical page can have a lot of duplications and some class names are long. A few examples from Bootstrap, but it is true for any framework.
Typical button: `class="btn btn-primary"`
"Advanced" button: `class="btn btn-primary dropdown-toggle dropdown-toggle-split"`
Dropdown items:  `class="dropdown-item"`
Carousel items: `class="carousel-item"`
Forms: `class="form-group"` and `class="form-control"`

Lupusa87:
Sorry if I bring extra noise but I think we can do something like serialize data before sending with special logic and static Key/Value dictionary.

This dictionary will be shared between server and client.
Sender will serialize and client deserialize with this dictionary.

If we can determine most often used keywords then we can store them in dictionary and set shorter code char. for example:
button => a (frequency 10)
onclick => b (frequency 5)
class => c (frequency 5)
btn btn-primary  => d (frequency 5)
div =>e (frequency 2)
style => f (frequency 1)
end etc
after limit we can use combinations $ab, $ff and etc.

using this dictionary for example this snippet we can represent with compact way.
Original:
```
<div style="vertical-align: middle; margin: 5px">
        <button class="btn btn-primary" onclick="@(e => AddString("4"))">4</button>
        <button class="btn btn-primary" onclick="@(e => AddString("5"))">5</button>
        <button class="btn btn-primary" onclick="@(e => AddString("6"))">6</button>
[...]

PR #1340, opened 2018-08-21
In RenderBatchWriter, deduplicate strings only when safe to do so

SteveSandersonMS:
We allow deduplication of HTML element and attribute names, plus whitespace text nodes / attribute values.

This is a partial implementation of #1330. It's the only part we plan to include in 0.6.0; others will come later.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchWriter.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchWriter.cs
@@ -34,11 +34,13 @@ namespace Microsoft.AspNetCore.Blazor.Server.Circuits
     internal class RenderBatchWriter : IDisposable
     {
         private readonly List<string> _strings;
+        private readonly Dictionary<string, int> _deduplicatedStringIndices;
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
The AppVeyor failure is that it's failing to install Google Chrome. This has happened on AppVeyor before and continued until someone at AppVeyor (or maybe Chocolatey) updated something.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RenderBatchWriter.cs
```diff
@@ -168,23 +171,25 @@ void Write(in RenderTreeFrame frame)
...
                     break;
                 case RenderTreeFrameType.Region:
                     _binaryWriter.Write(frame.RegionSubtreeLength);
                     WritePadding(_binaryWriter, 8);
                     break;
                 case RenderTreeFrameType.Text:
-                    WriteString(frame.TextContent);
+                    WriteString(
+                        frame.TextContent,
+                        allowDeduplication: string.IsNullOrWhiteSpace(frame.TextContent));
```
rynowak: nice :)

PR #1341, opened 2018-08-21
Add Blazor.temporaryOnConnectionClose to fix #1339

### LINKED ISSUE #1339: Subscribe to the onclose event via JS
Hello, I want to subscribe to the onclose event via JS. 

**Here is an example of what I need**: 
https://youtu.be/11I4lK00KeI

How do I do this without changing "/_framework/blazor.server.js" ?

---
**Blazor Server side**: 0.5.1
/_framework/blazor.server.js

SteveSandersonMS:
This API will be removed once we implement proper connection management (reconnects, etc.)

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Boot.Server.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Boot.Server.ts
@@ -47,6 +47,10 @@ function boot() {
       );
     })
     .catch(unhandledError);
+
+  // Temporary undocumented API to help with https://github.com/aspnet/Blazor/issues/1339
+  // This will be replaced once we implement proper connection management (reconnects, etc.)
+  window['Blazor'].onServerConnectionClose = connection.onclose.bind(connection);
 }
 
 function unhandledError(err) {


SteveSandersonMS:
As before, the AppVeyor failure is bogus and can be ignored.

danroth27:
Do we really need "temporary" in the API name? Can't we track it's temporary nature with a GitHub issue? I get that we expect change here, but that's true of much of Blazor at this point.

cores-system:
One of the options, you could open access to SignalR.
This would help us manage the connection yourself before the API is released.

`window['BlazorHub'] = connection;`

SteveSandersonMS:
@danroth27 OK, fair point. I've removed the "temporary" from the name. I still expect this API will be removed later, but I agree this is in line with not making back-compat guarantees about anything else either.

@cores-system Where possible I'd prefer to limit the number of APIs we expose that won't be supported in the long run.

SteveSandersonMS:
CI failure is the Kestrel "cannot access a disposed object" bug we've seen before. Awaiting an ASP.NET Core update to fix that.

Issue #1216, opened 2018-07-27
Referencing Microsoft.AspNetCore.Blazor.Server disables server-side tag helpers


SteveSandersonMS:
Repro:

 * Create a Blazor hosted or server-side app
 * In the server project, make it render a Razor page on the server (add `AddMvc` if not already present, then put `Index.cshtml` into a `Pages` directory)
 * Add `@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers` if you don't have it in `_ViewImports.cshtml` already
 * Observe that if you have syntax like `<img src="~/blah.jpg" />`, it does *not* resolve the `~` - it literally emits the URL `~/blah.jpg`

I expect that Blazor's MSBuild targets are hijacking the server-side Razor compilation process, eliminating the normal server-side tag helpers, or something of that sort.

danroth27:
@Eilon

floreseken:
Aha.. I had to remove the `~/` to make mine work also.. but I hadn't linked it to Blazor. Now it makes sense.

rynowak:
There are two issues here:
1. Our project templates should have a `privateassets=all`
2. Our workaround for https://github.com/aspnet/Razor/issues/2482 is breaking tag helper discovery

We'll need to solve this by making sure that our `ITagHelper` workaround is referenced by client projects but not server projects.

PR #1343, opened 2018-08-21
Add privateassets=all to Blazor.Build usages

rynowak:
We recommend using this package with PrivateAssets=all everywhere to
make sure that the MSBuild files brought in by this package aren't
applied to transitively.

When that happens, the Blazor.Build MSBuild files will take over the
Razor functionality for other projects, which breaks MVC's view
compilation functionality.

This is part of a fix for #1216.


rynowak:
It already is the equivalent behavior which I think is why we struggled to find the issue. PackageReference is transitive, including a bunch of loose MSBuild files is not.

PR #1345, opened 2018-08-21
Move ITagHelper to another assembly

rynowak:


src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.nuspec
```diff
@@ -11,10 +11,10 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="_._" target="lib/netstandard1.0/" />
+    <file src="$publishdir$netcoreapp2.1\Microsoft.AspNetCore.Blazor.TagHelperWorkaround.dll" target="lib/netstandard1.0/" />
```
rynowak: This will be referenced as part of Blazor.Build. Since we do private assets there, it won't be transitive.

rynowak: When we no longer need this workaround we can just remove the assembly from here. Since there's no code a user can see in this assembly it won't be needed.

src/Microsoft.AspNetCore.Blazor.Build/ReferenceFromSource.props
```diff
@@ -29,6 +29,11 @@
     <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="$(RazorPackageVersion)"/>
   </ItemGroup>
 
+  <!-- This is used as a P2P when building the repo. Normal Blazor projects will get this as a reference through the Blazor.Build package -->
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\Microsoft.AspNetCore.Blazor.TagHelperWorkaround\Microsoft.AspNetCore.Blazor.TagHelperWorkaround.csproj" />
+  </ItemGroup>
```
SteveSandersonMS: I'm guessing then that `ReferenceFromSource.props` inherently behaves like "privateassets=all", otherwise this change would break the "HostedInAspNet" sample (because the server project references the client, which imports `ReferenceFromSource.props` and therefore obtains a reference to TagHelperWorkaround).

Please let me know if I'm misunderstanding this!

src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.nuspec
SteveSandersonMS: OK good

src/Microsoft.AspNetCore.Blazor.Build/ReferenceFromSource.props
rynowak: So it turns out that p2p references support privateassets as well, which is needed here.

PR #1349, opened 2018-08-22
fix #1200: include credentials when fetching blazor.boot.json to enable windows auth

### LINKED ISSUE #1200: Chrome: ASP.NET Core Hosted Blazor app does not start if Windows authentication is enabled (Blazor 0.5.0)
Scenario: I have a Blazor app created using the „ASP.NET Core hosted“ template, using IISExpress to run the app. On the server project, I have the „Enable Windows authentication“ checkbox ticked, ie. the corresponding part of lauchSettings.json file is
```code
"iisSettings": {
    "windowsAuthentication": true,
    "anonymousAuthentication": false,
    "iisExpress": {
      "applicationUrl": "http://localhost:8000/",
      "sslPort": 0
    }
  }
```

When I try to launch the application in Chrome, the Console shows an error and the application will not start:

```code
blazor.webassembly.js:1 GET http://localhost:8000/_framework/blazor.boot.json 401 (Unauthorized)
blazor.webassembly.js:1 Uncaught (in promise) SyntaxError: Unexpected token < in JSON at position 0
    at Object.<anonymous> (blazor.webassembly.js:1)
    at blazor.webassembly.js:1
    at Object.next (blazor.webassembly.js:1)
    at i (blazor.webassembly.js:1)
```

The “Network” tab indicates a problem with loading the blazor.boot.json file using the “fetch” method (screenshot attached).
chromefetcherror

 

The problem started to show after updating Blazor to 0.5.0. Before the update, all was working fine.
I’ve tried many things to troubleshoot the issue:
- in Firefox and in Edge, the same app works fine
- I have tried to set-up CORS middleware on the server-side, no luck (but then I am not sure if this issue is related to CORS at all)
- if I run the Fiddler debugging proxy sitting between Chrome and the IISExpress server, the issue miraculously disappears and the application launches fine even in Chrome (after disabling Fiddler, the issue re-appears)
- tried to empty Chrome cache, disabled all extensions in Chrome, disabled all unnecessary middlewares in the server app -> no luck

Do you have any suggestions what might be wrong? Thanks.

chris-ray:
Included the suggested change by @radek7210 in issue #1200.
 - Resolves problem in both Client-side and Server-side
 - I could only reproduce this issue locally on an AD domain-connected computer and on an Azure App Service behind Azure AD authentication

Addresses #1200

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/BootCommon.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/BootCommon.ts
@@ -1,7 +1,7 @@
 export async function fetchBootConfigAsync() {
   // Later we might make the location of this configurable (e.g., as an attribute on the <script>
   // element that's importing this file), but currently there isn't a use case for that.
-  const bootConfigResponse = await fetch('_framework/blazor.boot.json');
+  const bootConfigResponse = await fetch('_framework/blazor.boot.json', { method: 'Get', credentials: 'include' });
   return bootConfigResponse.json() as Promise<BootJsonData>;
 }
 


SteveSandersonMS:
Thanks @chris-ray - this looks good! Adding to 0.6.0 so we'll aim to merge it in time for that.

PR #1354, opened 2018-08-23
fix #1352

### LINKED ISSUE #1352: Router intercepts clicks on <a> with target='_blank'
Hello, blazor intercepts a click even if you specify **target="_ blank"** in the link.

**Test Host**:
http://domain.com/

**Examples**
```
<a href="/api/rewriteurl/test" target="_blank">not work</a>
<a href="https://domain.com/api/rewriteurl/test" target="_blank">WORK</a>
<a href="http://domain.com/api/rewriteurl/test" target="_blank">not work</a>
<a href="http://domain2.com/api/rewriteurl/test" target="_blank">WORK</a>
```

I think that it's not necessary to intercept links with **target="_ blank"**

**Blazor Server side**: 0.6.0-preview1-20180822.3
/_framework/blazor.server.js

cores-system:
fix https://github.com/aspnet/Blazor/issues/1352

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
@@ -27,7 +27,7 @@ function enableNavigationInterception(assemblyName: string, functionName: string
     if (anchorTarget && anchorTarget.hasAttribute(hrefAttributeName) && event.button === 0) {
       const href = anchorTarget.getAttribute(hrefAttributeName)!;
       const absoluteHref = toAbsoluteUri(href);
-      const blank = anchorTarget.getAttribute('target') == '_blank';
+      const blank = anchorTarget.getAttribute('target') === '_blank';
 
       // Don't stop ctrl/meta-click (etc) from opening links in new tabs/windows
       if (isWithinBaseUriSpace(absoluteHref) && !eventHasSpecialKey(event) && !blank) {
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
@@ -27,9 +27,10 @@ function enableNavigationInterception(assemblyName: string, functionName: string
     if (anchorTarget && anchorTarget.hasAttribute(hrefAttributeName) && event.button === 0) {
       const href = anchorTarget.getAttribute(hrefAttributeName)!;
       const absoluteHref = toAbsoluteUri(href);
+      const blank = anchorTarget.getAttribute('target') == '_blank';
 
       // Don't stop ctrl/meta-click (etc) from opening links in new tabs/windows
-      if (isWithinBaseUriSpace(absoluteHref) && !eventHasSpecialKey(event)) {
+      if (isWithinBaseUriSpace(absoluteHref) && !eventHasSpecialKey(event) && !blank) {
         event.preventDefault();
         performInternalNavigation(absoluteHref);
       }
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
@@ -26,6 +26,7 @@ function enableNavigationInterception(assemblyName: string, functionName: string
     const hrefAttributeName = 'href';
     if (anchorTarget && anchorTarget.hasAttribute(hrefAttributeName) && event.button === 0) {
       const href = anchorTarget.getAttribute(hrefAttributeName)!;
+      const absoluteHref = toAbsoluteUri(href);
       const targetAttributeValue = anchorTarget.getAttribute('target');
       const opensInSameFrame = !targetAttributeValue || targetAttributeValue === '_self';
 
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
@@ -26,13 +26,11 @@ function enableNavigationInterception(assemblyName: string, functionName: string
     const hrefAttributeName = 'href';
     if (anchorTarget && anchorTarget.hasAttribute(hrefAttributeName) && event.button === 0) {
       const href = anchorTarget.getAttribute(hrefAttributeName)!;
-      const absoluteHref = toAbsoluteUri(href);
-
-      // https://github.com/aspnet/Blazor/issues/1352#issuecomment-415444870
-      const isExternal = ['_blank', '_parent', '_self', '_top'].indexOf(anchorTarget.getAttribute('target') || '') > -1;
+      const targetAttributeValue = anchorTarget.getAttribute('target');
+      const opensInSameFrame = !targetAttributeValue || targetAttributeValue === '_self';
 
       // Don't stop ctrl/meta-click (etc) from opening links in new tabs/windows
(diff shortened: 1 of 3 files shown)


SteveSandersonMS:
Thanks for submitting this. We'd also need an end-to-end (E2E) test for the new behaviour, perhaps similar to this: https://github.com/aspnet/Blazor/blob/master/test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/RoutingTest.cs#L86

cores-system:
@SteveSandersonMS, I'll try to do E2E tomorrow.

cores-system:
Unfortunately I did not find the optimal way to check _self.

cores-system:
Any thoughts on how to do it better?

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
```diff
@@ -28,8 +28,11 @@ function enableNavigationInterception(assemblyName: string, functionName: string
       const href = anchorTarget.getAttribute(hrefAttributeName)!;
       const absoluteHref = toAbsoluteUri(href);
 
+      // https://github.com/aspnet/Blazor/issues/1352#issuecomment-415444870
+      const isExternal = ['_blank', '_parent', '_self', '_top'].indexOf(anchorTarget.getAttribute('target') || '') > -1;
+
       // Don't stop ctrl/meta-click (etc) from opening links in new tabs/windows
-      if (isWithinBaseUriSpace(absoluteHref) && !eventHasSpecialKey(event)) {
+      if (isWithinBaseUriSpace(absoluteHref) && !eventHasSpecialKey(event) && !isExternal) {
```
SteveSandersonMS: This logic doesn't deal with other external targets, e.g., the custom names of popup windows opened previously.

Could we simplify the rule to "skip interception if there's a `target` attribute with any value other than `_self`"? That covers all the cases AFAIK.

Could be:

```typescript
const targetAttributeValue = anchorTarget.getAttribute('target');
const opensInSameFrame = !targetAttributeValue || targetAttributeValue === '_self';

if (isWithinBaseUriSpace(absoluteHref) && !eventHasSpecialKey(event) && opensInSameFrame) {
   ...
}
```

SteveSandersonMS: Also we don't need the comment that links to the GitHub issue.

test/testapps/BasicTestApp/RouterTest/Links.cshtml
```diff
@@ -20,3 +20,6 @@
 <a id="anchor-with-no-href">
     Anchor tag with no href attribute
 </a>
+
+<a href="/" target="_blank">Target (_blank)</a>
+<a href="" target="_self">Target (_self)</a>
```
SteveSandersonMS: Why do these two `<a>` tags have different `href` values? What distinction is being tested?

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/RoutingTest.cs
```diff
@@ -117,6 +117,47 @@ public void CanFollowLinkToOtherPageWithCtrlClick()
...
+        public void CanFollowLinkToTargetSelfOrParentAndTopClick()
+        {
+            SetUrlViaPushState("/");
+
+            var app = MountTestComponent<TestRouter>();
+            var initialUrl = Browser.Url;
+
+            app.FindElement(By.LinkText("Target (_self)")).Click();
+
+            WaitAssert.Equal(true, () => Browser.WindowHandles.Count == 1 && initialUrl == Browser.Url);
```
SteveSandersonMS: This doesn't look right. Why are you testing that `initialUrl == Browser.Url`? That's what would already be the case before you clicked the link.

We can't use `WaitAssert` to check for things that are already the case before the action you're testing, because it will evaluate as successful instantly even if your action later proceeds to violate the condition.

Instead we have to use `WaitAssert` to check for something that changes due to the action being tested. For this test, could we have the `target="_self"` link point to some different page, and then have the test check we successfully navigated to it?

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
cores-system: In this case, I can not open the file in the current tab.
`<a href="/file.pdf" target="_self">Open file this tab</a>`

**Error**: https://github.com/aspnet/Blazor/issues/1367

test/testapps/BasicTestApp/RouterTest/Links.cshtml
cores-system: 1) href="/" opens the link "/" new tabs
2) href="" reloads the page current tab

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/RoutingTest.cs
cores-system: In this example, target="_ self" performs a page reload.
If there are options how to improve this, suggest, unfortunately there are no options how to test for  target="_ self"

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
cores-system: Given that any explicit **target="_*"** value should be ignored by the router, you can simplify the code to this state.

```
const targetAttributeValue = anchorTarget.getAttribute('target');

if (isWithinBaseUriSpace(absoluteHref) && !eventHasSpecialKey(event) && !targetAttributeValue ) {
   ...
}
```

SteveSandersonMS: That sounds like you're trying to use `target="_self"` as a hack to skip click interception. I don't think that's the right way to indicate that you don't want to intercept the click, because `target="_self"` doesn't have any special behavior like that according to HTML specs.

Rather, if there is to be a way of indicating that a given link should open in the same frame but not do click interception, that would be a separate feature.

test/testapps/BasicTestApp/RouterTest/Links.cshtml
SteveSandersonMS: Sorry if I was unclear. I'm asking *why* you're doing that. Why does the `"/"` go with `_blank` and the `""` go with `_self`?

test/Microsoft.AspNetCore.Blazor.E2ETest/Tests/RoutingTest.cs
SteveSandersonMS: As per comment above, I don't think `_self` should trigger a page reload. If there is to be a way of skipping interception for links that open in the same frame, that should be expressed in some other way rather than trying to give a new meaning to `_self`.

cores-system: Forced to agree with you, we can choose the optimal solution for UriHelper.ts and after that based on the new UriHelper.ts I will pass the E2E test for `_self`

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/UriHelper.ts
SteveSandersonMS: There's a good chance we'll end up with a feature similar to your proposed `data-route="skip"`, but could we leave that to a separate PR?

Generally we find it good to keep different features in different PRs. So ideally this PR will handle only the `target` attribute handling.

cores-system:
Updated the code for the solution of problem https://github.com/aspnet/Blazor/issues/1352
Issue https://github.com/aspnet/Blazor/issues/1367 можно отметить в Backlog

cores-system:
I think the error is not related to the last update

PR #1357, opened 2018-08-24
Run compliance checks on official builds

natemcmaster:
See https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=1974924&_a=summary&view=logs for example


natemcmaster:
cc @Eilon

Issue #404, opened 2018-03-27
Template based components


danroth27:
For example, you should be able to create a Grid or ListView component that binds to each item in a collection using a define template.

SteveSandersonMS:
Specifically, this means we need some way to signal that a given part of the child content within a component should get compiled as an `Action<RenderTreeBuilder, T>`, for some type T that we need to be able to resolve dynamically from usage at design time. Then our design-time codegen will be able to produce the correct intellisense context when writing Razor expressions inside that child element.

Example:

```
<MyListWithHeader RowItems=@People>
    <Title>Here I'm still in the outer context so can reference properties on the component itself such as @TitleText</Title>
    <RowTemplate> <!-- Somehow need intellisense to know that RowTemplate is meaningful here -->
        Here I'm in the context of an Action<RenderTreeBuilder, Person> because we supplied
        an IEnumerable<Person> as the value for RowItems, so I can write @item.FirstName
    </RowTemplate>
</MyListWithHeader>
```

I'm not sure how general we want to make this feature. Possibly less general than indicated in the sketch above. @rynowak would be the expert on what's reasonable to implement here. All the complexity is at design/compile time. It doesn't require any runtime support that I can think of.

abatista667:
Hi i was trying to accomplish this by my own, I was able to generate a custom grid passing an IEnumerable<TEntity> as a parameter for the component, but I was not able to update this grid when the DataSource parameter change.

`<table class="@ClassName">
  <tr>
    @foreach (var col in Columns)
    {
      <th>@col</th>
    }
  </tr>

  @foreach (System.Data.DataRow row in table.Rows)
  {
    <tr>
      @foreach (var col in Columns)
      {
        <td>@row[col]</td>
      }
      @if (Select != null)
      {
        <td><a @onclick(() => Select(row))>Seleccionar</a></td>
      }
    </tr>
  }
</table>
@Log
<a @onclick(() => UpdateState())>Actualizar</a>
@functions{
    public string ClassName { get; set; }
    public List<string> Columns { get; set; }
    public object DataSource { get; set; }
    private System.Data.DataTable table;


    private string Log;

    //call backs it seens I will have many of them

    public Action Change { get; set; }
    public Action<object> Select { get; set; }

    protected void UpdateState()
    {
      int count = ((IEnumerable<object>)DataSource).Count();
      Log = $"Table count rows:{table.Rows.Count} count object ${count}";
      StateHasChanged();
    }

    protected override void OnInit()
[...]

PR #1360, opened 2018-08-25
Add support for Razor templates

rynowak:
Adds support for Razor templates and `RenderFragment<T>`.

Razor templates are a little-known Razor feature that looks like:
```
@<tag>....<tag>
```

It's so little known that it's not even covered in our docs, but it's
been around for many many years. This features hasn't been implemented
until now for Blazor, and this feature brings it back as a build
building block for templated components (more to come).

In Blazor land a template like:
```
@{ RenderFragment<Person> template = @<div>@item.Name</div>; }
```

complies to code like:
```
RenderFragment<Person> template = (__builder, item) =>
{
    __builder.OpenElement(...);
    ...
    __builder.CloseElement(...);
}
```

Since the declaration always has a generic type parameter inside, it
needs to be in a context where the type is known.. ie: not with `var`.

See tests for ways to consume templates.

NOTE: There are the following caveats for templates
- Templates require a single root element.
- Templates don't work in the `@functions { }` block

These limitations are baked into the core of Razor and will take a while
for us to address (v3.0).


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -468,6 +499,27 @@ IReadOnlyList<IntermediateToken> GetCSharpTokens(ComponentAttributeExtensionNode
...
+            }
+
+            if (node == null)
+            {
+                throw new ArgumentNullException(nameof(node));
+            }
+
+            // Since the user doesn't write this name themselves, we have to use something hardcoded
+            // and predicatable.
+            const string VariableName = "item";
```
rynowak: Thoughts on this vs context?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
```diff
@@ -202,7 +203,7 @@ public override void WriteHtmlElement(CodeRenderingContext context, HtmlElementI
                 context.RenderNode(capture);
             }
 
-            _scopeStack.OpenScope(tagName: node.TagName, isComponent: false);
+            _scopeStack.OpenElementScope(node.TagName);
```
rynowak: Renamed these since there are three scopes now

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComplexAttributeContentPass.cs
```diff
@@ -73,17 +74,59 @@ private static bool TrySimplifyContent(IntermediateNode node)
...
+                            if (expression.Children.Count > 0 &&
+                                expression.Children[0] is IntermediateToken transitionToken &&
+                                transitionToken.Content == "@")
+                            {
+                                expression.Children.RemoveAt(0);
+                            }
+                        }
+
+                        // We were able to simplify it, all good.
+                        return false;
```
rynowak: This class is accumulating more artifacts due to Razor bugs 😞

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComplexAttributeContentPass.cs
```diff
@@ -73,17 +74,59 @@ private static bool TrySimplifyContent(IntermediateNode node)
...
+                        cSharpNode.Children.RemoveAt(2);
+                        cSharpNode.Children.RemoveAt(0);
+
+                        // We were able to simplify it, all good.
+                        return false;
+                    }
+                }
+
+                // There's yet another special case when you have an attribute with a template
+                // like <div attr="@(@<div>Hi</div>)"
```
rynowak: So the story with templates inside attributes. I thought it woulnd't work, but it does. Well mostly. Colorization doesn't work in VS. Completion of C# inside the attribute works.  Basically with a bunch of workarounds we can have templates work inside component attributes. 

This seems worthwhile enough to me to tolerate a bit of tooling pain for now. Thoughts?

src/Microsoft.AspNetCore.Blazor/RenderFragmentExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <see cref="RenderFragment"/> so that it can be used by the rendering system from Razor code.
+        /// </summary>
+        /// <typeparam name="T">The type of the value used by the <paramref name="fragment"/>.</typeparam>
+        /// <param name="fragment">A <see cref="RenderFragment{T}"/>, usually produced by a Razor template.</param>
+        /// <param name="value">The value of type <typeparamref name="T"/>.</param>
+        /// <returns>A <see cref="RenderFragment"/>.</returns>
+        public static RenderFragment Bind<T>(this RenderFragment<T> fragment, T value)
+        {
+            return (b) => fragment(b, value);
+        }
```
rynowak: @SteveSandersonMS  - interested in your suggestions here. Basically you have no way to consume a `RenderFragment<T>` from 'loose' Razor code. Example:

```
@{ RenderFragment<Person> p = @<div>...</div>; }
...
<div>
   @p     @* doesn't work the render tree builder doesn't understand RenderFragment<Person> *@
</div>
<div>
   @p(new Person(...))    @* doesn't work, wrong number of args *@
</div>
<div>
   @p(???, new Person(...))    @* doesn't work, I don't have a builder to pass in *@
</div>
```

Also, credit to @uazo - I was inspired by some of the things he tried.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
```diff
@@ -102,6 +102,26 @@ public void AddContent(int sequence, RenderFragment fragment)
...
+        {
+            if (fragment != null)
+            {
+                // We surround the fragment with a region delimiter to indicate that the
+                // sequence numbers inside the fragment are unrelated to the sequence numbers
+                // outside it. If we didn't do this, the diffing logic might produce inefficient
+                // diffs depending on how the sequence numbers compared.
+                OpenRegion(sequence);
+                fragment(this, value);
+                CloseRegion();
```
rynowak: @SteveSandersonMS - intended for use from a component written in C#. Useful?

test/Microsoft.AspNetCore.Blazor.Build.Test/ComponentRenderingRazorIntegrationTest.cs
```diff
@@ -554,5 +554,57 @@ public class MyComponent : BlazorComponent
...
+                frame => AssertFrame.Component(frame, "Test.Repeater", 4, 0),
+                frame => AssertFrame.Attribute(frame, "Count", typeof(int), 1),
+                frame => AssertFrame.Attribute(frame, "Value", typeof(string), 2),
+                frame => AssertFrame.Attribute(frame, "Template", typeof(RenderFragment<string>), 3),
+                frame => AssertFrame.Element(frame, "div", 2, 4),
+                frame => AssertFrame.Text(frame, "hello, world!", 5),
+                frame => AssertFrame.Element(frame, "div", 2, 4),
+                frame => AssertFrame.Text(frame, "hello, world!", 5),
+                frame => AssertFrame.Element(frame, "div", 2, 4),
+                frame => AssertFrame.Text(frame, "hello, world!", 5));
```
rynowak: @SteveSandersonMS - this doesn't seem right to me but I'm not sure what I would do differently. I expected to see these elements and text inside a frame for `<Repeater>`

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
```diff
@@ -598,5 +598,70 @@ public void SupportsTwoWayBindingForEnumValues()
...
+        };
+    }
+}");
+
+            // Act
+            var frames = GetRenderTree(component);
+
+            // Assert
+            //
+            // Review: @stevesa
```
rynowak: @SteveSandersonMS

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComplexAttributeContentPass.cs
```diff
@@ -73,17 +74,59 @@ private static bool TrySimplifyContent(IntermediateNode node)
...
                 // handling this case as a tactical matter since it's important for lambdas.
-                if (cSharpNode.Children.Count == 3 &&
-                    cSharpNode.Children[0] is IntermediateToken token0 &&
-                    cSharpNode.Children[2] is IntermediateToken token2 &&
-                    token0.Content == "(" &&
-                    token2.Content == ")")
                 {
-                    cSharpNode.Children.RemoveAt(2);
-                    cSharpNode.Children.RemoveAt(0);
+                    if (cSharpNode.Children.Count == 3 &&
```
rynowak: Reminder to myself to recompute the span of these nodes after editing them. I think some of the issues related to completion are due to these spans being wrong.

uazo:
just out of curiosity, but why have another syntax? why simple
```
<component>
   <templateprop>
       <div>test</div>
   </templateprop>
</component>
```
was not it enough?
and also, how does the new syntax with nested template work?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
RyoukoKonpaku: Hmm how about ```DataContext```? Not too sure but It would probably help out those coming from the XAML world since it uses a lot of Templates for user controls there.

RyoukoKonpaku:
@uazo I think it's to accommodate the existing functionality that Razor had (which is something I didn't know about as well which is a cool tidbit.). Though I do agree inlined templates would be better imo. I liked the property syntax you had last time ```Component.TemplateName``` but I guess this is a good starting point.

uazo:
perfect, let's have both of them, obviously with `Component.TemplateName`. that syntax has no problems with intellisense and colorization, and doesn't require a single root element

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
```diff
@@ -130,38 +131,63 @@ public override void VisitTagHelperHtmlAttribute(TagHelperHtmlAttributeIntermedi
                 // node to the rather simpler form that property nodes usually have.
                 for (var i = 0; i < attribute.Children.Count; i++)
                 {
+                    IntermediateNode child = null;
```
SteveSandersonMS: Would it be fair to say that the most complex aspect of the implementation is handling templates inside attributes? It looks like it takes up the bulk of the really complex cases you're having to handle here.

I'm wondering to what extent this scenario is going to come up in practice. Actually declaring any markup within an attribute looks pretty unfriendly. For example, given the choice between:

```
<MyFancyList GenericType="Customer" Data=customerRecords
   ItemTitle=@(@<div>@context.Name (@context.Email)</div>)>
   ItemBody=@(@<div>
       <ul>
       @foreach (var message in context.RecentMessages)
       {
           <li>@message</li>
       }
       </ul>
   </div>)>
</MyFancyList>
```

... and:

```
<MyFancyList GenericType="Customer" Data=customerRecords>
    <ItemTitle>@context.Name (@context.Email)</ItemTitle>
    <ItemBody>
       <ul>
       @foreach (var message in context.RecentMessages)
       {
           <li>@message</li>
       }
       </ul>
    </ItemBody>
</MyFancyList>
```

... people would choose the latter. Similarly for non-list scenarios, e.g., 

```
<MyDialog
    Header=@(@<div>Reorder @product.Name</div>)>
    Footer=@(@<div>
        <button onclick=@context.Accept>OK</button>
        <button onclick=@context.Dismiss>Close</button>
    </div>)>
    Some other markup here about @product
</MyDialog>
```

... versus:

```
<MyDialog>
    <Header>Reorder @product.Name</Header>
    <Footer>
        <button onclick=@context.Accept>OK</button>
        <button onclick=@context.Dismiss>Close</button>
    </Footer>
    <Body>
        Some other markup here about @product
    </Body>
</MyDialog>
```

Do you think there will be compelling scenarios for declaring templates inline in attributes? All this to say, let's chat about what those scenarios will be and whether the implementation will give us maintenance challenges as the templating feature is developed further in Razor itself.

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
SteveSandersonMS: Not sure whether people will use it, but it's totally fine to have it :)

src/Microsoft.AspNetCore.Blazor/RenderFragmentExtensions.cs
SteveSandersonMS: The approach looks fine and mirrors how we handle some other cases where we have a `RenderFragment` (for example, how we lift child content into a lambda that takes a `RenderTreeBuilder`). I know it costs a heap allocation for the delegate but we have no better alternative.

My personal preference would be to avoid the name `Bind` just because it's so overloaded. Maybe `WithValue`?

test/Microsoft.AspNetCore.Blazor.Build.Test/RenderingRazorIntegrationTest.cs
SteveSandersonMS: This is totally fine and matches how the sequence numbers look in a regular `@for` or `@foreach` loop. The diff algorithm notices when the sequence number steps back to an earlier number, and uses that to detect the case where leading or trailing items are added within a given iteration.

test/Microsoft.AspNetCore.Blazor.Build.Test/RuntimeCodeGenerationTest.cs
```diff
@@ -1269,5 +1269,219 @@ public void Component_WithDocType()
...
+}");
+
+            // Assert
+            AssertDocumentNodeMatchesBaseline(generated.CodeDocument);
+            AssertCSharpDocumentMatchesBaseline(generated.CodeDocument);
+            CompileToAssembly(generated);
+        }
+
+        [Fact]
+        public void RazorTemplate_InImplicitExpression()
```
SteveSandersonMS: The test cases for `InExplicitExpression` and `InImplicitExpression` appear to be identical. Am I missing something, or is this a copy-paste error?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComplexAttributeContentPass.cs
SteveSandersonMS: Let's discuss in the sync today.

test/Microsoft.AspNetCore.Blazor.Build.Test/ComponentRenderingRazorIntegrationTest.cs
SteveSandersonMS: They are inside a frame for `<Repeater>` - that's the first item in the collection. Its children are the output from the `Repeater` component.

What I don't understand is why the children aren't grouped in `Region` frames. Your new `AddContent` overload does emit region delimiters, so why don't they appear in the output here?

Liander:
I agree with @SteveSandersonMS  that having to type markup in the attribute looks very unfriendly and I would very much prefer the way @uazo did it with qualifying the property name with the component so you would get the tags `<MyDialog.Header>` and `<MyFancyList.ItemBody>` in your examples. 

@rynowak I saw that the type is defined as
`   public delegate void RenderFragment<T>(RenderTreeBuilder builder, T value);`
Have you considered defining it as 
`    public delegate RenderFragment RenderFragment<T>(T value);`
then I assume you don’t need any predefined variable name for item?,  and can define a template like:
`  @{ RenderFragment<Person> template = person => @<div>@person.Name</div>; }`

I would very much prefer that. Then a TemplateProperty(item) would be just a RenderFragment.

I would also prefer being able to set that variable name when giving it as a property-tag like: `<MyTitleBar.Body @params(user)>`. Here one would read the `@params` as a blazor directive for defining lambda parameters. See also my comments in https://github.com/aspnet/Blazor/issues/404#issuecomment-410556289.

src/Microsoft.AspNetCore.Blazor/RenderFragmentExtensions.cs
rynowak: sure

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeBuilder.cs
rynowak: I think we'll have to decide which of these two is the core experience, and whether we can avoid allocations in many cases. 

There's feedback elsewhere about a different signature for `RenderFragment<>` that would result in an allocation on each use, but has some nice experience improvments.

test/Microsoft.AspNetCore.Blazor.Build.Test/ComponentRenderingRazorIntegrationTest.cs
rynowak: OK I'll dig into this and try to understand more.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComplexAttributeContentPass.cs
```diff
@@ -73,17 +74,59 @@ private static bool TrySimplifyContent(IntermediateNode node)
...
+                        return false;
+                    }
+                }
+
+                // There's yet another special case when you have an attribute with a template
+                // like <div attr="@(@<div>Hi</div>)"
+                //
+                // This will look just like the above case except there are 4 tokens, and one of them is
+                // an empty C# token.
+                {
```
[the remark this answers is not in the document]
rynowak: These are all artifacts of using templates in an attribute, since we're removing that this will go away as well.

rynowak:
thanks for the feedback about the attributes. Perhaps I should have made it clear earlier, but putting templates inside attributes is not a counter proposal to having them inside child-elements, it was something extra. It also requires us to fight with a lot of existing infrastructure, so we're cutting it.

This PR doesn't attempt to address adding templates as child elements, we'll be doing that in the near future.

>  I saw that the type is defined as `public delegate void RenderFragment<T>(RenderTreeBuilder builder, T value);` Have you considered defining it as `public delegate RenderFragment RenderFragment<T>(T value);`
then I assume you don’t need any predefined variable name for item?, and can define a template like:
`@{ RenderFragment<Person> template = person => @<div>@person.Name</div>; }`

I briefly considered it earlier, and avoided it because it guarantees an allocation for every usage of every template. I agree that it has some usability benefits. 

Trying to figure out how to let you name the lambda parameter(s) is something that the team spent a lot of time discussing, and we didn't generate any great ideas based on the `(RenderTreeBuilder, T) => void` signature. There are real limitations in what Razor can do when we have to parse/understand C# constructs., which is what we were trying to avoid.

Doing something like: `@{ RenderFragment<Person> template = person => @<div>@person.Name</div>; }`  is a nicer experience and seems really achievable., but it's got two things against it:
1. Requires an allocation every time we use it
2. This is a break from how templates work in vanilla razor. This feature exists already in Razor, and dates back to the original 1.0 release http://vibrantcode.com/blog/2010/8/2/inside-razor-part-3-templates.html - it's just not that popular in vanilla Razor

We'll discuss again and see if this makes more sense, thanks for the idea. 

> I would also prefer being able to set that variable name when giving it as a property-tag l

We had something similar in mind.

rynowak:
> This is a break from how templates work in vanilla razor. This feature exists already in Razor, and dates back to the original 1.0 release http://vibrantcode.com/blog/2010/8/2/inside-razor-part-3-templates.html - it's just not that popular in vanilla Razor

Actually I'm going to disagree with myself. It's not a breaking change from vanilla Razor's semantics if we define templates `@<div>...</div>` as creating a `RenderFragment` (`Action<RenderTreeBuilder>`), instead of directly creating a `RenderFragment<T>`.

The generated code for this example looks like:
```
@{ RenderFragment<Person> template = person => @<div>@person.Name</div>; }
...
RenderFragment<Person> template = (person) => ((__builder) => { ... });
```

Which means that you can do 
```
@{ RenderFragment template=> @<div>...</div>; }
...
RenderFragment template = ((__builder) => { ... });
```

This seems like a much more powerful construct 👍.

test/Microsoft.AspNetCore.Blazor.Build.Test/ComponentRenderingRazorIntegrationTest.cs
rynowak: So from what I can understand the `GetRenderTree` method used by these tests will never include regions, since it's the list of edits that would get applied to an 'empty' tree. I can confirm that the regiions are part of the reference frames, but aren't part of the 'render batch'. This is as far as I got because the terms are all a bit abstract to me. 

We don't have any tests written in this style that include/verify regions. All of the tests that do verify regions get them directly from `RenderTreeBuilder.GetFrames()` with the exception of PassesThroughRegionsInsidePrependedElements

Let me know if you think this needs more investigation or follow up.

Liander:
I agree.

You said you would cut having templates inside attributes, but if you would use this mechanism would it look like `<MyCard template="@(person => @<div>@person.Name</div>")>` and you don't have that empty c#-token?

rynowak:
Appveyor failure appears to be the one we've been seeing regularly related to Kestrel, ignoring.

rynowak:
@Liander it's kinda academic. If you did templates in attributes and also did that particular definition of templates it would look like that. 

> and you don't have that empty c#-token?

I'm not sure what you mean by that.. care to explain?

Liander:
You can ignore that comment. I probably used wrong terminology there. I just meant that I saw that an extra `@( )` was present when put at the attribute level `@(@<div>…</div>)` compared to in code which looked a bit weird. It looks less weird though if it would encapsulates a lambda call `@(item => @<div>..</div>)`.

Liander:
> Requires an allocation every time we use it

@rynowak Let me see if I understand this. The concern is that `RenderFragment<Person> template = (person) => ((__builder) => { ... }); ` whould allocate delegate objects for every rendering. Is that correct?

The delegate is completely determined from the person object and the __builder object you pass in, meaning the assignment will receive the same function object for each component instance each rendering. Maybe I am missing something but wouldn’t the construction `RenderFragment<Person> template = template ?? (person) => ((__builder) => { ... }` resolve that? So it will only be evaluated in the first rendering. (And if it was an ItemTemplate in a list view component it will be resolved just once independent of having many items.) 
What do you think?

rynowak:
> The delegate is completely determined from the person object and the __builder object you pass in, meaning the assignment will receive the same function object for each component instance each rendering. 

The issue here is that the more expanded signature is `Func<Person, Action<RenderTreeBuilder>>`. You need to pass in a `Person` object, return a new closure/delegate of type `Action<RenderTreeBuilder>` which can be executed. 

You can't cache anything in this case without breaking the user experience, because the new object (in practical cases) **always** captures the value of the `Person` that was passed in. If you cached, you would see the template evaluated for the same person object over and over.

I still think this idea is worth considering, but that was my original motivation for the other signature.

Liander:
Sorry if I am not following... but if you cache `Func<Person, Action<RenderTreebuilder>>` it is that function that is chached and it will render differently with feeding it with different person objects. What am I missing? Is the call to it causing allocations?

rynowak:
> Reading again I think I understand. It will be another allocation for the action with new closure.

yes, exactly. We need to call the `Func<Person, Action<RenderTreebuilder>>` each time which will allocate each time.

rynowak:
Will be working that as one of the next steps.... 👍

RyoukoKonpaku:
@rynowak that's great news :smile:!

Liander:
@rynowak Thanks for replying, I really appreciate what you and the team are doing. It might be disruptive with my questions, but here we go again…

In the light of previous discussion and efficiency in general, it might be worth considering having the ability to tell if the input to the template has changed. So maybe add something that allows providing an optional comparer, like 
```
   public class TemplateRender<T>
   {
      public RenderFragment<T> TemplateFragment;
      public RenderFragment RenderFragment; // In same class for brevity
      public Func<T, T, bool> When;

      public TemplateRender<T> Render(T value) 
          { RenderFragment = TemplateFragment(value); return this; }
   }
```
So you can provide both a RenderFragment and an optional predicate condition:
```
@{ TemplateRender<Person> template = TemplateRender.Create(
   person => @<div>@person.Name</div>, 
   (p1, p2) => p1.Name != p2.Name); }
```
And use it as `@(template.Render(person))`. Or when applied in a child content property-tag, 
```
  <MyContactList.ItemTemplate $params(person) when=”@((p1, p2) => p1.Name != p2.Name))”>
  	<div>@person.Name</div>
  </MyContactList.ItemTemplate>
```
As for some implementation thoughts… In cases you have a comparer you could add some “LazyContentFrame” that can compare the input object using the comparer and that builds its child frames during the diff process. The child frames can either be captured from the old tree when there is no status change (just for the sake of generating the complete tree ones the diff is completed, even though it might not be needed) otherwise build a new local set of frames array and continue diffing the children.

Well… I felt I needed to share this thought.

SteveSandersonMS:
@Liander Knowing whether to re-render is essentially what factoring out into a component achieves. Blazor components make you be explicit about their inputs (i.e., `[Parameter]`) so by default their will re-render only when those inputs change. You can also override the "when to re-render" logic explicitly by overriding `ShouldRender`, which is similar to the optional comparer you've described above.

I recognise this could *also* be provided as a template-level feature, but we'll have to wait and see if there turn out to be compelling use cases for that, as maybe components are the right level of abstraction for this already.

Liander:
@SteveSandersonMS  I agree. Nothing to bother about.

I think people would prefer not being needed to factor out into a component when using templates inline. But thinking about it again I realize one should be able to achieve the same with some generic utility component that handles the condition for rendering, like 
```
<MyContactList.ItemTemplate $params(person)>
   <ConditionalRender value=”$(person)” when=”@((p1, p2) => p1.Name != p2.Name))>
  	<div>@person.Name</div>
   </ConditionalRender>
</MyContactList.ItemTemplate>
```
(One could also let the component itself that has the template parameter apply the render-wrapper, so in this case it would be that MyContactList would take a predicate for item rendering and apply the conditional-render-wrapper around the item template when using it.)

Thanks for taking the time to discuss. 😊

PR #1372, opened 2018-08-30
Enable default HttpClient handler

SteveSandersonMS:
This uses the feature implemented by @kjpou1 https://github.com/mono/mono/pull/10152. It means you can now do `new HttpClient()` directly and it works under Blazor on WebAssembly.

Note that we still register a `HttpClient` in the default DI config, because that instance has its `BaseUri` populated, meaning that it knows how to make requests with relative URLs. If you do `new HttpClient()` then you can only make requests to absolute URLs. This is standard for .NET's `HttpClient`.

Additionally, this involved updating to a newer version of mono (bfc35fc6e65). Benchmarks show that this newer version of Mono is consistently 10-20% faster on our checked scenarios (app startup, rendering, JSON handling), but it's also about 150KB bigger after linking.


SteveSandersonMS:
cc @danroth27 because I know he's been waiting a while for this.

SteveSandersonMS:
@rynowak I expect the only parts you'll want to review are the changes to the app startup process. Feel free to review the rest if you want, but it's mostly just the Mono update.

kumpera:
@SteveSandersonMS you can use this new mechanism to return a properly configured HttpClient.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/WebAssemblyHost.cs
```diff
@@ -97,5 +101,27 @@ public void Dispose()
...
+            // running under Mono WebAssembly
+            if (getHttpMessageHandlerField != null)
+            {
+                // Just in case you're not actually using HttpClient, defer the construction
+                // of the BrowserHttpMessageHandler
+                var handlerSingleton = new Lazy<HttpMessageHandler>(
+                    () => new BrowserHttpMessageHandler());
+                Func<HttpMessageHandler> handlerFactory = () => handlerSingleton.Value;
+                getHttpMessageHandlerField.SetValue(null, handlerFactory);
+            }
```
rynowak: Does it make sense for us to make a lot of noise if this fails? I'm not sure we have a good way to report it other than logging to the console, but it seems worth talking about.

src/Microsoft.AspNetCore.Blazor.Browser/Hosting/WebAssemblyHostBuilder.cs
```diff
@@ -75,7 +74,7 @@ private void CreateServiceProvider()
             {
                 // Creating the URI helper needs to wait until the JS Runtime is initialized, so defer it.
                 var uriHelper = s.GetRequiredService<IUriHelper>();
-                return new HttpClient(new BrowserHttpMessageHandler())
```
rynowak: 🎆 🍺

src/mono/HowToUpgradeMono.md
```diff
@@ -80,15 +71,13 @@ Currently the build reports these warnings:
 
  * `EXEC : warning : unresolved symbol: putchar` (during both wasm and asm.js builds)
  * `EXEC : warning : root:BINARYEN_ASYNC_COMPILATION disabled due to user options` (during the asm.js build only)
- * `EXEC : warning : unresolved symbol: mono_wasm_invoke_js_with_args` (during both wasm and asm.js builds). I expect this is a temporary issue that will go away when Mono finishes implementing whatever they are doing with their JS interop, which we aren't currently using anyway since we have our own interop in `driver_blazor.c`.
- * `EXEC : warning : unresolved symbol: mono_sdb_single_step_trampoline` (during both wasm and asm.js builds). I expect this will eventually be implemented as part of the debugging work.
+ * `EXEC : warning : unresolved symbol: mono_wasm_invoke_js_with_args / mono_wasm_get_object_property / mono_wasm_set_object_property` (during both wasm and asm.js builds). I expect this is due to us not including Mono's `binding_support.js` in the build, which we could do but don't presently require.
```
kjpou1: @SteveSandersonMS,   Yes these are due to not including the `binding_support.js`.

SteveSandersonMS:
@kumpera I thought that's what we already are doing in this PR. Or did you have something else in mind?

SteveSandersonMS:
I'll merge this once the rebased branch build passes, but it's worth noting that once this goes in, the default app size will (temporarily) increase from 1.9MB to 2.5MB. This is due to it now including all the `HttpClientHandler` code paths (which transitively are pretty huge, and we don't even use them at runtime), which in turn is because we now call the default `HttpClient` constructor.

As a fix for the size bloat, I'm about to work on adding some infrastructure for pre-stripping parts of the BCL that we never want to use at runtime in WebAssembly builds. This will chop out `HttpClientHandler` completely (as it will never work, and is never required, on WebAssembly), giving us back that 500KB. It also makes it possible to strip out more - there might be some other easy wins to shrink typical apps further.

SteveSandersonMS:
@kumpera How can we configure `new HttpClient()` to have a preset `BaseUri`?

kumpera:
I guess I sound really stupid here, I thought it was possible to do it from HttpClientHandler.
If that's not true, we can change the callback to allow it to do so.

SteveSandersonMS:
No problem - it's fine as-is!

PR #1382, opened 2018-09-04
Pre-trim Mono BCL

SteveSandersonMS:
This PR adds infrastructure for trimming the Mono WebAssembly BCL as part of our build, so the BCL we distribute gets smaller (especially after linking). The main aspects are:

* A new build tool command, `ilwipe`, that takes a set of "spec" files and uses them as a config to know which parts of the BCL to "wipe"
   * *wipe* means "replace specified method bodies with a single `throw` instruction". Doing this (instead of actually removing the method entirely) means that the assembly retains a completely standard API surface, and if you try to use one of the wiped methods, you get an easy-to-understand exception stack trace that tells you which wiped method you're trying to call.
   * Spec files are just text files, each line of which is either whitespace, a comment, or a descriptor of something to wipe. Descriptors are of the form `Namespace.Type[::MethodName[(argtypes)]]`, where method name and arg types are optional (if not set, means "all"). Also supports wildcards (e.g., `System.Reflection.*`) and negation (e.g., `!System.Reflection.MethodInfo::ToString`) to override earlier lines.
   * When the linker processes wiped (near-empty) methods such as these, it won't mark their orignial dependencies for retention, so even if the "wiped" assembly isn't much smaller, the overall linker result may be dramatically smaller
* Build config changes that run `ilwipe` as part of the build. I also found numerous places where our build config was doing odd things or depending on not-actually-guaranteed build-order coincidences, so have (hopefully) tidied this up.

With the wipe config in this PR, the out-of-box standalone app goes down from 2.5MB to 2.0MB, i.e., it undoes the extra bloat introduced in #1372. I expect there are plenty more easy opportunities for wiping out other large would-never-work-anyway areas of the BCL, but that is a job for later.


src/Microsoft.AspNetCore.Blazor.Browser.JS/Microsoft.AspNetCore.Blazor.Browser.JS.csproj
```diff
@@ -5,18 +5,20 @@
     <TypeScriptCompileBlocked>true</TypeScriptCompileBlocked>
     <TypeScriptToolsVersion>Latest</TypeScriptToolsVersion>
     <DefaultItemExcludes>${DefaultItemExcludes};node_modules\**</DefaultItemExcludes>
+    <IsPackable>false</IsPackable>
```
SteveSandersonMS: Since the beginning, we've been shipping `Microsoft.AspNetCore.Blazor.Browser.JS` to NuGet.org, but nothing actually depends on it. It only exists to run as part of our build in this repo. As far as I can tell, we shouldn't be shipping this as a package to NuGet.

src/Microsoft.AspNetCore.Blazor.Build/targets/Blazor.MonoRuntime.props
```diff
@@ -25,13 +25,4 @@
...
   </PropertyGroup>
 
-  <ItemGroup>
-    <MonoBCLFile Include="$(MonoBaseClassLibraryPath)*.dll" />
-    <MonoBCLFacadeFile Include="$(MonoBaseClassLibraryFacadesPath)*.dll" />
-    <MonoAsmjsFile Include="$(MonoAsmjsRuntimePath)**/*.*" />
-    <MonoWasmFile Include="$(MonoWasmRuntimePath)**/*.*" />
-    <BlazorJsFile Include="$(BlazorJsPath)" />
-    <MonoBaseClassLibraryFolder Include="$(MonoBaseClassLibraryPath);$(MonoBaseClassLibraryFacadesPath)" />
-  </ItemGroup>
```
SteveSandersonMS: These were being evaluated too early. It only worked before as a random fluke of arbitrary build ordering. It no longer works following other build cleanups elsewhere. So now I've moved them to be evaluated later at a more suitable time.

src/Microsoft.AspNetCore.Blazor.BuildTools/Core/ILWipe/MethodWipedExceptionMethod.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            // Adds the following method to the assembly:
+            // namespace ILWipe
+            // {
+            //     internal static class ILWipeHelpers
+            //     {
+            //         public static Exception CreateMethodWipedException()
+            //         {
+            //             return new NotImplementedException("Cannot call method because it was wiped. See stack trace for details.");
+            //         }
```
SteveSandersonMS: Encapsulating this logic in a shared method means that all the places where it's used (the "wiped" methods) involve fewer repeated IL instructions. Plus it saves us the trouble of declaring a string const to avoid repeating that.

src/mono/bclwipe/mscorlib.txt
```diff
@@ -0,0 +1 @@
+# Nothing to do here yet
```
SteveSandersonMS: Pretty sure we'll add something here soon. For now I just wanted this file to verify we could process multiple spec files.

src/mono/mono.csproj
```diff
@@ -0,0 +1,80 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <!--
+    This project is not really a .NET package. It's just defined as a csproj to simplify
+    the build config. In reality the output assembly is empty and ignored.
```
SteveSandersonMS: I converted `mono.targets` to `mono.csproj` because it simplifies the incremental build.

src/Microsoft.AspNetCore.Blazor.Browser.JS/Microsoft.AspNetCore.Blazor.Browser.JS.csproj
rynowak: Sounds good.

Issue #261, opened 2018-03-14
Standalone dev server can't run app in Release mode


SteveSandersonMS:
Currently, the standalone dev server in `Microsoft.AspNetCore.Blazor.Cli` is hardcoded to look for app binaries in `(sourceRoot)\bin\Debug`. If you're launching from the command line, it isn't aware of any `-c Release` flag, and if you're launching from VS, it doesn't know which solution configuration you have enabled.

At the moment although I have an idea of how we could fix it for the command line, I'm not sure how we can make it aware of the active VS solution configuration.

The main limitation seems to be that we're using the Web SDK. If the standalone app didn't use the Web SDK and was instead more like a pure console app, then on startup the process working directory would be set to the client app's active `bin\(Debug|Release)\(framework)` directory, which would be ideal because then we'd know precisely which `dist` directory to serve. Unfortunately this doesn't work if you use the web SDK because VS *always* sets the working directory to the project root if you're using the Web SDK (https://github.com/dotnet/cli/issues/4473#issuecomment-256674279).

In the Web SDK source, there's [a flag called `EnableDefaultRunWorkingDirectory`](https://github.com/aspnet/websdk/blob/dev/src/ProjectSystem/Microsoft.NET.Sdk.Web.ProjectSystem.Targets/netstandard1.0/Microsoft.NET.Sdk.Web.ProjectSystem.targets#L22) that you can set to `false` to cause the working directory to be the bin dir (i.e., `bin\(Debug|Release)\(framework)`). That would be perfect, except it seems that VS doesn't know about this flag and still uses the project root anyway.

Putting this on backlog because it's not an insane limitation that the dev server only supports Debug builds (after all, it is the *dev* server).

@BillHiebert Do you know:

* If there's any way to make VS launch web projects with the CWD set to the active bin folder instead of the project root directory?
* And if not, is there any way for a launched executable (set via `<RunCommand>`) to know which solution configuration is currently active?
* And if not, might it be possible to extend the VS tooling to respect the `EnableDefaultRunWorkingDirectory` flag in the future?

Thanks!

SteveSandersonMS:
**Update**: I've made the dev server assert that only Debug *or* Release dirs exist, and pick whichever one does. If you have both, it refuse to start (and says why). We still need a proper fix though!

torhovland:
[...]

PR #1383, opened 2018-09-04
Add -c option in the blazor server

kant2002:
- Pass configuration using MSBuild variable $(Configuration)
- Parse command line parameters -c or --configuration to match dotnet command way of defining configuration

See #261


danroth27:
Thanks for the contribution! We think there might be a better way to do this using launch settings. Our preferred solution would be to somehow specify the output directory instead of the configuration. We'll need to investigate to see how feasible this is. 

I'm going to go ahead and close this PR as we don't expect to take the currently proposed approach. Let us know though if you're interested in looking into a launch settings based approach.

Issue #1324, opened 2018-08-16
VS 15.8: New project from template results in empty solution


arpadbarta:
After upgrading VS to 15.8 and .Net Core to 2.1.400 newly created project with the Blazor template results in empty solutions. Trying to add generated projects fails with incomplete project files.

Changing the .Net Core version in global.json to 2.1.400 solves the problem and generated projects can be added.

danroth27:
I'm not able to reproduce this on my machine.

@mlorbetske Any ideas on what could cause this?

ShadowDancer:
@danroth27
Template creates `global.json` file with  `"version": "2.1.300"`, however if You updated visual studio You have 2.1.400, and You cannot install 2.1.300 (installer offers repair). If You change version to `"2.1.400"` solution loads without problems.

MarkStega:
I can't remember if I deleted the file or if it wasn't generated but when I tried to create a new ASP.Net Core project with either Blazor template I get a global.json file in the root of the solution that has
```
{
  "sdk": {
    "version": "2.1.300"
  }
}
```
@danr I bet you have the 2.1.300 sdk installed on your system. I do not.

danroth27:
> @danroth27 I bet you have the 2.1.300 sdk installed on your system. I do not.

Yeah, I have more SDK versions installed than I care to admit 😝. Let me try uninstalling 2.1.300 and see what that does.

danroth27:
OK, I'm seeing this issue now with the multi-project templates.

mlorbetske:
@phenning any ideas?

danroth27:
I looked at this with @phenning and it seems to be caused by the *global.json* file that we drop with the template. It only affects the multi-project templates though.

phenning:
Note that we do actually create the individual projects, and that manually updating global.json to point to 2.1.400, or deleting it allows the projects to be loaded.

danroth27:
At this point I think we should remove the *global.json* files from our templates. I don't think they are adding any value.

River39:
editing the global.josn and settings the "version" to "2.1.400" fixed the issue for me.

danroth27:
This should now be fixed by https://github.com/aspnet/Blazor/pull/1386

PR #1386, opened 2018-09-04
Removal of global.json from the template

kant2002:
I remove only global.json from the template, and leave global.json in the repository root, since it is required for the build.
See #1324


SteveSandersonMS:
Thanks, @kant2002!

PR #1387, opened 2018-09-04
Support streaming HTTP responses and HttpCompletionOption.ResponseHeadersRead

### LINKED ISSUE #1384: BrowserHttpMessageHandler reads body content to end
Inside "Blazor/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts", the response data is handled via " responseData = await response.arrayBuffer();", which reads the entire response stream to the end, and then returns the data, however this is the incorrect behaviour.

HttpMessageHandler's should not buffer the response as this is handled within the HttpClient.

This means that method such as "_httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, CancellationToken.None);", are not handled correctly as the ResponseHeadersRead flag is not respected, due to the content body being read to the end.

We would need to wrap the ReadableStream from the JS side to the c# side as the HttpContent in the ReceiveResponse in BrowserHttpMessageHandler.

Is there any chance of getting this fixed? As I've managed to get the c# signalr client working with long polling, but can't get the "ServerSentEventsTransport" to work until this is fixed, as otherwise the client timeouts trying to read the content body to the end, when it shouldn't as the ResponseHeadersRead is passed.

Suchiman:
This delays reading the response body as  `ArrayBuffer`  by moving it into a custom `HttpContent` type which allows `httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead` to work as intended.
Additionally this introduces optional support for streaming the response body if supported by the browser and enabled by the user (controlled via `BrowserHttpMessageHandler.StreamingSupported` and `BrowserHttpMessageHandler.StreamingEnabled`).
Streaming is not enabled by default because the default behavior of `HttpClient` is to buffer anyway while the streaming mode increases the overhead (many interop transitions, Tasks allocated for every `ReadAsync` call).
Last but not least a small change to cascade the `CancellationToken` to an `abortController` on the JS side so an in-flight fetch request can be properly canceled.
fixes #1384

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts
@@ -3,18 +3,22 @@ import { MethodHandle, System_String, System_Array } from '../Platform/Platform'
 const httpClientAssembly = 'Microsoft.AspNetCore.Blazor.Browser';
 const httpClientNamespace = `${httpClientAssembly}.Http`;
 const httpClientTypeName = 'BrowserHttpMessageHandler';
+const httpContentTypeName = 'BrowserHttpContent';
 const httpClientFullTypeName = `${httpClientNamespace}.${httpClientTypeName}`;
+const pendingResponses = new Map<number, Response>();
+let receiveResponseDataMethod: MethodHandle;
 let receiveResponseMethod: MethodHandle;
 let allocateArrayMethod: MethodHandle;
 
 // These are the functions we're making available for invocation from .NET
 export const internalFunctions = {
-  sendAsync
+  sendAsync,
+  getResponseData,
+  discardResponse
 }
 
 async function sendAsync(id: number, body: System_Array<any>, jsonFetchArgs: System_String) {
   let response: Response;
-  let responseData: ArrayBuffer;
 
   const fetchOptions: FetchOptions = JSON.parse(platform.toJavaScriptString(jsonFetchArgs));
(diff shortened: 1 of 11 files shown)


Suchiman:
I'm currently working on feature detected streaming support (on top of these changes) to really fix #1384 and have a working prototype but i need to get some sleep first.

Suchiman:
This is ready for review now.
`BrowserHttpMessageHandler.StreamingSupported` can be used to detect browser support and streaming can be turned on (if supported) via `BrowserHttpMessageHandler.StreamingEnabled`.
Any idea how i can best run the Http Unit tests a second time (once with streaming off, once with streaming on?)

SteveSandersonMS:
Thanks @Suchiman for contributing this!

I hope it's OK, but I'm going to leave it until after the 0.6.0 release before going through it in detail. That's because it's quite a large change and will need some intricate consideration.

There's also the matter that it depends on a feature that not all browsers support, and we have to make a strategic choice about whether that is the sort of feature we want to include by default. Currently all the .NET BCL features work the same on all browsers, so this would be a substantial change of design.

If BTW you can think of any extra hooks we could put in so that this functionality could be enabled as a third-party NuGet package, that would of course simplify the matter a lot, because we could freely add such extensibility points without raising the problem of consistency across browsers.

Suchiman:
@SteveSandersonMS i don't know if you still wanna take this, i have upstreamed the changes here https://github.com/mono/mono/pull/11255#event-1931830193 if you wanna switch to mono's Version.

danroth27:
Thanks @Suchiman! We expect to pull this in via Mono for our 0.7.0 release (#1546). I assume we're ok to close this.

PR #1396, opened 2018-09-07
Refactor server-side blazor startup to allow Azure SignalR. Fixes #1227

### LINKED ISSUE #1227: Option to plug Azure SignalR in server side rendering.

SteveSandersonMS:
Fixes #1227

The main challenge with this is:
(1) We don't want to take a strict dependency on the `Microsoft.Azure.SignalR` package (at least not for everyone)
(2) The `Microsoft.Azure.SignalR` package uses its own custom equivalent to `MapHub` that looks the same as the standard `Microsoft.AspNetCore.SignalR` one, but actually uses different types. So if we comply with (1), then we can't possibly call `MapHub` on the developer's behalf.

Possible resolutions, then:

* A: Use reflection tricks and make it more magic (no thanks)
* B: Have a separate `Microsoft.AspNetCore.Blazor.AzureSignalR` package that depends on `Microsoft.Azure.SignalR`
* C: Let the developer call `MapHub` themselves, meaning that (a) we have to remove all the other magic from `UseServerSideBlazor` and (b) we have to make `BlazorHub` public.

Between B and C, I prefer option C because it makes us less tightly coupled to SignalR APIs, so if SignalR adds more overloads to `MapHub` or other config styles in the future, we should automatically be composable with it.

This PR implements options C.

### Result

The result is that `app.UseServerSideBlazor<TClient>();` is now exactly equivalent to:

```csharp
app.UseSignalR(route => route.MapHub<BlazorHub>(BlazorHub.DefaultPath));
app.UseBlazor<TClient>();
```

So, if a developer wants, they can choose *not* to call `UseServerSideBlazor`, but rather use those two lines instead, and from there they can replace `UseSignalR` with `UseAzureSignalR` if they want.

Therefore the way to use Azure SignalR with a server-side Blazor app becomes (in the server app's `Startup.cs`):

```csharp
    public class Startup
    {
        {
            services.AddSignalR().AddAzureSignalR(/* optionsOrConnectionString */);
            services.AddServerSideBlazor<Client.Startup>();
            // ...
        }

        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            app.UseAzureSignalR(route => route.MapHub<BlazorHub>(BlazorHub.DefaultPath));
            app.UseBlazor<Client.Startup>();
            // ...
        }
    }
```

Note: This means that `AddSignalR` is being called twice - once by the developer manually, and once inside `AddServerSideBlazor`. @anurse, do you know if there's any problem or disadvantage with calling that twice? We can stop it if we absolutely must, but it will be easier for developers to understand this way (because it works both with and without Azure SignalR with a minimum of user code changes).

### More notes, mainly for @rynowak

I moved all the "register startup actions" logic out of `UseServerSideBlazor` and into `AddServerSideBlazor` (so that `UseServerSideBlazor` can be equivalent to the two lines above). I believe the net result is the same, and it still can be extended to support registering multiple server-side Blazor endpoints in the future (with different `TClient` values), but for now I still left it hard-coded to use `/_blazor` as the path without implementing support for configuring an alternate path.


src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
```diff
@@ -2,32 +2,36 @@
...
+            _scopeFactory = scopeFactory ?? throw new ArgumentNullException(nameof(scopeFactory));
+            _options = options.Value;
         }
 
-        public Dictionary<PathString, Action<IBlazorApplicationBuilder>> StartupActions { get; }
-
         public override CircuitHost CreateCircuitHost(HttpContext httpContext, IClientProxy client)
         {
-            if (!StartupActions.TryGetValue(httpContext.Request.Path, out var config))
+            if (!_options.StartupActions.TryGetValue(httpContext.Request.Path, out var config))
```
SteveSandersonMS: I moved the `StartupActions` onto a new config class so we can use the DI system's `Configure` callbacks to register multiple startup actions (one for each `AddServerSideBlazor<TClient>` call). For now you can only call that once, because there's no way to specify distinct paths, but in the future we could have `AddServerSideBlazor` take an optional "path" arg.

src/Microsoft.AspNetCore.Blazor.Server/ServerSideBlazorOptions.cs
```diff
@@ -1,12 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace Microsoft.AspNetCore.Blazor
-{
-    /// <summary>
-    /// Options for Server-Side Blazor.
-    /// </summary>
-    public class ServerSideBlazorOptions
```
SteveSandersonMS: I killed this because we're not doing anything with it yet, and there's a good chance that if we do add options in the future, they will mostly go to `AddServerSideBlazor` instead of `UseServerSideBlazor`. In any case it's unclear, so I'd prefer to drop it now and re-add something later when we know what it will be.

anurse:
> @anurse, do you know if there's any problem or disadvantage with calling that twice?

I do not believe this will cause a problem. We use `TryAdd` to register things in DI so the second time should just no-op (other than giving you an `ISignalRBuilder` to register additional stuff).

anurse:
I spoke a little too soon. There are two things that will be duplicated in the DI container if both you **and the user** call `.AddSignalR`. But I think we're OK:

* `SignalRMarkerService` - This is just used to detect that you've called `AddSignalR` when you call `UseSignalR` (like MVC), duplicates here should be fine.
* `HubOptionsSetup` - This establishes the defaults for `IOptions<HubOptions>`. Two instances of this should be fine, it'll just mean we set the options twice.

I'll file a bug to make those `TryAdd`, there's not really a reason they can't be. But I think this change can go forward without changing them since there isn't really a negative impact to the doubling up.

src/Microsoft.AspNetCore.Blazor.Server/Builder/BlazorApplicationBuilderExtensions.cs
```diff
@@ -38,9 +39,10 @@ public static class BlazorApplicationBuilderExtensions
...
         /// </summary>
-        /// <param name="app"></param>
-        /// <param name="options"></param>
-        public static void UseBlazor(
+        /// <param name="app">The <see cref="IApplicationBuilder"/>.</param>
+        /// <param name="options">Options to configure the middleware.</param>
+        /// <returns>The <see cref="IApplicationBuilder"/>.</returns>
+        public static IApplicationBuilder UseBlazor(
             this IApplicationBuilder app,
             BlazorOptions options)
```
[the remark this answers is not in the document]
rynowak: Usually with middleware we just accept the options as-is becuase the options is constructed by the user, and not from DI. `UseBlazor()` would use the options from DI.

`AddBlazor(...)` would take an action, since that's going to compose with DI.

src/Microsoft.AspNetCore.Blazor.Server/Builder/ServerSideBlazorApplicationBuilderExtensions.cs
```diff
@@ -20,94 +19,57 @@ public static class ServerSideBlazorApplicationBuilderExtensions
...
-            UseServerSideBlazorCore(builder, configure);
-
-            builder.UseBlazor(new BlazorOptions()
-            {
-                ClientAssemblyPath = startupType.Assembly.Location,
-            });
-
-            return builder;
+            // TODO: Also allow configuring the endpoint path.
+            return UseSignalR(builder, BlazorHub.DefaultPath)
```
rynowak: This would be like 5 LOC to add 😀

src/Microsoft.AspNetCore.Blazor.Server/Builder/ServerSideBlazorApplicationBuilderExtensions.cs
```diff
@@ -20,94 +19,57 @@ public static class ServerSideBlazorApplicationBuilderExtensions
...
+                .UseBlazor(options);
         }
 
-        private static IApplicationBuilder UseServerSideBlazorCore(
-            IApplicationBuilder builder,
-            Action<IBlazorApplicationBuilder> configure)
+        private static IApplicationBuilder UseSignalR(IApplicationBuilder builder, PathString path)
```
rynowak: nit: different name? this makes the call sites look like a call to the vanilla UseSignalR, but it's totally not.

src/Microsoft.AspNetCore.Blazor.Server/DependencyInjection/ServerSideBlazorServiceCollectionExtensions.cs
```diff
@@ -75,101 +60,68 @@ public static IServiceCollection AddServerSideBlazor(this IServiceCollection ser
...
-        {
-            if (services == null)
-            {
-                throw new ArgumentNullException(nameof(services));
+                // Configure the circuit factory to call a startup action when each
+                // incoming connection is established. The startup action is "call
+                // TStartup's Configure method".
+                services.Configure<DefaultCircuitFactoryOptions>(circuitFactoryOptions =>
+                {
+                    var endpoint = BlazorHub.DefaultPath; // TODO: allow configuring this
```
rynowak: I'm not suggesting that we fix it in this PR, but this is leading us to a design where you specify the 'path' in both `ConfigureServices` and `Configure`. 

One idea that I had about this `CFOptions` would need a map from path -> type, and type->startup. So during `Add...` we map the type->startup, and during `Use...` we map the path->type

Again, I don't expect that we'll solve it as part of this change, but I think we'd like to prevent you from having the specify the path in two places.

src/Microsoft.AspNetCore.Blazor.Server/DependencyInjection/ServerSideBlazorServiceCollectionExtensions.cs
```diff
@@ -75,101 +60,68 @@ public static IServiceCollection AddServerSideBlazor(this IServiceCollection ser
...
-        private static IServiceCollection AddServerSideBlazorCore(
-            IServiceCollection services,
-            Type startupType,
-            Action<ServerSideBlazorOptions> configure)
+        private static void AddStandardServerSideBlazorServices(IServiceCollection services)
         {
+            // Here we add a bunch of services that don't vary in any way based on the
+            // user's configuration. So even if the user has multiple independent server-side
+            // Blazor entrypoints, this lot is the same and repeated registrations are a no-op.
             services.AddSingleton<CircuitFactory, DefaultCircuitFactory>();
```
rynowak: We should think about these all being `TryAdd` so that we don't register many instances. Generally we want framework code to be idempotent, and I think we expect users to call `AddBlazor<T>` multiple times for some scenarios.

test/testapps/TestServer/Startup.cs
```diff
@@ -39,7 +40,14 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
             {
-                subdirApp.UseServerSideBlazor<BasicTestApp.Startup>();
+                // The following two lines are equivalent to:
+                //     subdirApp.UseServerSideBlazor<BasicTestApp.Startup>();
+                // However it's expressed using UseSignalR+UseBlazor as a way of checking that
+                // we're not relying on any extra magic inside UseServerSideBlazor, since it's
+                // important that people can set up these bits of middleware manually (e.g., to
+                // swap in UseAzureSignalR instead of UseSignalR).
+                subdirApp.UseSignalR(route => route.MapHub<BlazorHub>(BlazorHub.DefaultPath));
+                subdirApp.UseBlazor<BasicTestApp.Startup>();
```
rynowak: I love it. I 100% agree with the analysis and I think this is the right solution.

rynowak:
@anurse - if you want to nuke this kind of problem from orbit, consider doing something like https://github.com/aspnet/Mvc/blob/release/2.2/test/Microsoft.AspNetCore.Mvc.Test/MvcServiceCollectionExtensionsTest.cs#L54

We unit test this in MVC, and it's a little annoying to maintain, but we know for sure that everything we do only happens once 😬

rynowak:
MVC doesn't have the ability to 'discover' public things for much the same reason signalR doesn't. I think there's a strong case to be made for 'discovery' to be based on public visibility, but manual wireups to support any visibility.

src/Microsoft.AspNetCore.Blazor.Server/BlazorHub.cs
```diff
@@ -3,40 +3,58 @@
...
+        /// </summary>
+        public static PathString DefaultPath => "/_blazor";
+
+        private CircuitHost CircuitHost
         {
             get => (CircuitHost)Context.Items[CircuitKey];
             set => Context.Items[CircuitKey] = value;
         }
 
+        /// <inheritdoc />
```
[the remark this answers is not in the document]
SteveSandersonMS: Also I do think we should "fix" this in any places where it meaningfully disadvantages a user by failing to show important docs.

src/Microsoft.AspNetCore.Blazor.Server/BlazorHub.cs
```diff
@@ -3,40 +3,58 @@
...
-
         private readonly CircuitFactory _circuitFactory;
         private readonly ILogger _logger;
 
+        /// <summary>
+        /// Constructs an instance of <see cref="BlazorHub"/>.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger{BlazorHub}"/>.</param>
+        /// <param name="services">The service provider.</param>
         public BlazorHub(
```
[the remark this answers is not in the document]
SteveSandersonMS: Good point. I've been through and updated all the XML docs to be more specific about "don't call this" (including on the constructor). Also I've `sealed` the class to reduce the scope for any virtual method shenanigans.

src/Microsoft.AspNetCore.Blazor.Server/Builder/ServerSideBlazorApplicationBuilderExtensions.cs
SteveSandersonMS: In this specific place, yes, but an E2E working system also requires a corresponding way to communicate the endpoint to the client (probably as an extra `script` tag arg), probably via some new JavaScript "boot with specific config" API.

src/Microsoft.AspNetCore.Blazor.Server/Builder/ServerSideBlazorApplicationBuilderExtensions.cs
SteveSandersonMS: Good point. Renamed to `UseSignalRWithBlazorHub`.

src/Microsoft.AspNetCore.Blazor.Server/DependencyInjection/ServerSideBlazorServiceCollectionExtensions.cs
SteveSandersonMS: Makes sense. Yes, I had been thinking of the "path" as being the identifier for the endpoint and thus the thing that defines correspondence between the config and the middleware.

Hopefully you're right that when we come to support multiple endpoints, we can consider possibly friendlier APIs. It might be that we use `TStartup` as the identifier, though technically someone might want to mount the same `TStartup` app paths two different paths (no idea why, but just that it's technically possible).

src/Microsoft.AspNetCore.Blazor.Server/DependencyInjection/ServerSideBlazorServiceCollectionExtensions.cs
SteveSandersonMS: OK cool - changed these to `TryAdd...`.

src/Microsoft.AspNetCore.Blazor.Server/Builder/ServerSideBlazorApplicationBuilderExtensions.cs
```diff
@@ -20,94 +19,57 @@ public static class ServerSideBlazorApplicationBuilderExtensions
...
-            Type startupType)
-        {
-            if (builder == null)
-            {
-                throw new ArgumentNullException(nameof(builder));
-            }
+            // WARNING: Don't add extra setup logic here. It's important for
+            // UseServerSideBlazor just to be shorthand for UseSignalR+UseBlazor,
+            // so that people who want to call those two manually instead can
+            // also do so. That's needed for people using Azure SignalR.
```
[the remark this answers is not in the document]
SteveSandersonMS: On the discoverability, I agree. But TBH there's no realistic chance that anyone's setting this up without using either a template or the docs (which I think is also true for many parts of ASP.NET Core and close relations such as EF and Azure SignalR).

If you make a design for new SignalR 3.0 routing config I'd definitely be interested to see it!

PR #1404, opened 2018-09-10
Add  Support for Templated Components

rynowak:
## Summary

This change adds support for templated components
- Components can accept multiple, named child content elements as arguments
- Components can render a child content that accepts a parameter
- We revised the signature of `RenderFragment<T>` to be more usable

**We don't yet have support for genericly typed components - that will come next.**

## How to use

```
@* In TemplatedTable.cshtml *@
@functions {
    [Parameter] RenderFragment Header { get; set; }

    [Parameter] RenderFragment<Item> ItemTemplate { get; set; }

    [Parameter] RenderFragment Footer { get; set; }

    [Parameter]
    IReadOnlyList<Item> Items { get; set; }
}

...

@* Elsewhere... *@
<TemplatedTable Items="@Items">
    <Header><tr>...</tr></Header>
    <ItemTemplate>
        <tr><td>@context.Col1</td><td>@context.Col2</td><td>@context.Col3</td></tr>
    </ItemTemplate>
</TemplatedTable>
```

When a component defines a parameter of type `RenderFragment` or `RenderFragment<T>` (for some `T`), that parameter can have its value bound from a child element (child content). This feature now supports any number of child content parameters and they can have any valid element name as names. Previously only the name `ChildContent` could be used, and the type was always `RenderFragment` of `T`. 

Child content arguments can be can be passed explicitly by wrapping the content in an element that matches the parameter name. Child content arguments must be passed explicitly when passing multiple child content arguments through the component body. The tooling will help with completion for child content  both as attributes and as elements in the body.

We've improved the validation and diagnostics of these features to try and make the limitations and mistakes clear. I don't anticipate that any of these new features are breaking changes. Passing child content to the `ChildContent` parameter still works implicitly, even when the `ChildContent` parameter isn't declared.

----

Child content arguments of type `RenderFragment<T>` passed as elements will have an implict parameter name of `context`. You can change the parameter name if you wish:
```
    <ItemTemplate Context="item">
        <tr><td>@item.Col1</td><td>@item.Col2</td><td>@item.Col3</td></tr>
    </ItemTemplate>
```

The value passed through `Context` must be a static/constant value. You should see completion/intellisense inside the `<ItemTemplate>` element for `item` in this example.

----

Additionally, we've changed the siganture of `RenderFragment<T>` to be more usable. Now a `RenderFragment<T>` will return a `RenderFragment` when evaluated. This makes defining templates using the existing Razor syntax more natural.

```
@{
    RenderFragment t1 = @<div>...</div>;
    RenderFragment<Person> t2 = (person) => @<div>...</div>;
}
```

When using a `RenderFragment<T>` with Razor templates, you need to define the parameter name yourself. 

## Notes
- It's probably best to look at this commit by commit. As part of this work I reorganized and de-duplicated a bunch of the baseline tests.
- There is an issue with child content argument completion when the child content parameter name is the same as an HTML element. The editor will complete it as lowercase.


Liander:
@rynowak Looking at the summary I must first say it looks excellent!

I am however little curious what the implications are of having the same construct for component tags and content parameter tags? Say that you have a component Header and a component MainPage then `<MainPage><Header/></MainPage>` will mean different things if MainPage has a ChildContent parameter or a Header parameter or both. Does the intellisense account for the differences?

SteveSandersonMS:
@Liander How the enclosed markup is interpreted is determined by the component you're referencing:

 * If you're referencing a component that *only* accepts a `ChildContent`, then all the enclosed markup is passed as `ChildContent`. 
 * If you're referencing a component that takes any other combination of `RenderFragment` params, then the top-level elements in the enclosed markup has to match the names of those params, and each nested block is passed to the corresponding param.

So AFAIK there's no case where it is ambiguous.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TagHelperDescriptorExtensions.cs
```diff
@@ -125,5 +126,27 @@ public static string GetEventArgsType(this TagHelperDescriptor tagHelper)
...
+            }
+
+            for (var i = 0; i < tagHelper.BoundAttributes.Count; i++)
+            {
+                var attribute = tagHelper.BoundAttributes[i];
+                if (attribute.IsChildContentProperty())
+                {
+                    yield return attribute;
+                }
+            }
```
SteveSandersonMS: I'm mildly curious about the reason for preferring this kind of construction over a LINQ `.Where(...)` expression. Of course there's nothing wrong with the code as-is here - just interested in your thoughts on the subject.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ScopeStack.cs
```diff
@@ -26,19 +26,39 @@ public void OpenElementScope(string tagName)
...
+            // builder.AddAttribute(0, "{name}", ({type})((context) => (__builder) => { ... }));
+
+            context.CodeWriter.Write($"({type})(");
+
+            if (parameterName != null)
+            {
+                context.CodeWriter.Write($"({parameterName}) => ");
+            }
+
+            scope.LambdaScope = context.CodeWriter.BuildLambda(BuilderVarName);
```
SteveSandersonMS: Previously, the way we avoided emitting unnecessary empty lambdas was only to write them when we observed the first child item within a component.

If I'm following the new implementation correctly, it now works by creating a ComponentChildContent node, but only when there's some nonempty content inside the component. And then it only creates the lambda when writing the ComponentChildContent node, hence no further need to call `IncrementCurrentScopeChildCount`.

Have I followed that correctly? If so, looks good to me!

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
```diff
@@ -115,9 +118,25 @@ public override void VisitTagHelper(TagHelperIntermediateNode node)
...
                 {
-                    _children.Add(node.Children[i]);
+                    var childContent = new ComponentChildContentIntermediateNode();
+                    var childContentAttribute = _component.Component.BoundAttributes
+                        .Where(a => a.Name == BlazorApi.RenderTreeBuilder.ChildContent)
+                        .FirstOrDefault();
+                    if (childContentAttribute != null)
+                    {
+                        // This component accepts child content explicitly.
+                        childContent.Attribute = childContentAttribute;
```
SteveSandersonMS: The use of the term "attribute" here is a bit quirky. Is this a Razorism based on the fact that normally, parameters are represented as attributes? In this case we're dealing with something like a "parameter" or a "child element" in most cases rather than an HTML attribute, right?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
```diff
@@ -178,15 +178,26 @@ public static RazorDiagnostic Create_DisallowedScriptTag(SourceSpan? source)
...
+        public static RazorDiagnostic Create_TemplateInvalidLocation(SourceSpan? source)
         {
-            return RazorDiagnostic.Create(Template_InvalidLocation, source ?? SourceSpan.Undefined);
+            return RazorDiagnostic.Create(TemplateInvalidLocation, source ?? SourceSpan.Undefined);
+        }
+
+        public static readonly RazorDiagnosticDescriptor ChildContentSetByAttributeAndBody =
+            new RazorDiagnosticDescriptor(
+                "BL9995",
+                () => "The child content property '{0}' is set by both the attribute and the element contents.",
```
SteveSandersonMS: This is great!

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
```diff
@@ -199,5 +199,17 @@ public static RazorDiagnostic Create_ChildContentSetByAttributeAndBody(SourceSpa
         {
             return RazorDiagnostic.Create(ChildContentSetByAttributeAndBody, source ?? SourceSpan.Undefined, attribute);
         }
+
+        public static readonly RazorDiagnosticDescriptor ChildContentMixedWithExplicitChildContent =
+            new RazorDiagnosticDescriptor(
+                "BL9996",
+                () => "Unrecognized child content. When using explicit child content, all significant content must be inside a " +
+                "child content element such as '<ChildContent>'.",
```
SteveSandersonMS: Not absolutely essential, but I wonder if we could make this easier to understand. Possible phrasing (if we have the necessary info):

```
Unexpected element 'div' inside component 'SuperTable'. The component 'SuperTable' accepts the following top-level items only: 'Header', 'Body', 'Footer'.
```

Is that possible, and does that match the intent of the diagnostic?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
```diff
@@ -178,15 +178,61 @@ public static RazorDiagnostic Create_DisallowedScriptTag(SourceSpan? source)
...
+            return RazorDiagnostic.Create(ChildContentSetByAttributeAndBody, source ?? SourceSpan.Undefined, attribute);
+        }
+
+        public static readonly RazorDiagnosticDescriptor ChildContentMixedWithExplicitChildContent =
+            new RazorDiagnosticDescriptor(
+                "BL9996",
+                () => "Unrecognized child content. When using explicit child content, all significant content must be inside a " +
+                "child content element such as '<ChildContent>'. Check that the spelling of all child elements matches the component's" +
+                "declared parameters.",
+                RazorDiagnosticSeverity.Error);
```
SteveSandersonMS: Not absolutely essential, but I wonder if we could make this easier to understand. Possible phrasing (if we have the necessary info):

```
Unexpected element 'div' inside component 'SuperTable'. The component 'SuperTable' accepts the following top-level items only: 'Header', 'Body', 'Footer'.
```

Is that possible, and does that match the intent of the diagnostic?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
```diff
@@ -115,12 +124,166 @@ public override void VisitTagHelper(TagHelperIntermediateNode node)
...
+                //
+                // Using implicit child content:
+                // 1. All content is grouped into a single child content lambda, and assiged to the property 'ChildContent'
+                //
+                // Using explicit child content:
+                // 1. All content must be contained within 'child content' elements that are direct children
+                // 2. Whitespace outside of 'child content' elements will be ignored (not an error)
+                // 3. Non-whitespace outside of 'child content' elements will cause an error
+                // 4. All 'child content' elements must match parameters on the component (exception for ChildContent,
+                //    which is always allowed.
```
SteveSandersonMS: Rule 4 here is interesting.

I'm sure it's fine right now. If, in the future, we wanted to relax this so that we offered code completion for known child items, but also allowed unknown ones, how far would that diverge from the current implementation? This would allow us to make child items have the same semantics as normal attribute-style parameters, where you can also pass unknown things that show up like a name-value collection at runtime, and would allow things like:

```
<StatusCodeDisplay Status=currentStatus>
    <Http200>OK</Http200>
    <Http400>BadRequest</Http400>
    ... etc, without having to declare explicit [Parameter] properties for each possible item
</StatusCodeDisplay>
```

Not saying we need this right now. Just interested in whether we could extend to this in the future.

test/Microsoft.AspNetCore.Blazor.Build.Test/ChildContentRazorIntegrationTest.cs
```diff
@@ -181,6 +211,39 @@ public void Render_AttributeChildContent_IgnoresWhitespaceBody()
...
 
+        [Fact]
+        public void Render_MultipleChildContent()
+        {
+            // Arrange
+            AdditionalSyntaxTrees.Add(RenderMultipleChildContent);
+
+            var component = CompileToComponent(@"
+@addTagHelper *, TestAssembly
+@{ RenderFragment<string> header = @<div>@context.ToLowerInvariant()</div>; }
```
SteveSandersonMS: We're changing this syntax later, right? (to `value => @<div>@value.ToLowerInvariant()</div>`)

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
```diff
@@ -1200,7 +1200,7 @@ public void RazorTemplate_InCodeBlock()
             // Act
             var generated = CompileToCSharp(@"
 @{
-    RenderFragment<Person> p = @<div>@context.Name</div>;
+    RenderFragment<Person> p = (person) => @<div>@person.Name</div>;
```
SteveSandersonMS: So cool that we can change the meaning of syntax like this.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
```diff
@@ -212,5 +212,27 @@ public static RazorDiagnostic Create_ChildContentMixedWithExplicitChildContent(S
...
+
+        public static RazorDiagnostic Create_ChildContentHasInvalidAttribute(SourceSpan? source, string attribute, string element)
+        {
+            return RazorDiagnostic.Create(ChildContentHasInvalidAttribute, source ?? SourceSpan.Undefined, attribute, element);
+        }
+
+        public static readonly RazorDiagnosticDescriptor ChildContentHasInvalidParameter =
+            new RazorDiagnosticDescriptor(
+                "BL9998",
+                () => "Invalid parameter name. The parameter name attribute '{0}' on child content element '{1}' can only include literal text.",
```
SteveSandersonMS: Glad you anticipated this sort of thing.

SteveSandersonMS:
Will we have an issue with nested templated components if the developer fails to override the `context` name? Example:

```
<MyComponent>
    <SomeTemplate>
        <AnotherComponent>
            some content
        </AnotherComponent>
    </SomeTemplate>
</MyComponent>
```

Will the compiler complain:

```
CS0136	A local or parameter named 'context' cannot be declared in this scope because that name is used in an enclosing local scope to define a local or parameter
```

Liander:
@SteveSandersonMS I actually meant implications more in terms of user experience. How do I know in the code what it is when I see something like `<MyPage><Header /></MyPage>`? Do I have to look at the implementation of MyPage or do the tooling help me, like when hovering? I guess that could work when you are in doubt, since you usually will know by its context what it should be. To have the same construct I think is a bit odd to be frank, but I will get used to it though 😄

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TagHelperDescriptorExtensions.cs
rynowak: This is more debuggable 😆 

I wouldn't say I have a personal preference for one or the other, if anything I tend to avoid both iterators and linq for explicit loops and lists/arrays.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ScopeStack.cs
rynowak: yes that's right. We need to know while we're building the tree if you have non-trivial content inside a component, so that knowledge is now represented in the tree instead of in the writer.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
rynowak: This is indeed a Razor-ism. The `childContent.Attribute` is a property of type `BoundAttributeDescriptor` defined in Razor. We're using the existing OM for tag helpers to describe components.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
rynowak: Yeah, I agree that would be nicer. I'll see what it would take.

rynowak: I want to omit the offending text in the error message (`div` in your example), because we're going to add one error for the whole lot of content, and printing the text that appeared could be quite large in many cases. 

It's fairly straightforward to include the supported child content parameter names.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
rynowak: As discussed on this mornings call, this would require component child content to operate in a bi-modal fashion. We'd allow component authors to opt-in to a model where child content must be explicit, but arbitrary elements are accepted. 

Anyone who needs a feature like this today can accomplish this using the runtime (make a `Dictionary<string, RenderFragment>` parameter.

test/Microsoft.AspNetCore.Blazor.Build.Test/ChildContentRazorIntegrationTest.cs
rynowak: Yes, this was changed in a later commit.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
rynowak: This is also kinda a cheat. The parameter name goes into an attribute, but it also gets mapped into the generated code in a way that doesn't really have a precedent in Razor. 

For instance if you put in a name that wasn't a valid c# identifier, the experience might not be  great. There's a more robust solution to this that I didn't pursue in this PR.

rynowak:
@Liander  - tooling will help you. child content elements generate tag helpers that only match in the context of the parent tag.

rynowak:
OK, up next.... components with generics!

SteveSandersonMS:
Nice! The new "nested content ambiguity" diagnostic looks great.

RyoukoKonpaku:
@Liander I'd assume a simple ```Ctrl + .``` should show a completion list of components as how @rynowak stated. Though would it probably get tedious on big templates or when components are plenty? As the completion would make it kinda hard to differentiate a ```template``` tag to a ```component``` tag not too sure.

Liander:
@RyoukoKonpaku I have confidence that the parsing and tooling will work even when you have deep nesting of components, elements and template properties but the user must also understand it. 

Am I the only one that thinks it's odd not being able to tell what even a simple statement like `<MyPage><Header><div>Hello World</div></Header></MyPage>` is? Nesting it further will get it even harder.

It could be that MyPage and Header are both taking a ChildContent or it could be that Header is a property of MyPage. I think a design like that will lead people to name properties like MyPageHeader, which I think is a pity. To rely on Ctrl + ‘.’ to tell is too tediously, hover information of type might get you little closer, and coloring the content property tags different from component tags will almost get you there… but it is still odd in my mind to rely on tooling. 

Again, my personal preference would be to say `<MyPage><MyPage.Header><div>Hello World</div></MyPage.Header></MyPage>`, so you are explicit that it is really an extension of the specification of the MyPage component tag. But I seem to be quite alone in this thinking… and I seem to be alone to think it’s odd to overload the same construct for two completely semantically different things... Am I missing something?

RyoukoKonpaku:
👍 @Liander I agree with your thoughts overall since I'm coming from a ```XAML``` background which uses that format extensively. One advantage intellisense wise as well with that is that by typing the parent tag ```MyPage<dot>``` should accurately show only the templates that are available for that component.


But I do recall the standard tags format is somewhat more familiar to those who does SPA frameworks like Angular, although it does feel like "magic" as at first glance you'd have no idea on it's relation to the parent tag. Though this is my bias kinda showing as well as I do like that format as I'm familiar with it.

Liander:
@RyoukoKonpaku So it is because of an allergy to XAML...
Maybe there is another construct one can use? Will it be less XAML with a prefix, like just a dot `<.Header />` or does is it become an illegal tag then?

RyoukoKonpaku:
@Liander maybe an opt-in or optional syntax I wonder? This is more on comments for @rynowak since it's a design decision and might have a better view on the decision overall than us.

Liander:
@rynowak @SteveSandersonMS A somewhat related aspect is the definition and the meaning of the Parameter attribute. Now it looks like it is given by the types (`RenderFraction` and `RenderFraction<T>`) that it can be set as both attributes in tags and as child content. Is that correct? If this is true, it's this “certain types” I feel little unease about and that the behavior is not fully determined by the Parameter attribute itself. Maybe consider to generalize it so all properties can be set either as an attribute or as child content? So the following would all mean the same thing:
```
<MyPage Title=”Hello Blazor!” Header=@(@<div>Hello World!</div>)" />
```
```
<MyPage Title=”Hello Blazor!”>
    <MyPage.Header><div>Hello World!</div></MyPage.Header>
</MyPage>
```
```
<MyPage>
  <MyPage.Title>”Hello Blazor!”</MyPage.Title>
  <MyPage.Header><div>Hello World!</div></MyPage.Header>
</MyPage>
```
This can open up for further constructs further down the road so that a property-tag is not only for producing an action or a function but could be made more general and also take a factory. As an example, if you have a property of type IDictionary<int, RenderFraction> those could also be set as child contents, similar to:
```
<StatusCodeDisplay>
   <StatusCodeDisplay.Switches Factory=”new Dictionary<int, RenderFraction>”>
      <SetKeyValue instance=”@Switches” key=”404”><Http404/></SetKeyValue>
   </StatusCodeDisplay.Switches>
 </StatusCodeDisplay>
```
That particular example might be bad since you could also make use of a conditional construct:
```
<StatusCodeDisplay>
   <StatusCodeDisplay.Switches Context=”statusCode” >
      <Case on=”@(statusCode==404)”><Http404/></Case>
   </StatusCodeDisplay.Switches>
 </StatusCodeDisplay>
```
But I think you get my point, which is making a generalization of the Parameter concept to be more future proof. That thinking can also help in deciding what construct you want to have for those property-tags.

It would be interesting to hear your thoughts on this.

conficient:
Can we get the compiler issue a warning if you use a HTML tag name as a parameter for a template?

In @rynowak 's first example he used
```
[Parameter] RenderFragment Header { get; set; }
[Parameter] RenderFragment Footer { get; set; }
``` 
I tried this and Header/Footer didn't show up in Intellisense, presumably because these conflict with the `<header>` and `<footer>` HTML tags.

PR #1405, opened 2018-09-10
Small tweaks to support publish

shibutamang:


SteveSandersonMS:
Thanks @shibutamang for contributing this.

So that we can review it, could you please comment to describe what functionality this is intended to add, or what specific problem (ideally with repro steps) it solves?

PR #1407, opened 2018-09-11
Update linking

SteveSandersonMS:
Updates to newer linker binary, plus config changes recommended by @marek-safar.


csnewman:
Is it worth including the "System.Linq.Expressions*" entry from the docs? https://blazor.net/docs/host-and-deploy/configure-linker.html to allow JSON.NET etc to work by default.

marek-safar:
The "System.Linq.Expressions* changes are the temporary setting it should not be needed soon

SteveSandersonMS:
I think we'd be looking for a more targetted config change than forcing `System.Expressions*` to be retained unconditionally. For example,

* AFAIK, it should be possible for Json.NET to embed its own linker config as an embedded `.xml` resource (like Mono's `mscorlib.dll` does), so the config would apply only when people are using Json.NET
* Or if that's not viable for some reason, perhaps we can instead identify which specific method(s) it's consuming only through reflection and whitelist those only. Then we may be able to get Json.NET to work without meaningfully changing the default post-linker BCL size.

So let's keep that separate from this PR.

SteveSandersonMS:
Only about 40KB, so the size saving isn’t really meaningful, but this does have the benefit of removing some hard coded items from the linker whitelist, and eliminates 3 HTTP requests from app startup (they were type-forwarder-only assemblies).

PR #1417, opened 2018-09-14
Add support for generic-typed components

rynowak:
## What is it?

This change adds the ability to define and consume generic-typed components from both C# and Razor.

**GenericBlink.cshtml**
```
@typeparam TItem

<blink>
  @ChildContent(Item)
</blink>

@functions {
    [Parameter] RenderFragment<TItem> ChildContent { get; set; }
    [Parameter] TItem Item { get; set;}

}
```

```
<GenericBlink TItem=string Item="@Message">
  <div>Your message was: @context.ToLowerInvariant()</div>
</GenericBlink>

@functions {
    [Parameter] string Message { get; set; }
}
```

Currently when consuming a component from Razor, you must specific the type arguments. Support for type inference of usage of generic-typed components is coming next.

You can define type parameters on a component with the new `@typeparam` directive. There is currently no support for type constraints when defining generic-typed components from Razor.

If you are defining a component from C# then everything should 'just work' the way you'd expect.

Most of the work and tricky cases are related to making it possible to consume generic-typed components in generated code.

## How does it work

When we discover a generic-typed component, we synthesize an extra 'type-argument' attribute for each type-parameter. This gives us intellisense and completion support for specifying types when using the component. The type argument must be a nominal type (like the inside of a `typeof(...)` expression.. ie, it can't be a variable.

Once we've bound the component tree and processed cross-cutting features like event handlers and `bind` we find each generic-typed component and process the type parameters. We rewrite the type name of the component and the type name of each parameter that's set (for type-checking) and do code generation with the type-arguments substituted in.

This approach obviously won't be used for type-inference, but it's an important feature to provide for cases where you want to (or need to) be specific. 

## Misc

When setting a generic-typed parameter, you won't get support for the shorthand the Razor supports for `string` attributes.

Ex: `<Generic TValue=string Value="Foo" />`
In this example if `Value` is of type `TValue` then `Foo` will be interpreted as C# not as the contents of a string. 

--

When using generics, there's always the possibility that the error messages could be confusing or misleading when the generic types can't be figured out. I've tried my best to minimize the error cases that you'll get in the editor, so that you can clearly see the error that Blazor reports about missing generic types. If you hit cases where the error messages are misleading, please let us know.


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -111,38 +126,103 @@ private TagHelperDescriptor CreateDescriptor(BlazorSymbols symbols, INamedTypeSy
...
+                if (type is ITypeParameterSymbol)
+                {
+                    return true;
+                }
 
-                    if (property.kind == PropertyKind.Delegate)
+                // We need to check for cases like:
+                // [Parameter] List<T> MyProperty { get; set; }
+                // AND
+                // [Parameter] List<string> MyProperty { get; set; }
```
SteveSandersonMS: Not necessarily a problem, but I didn't spot any unit test coverage for the pre-closed generic types (like your `List<string>` example here) or a nested generic type. Though maybe I'm missing the subtleties of what's different about those cases. Do you feel L191-L207 is adequately covered?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
```diff
@@ -0,0 +1,169 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.AspNetCore.Blazor.Razor
+{
+    // This pass:
+    // 1. Adds diagnostics for missing generic type arguments
+    // 2. Rewrites the type name of the component to substitute generic type arguments
+    // 3. Rewrites the type names of parameters/child content to substitute generic type arguments
+    internal class GenericComponentPass : IntermediateNodePassBase, IRazorOptimizationPass
```
SteveSandersonMS: It's very cool that what this file does is possible. Really helps to have this level of control over customising the compiler!

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
```diff
@@ -0,0 +1,169 @@
...
+            public override SyntaxNode Visit(SyntaxNode node)
+            {
+                // We can handle a single IdentifierNameSyntax at the top level (like 'TItem)
+                // OR a GenericNameSyntax recursively (like `List<T>`)
+                if (node is IdentifierNameSyntax identifier)
+                {
+                    if (_bindings.TryGetValue(identifier.Identifier.Text, out var binding))
+                    {
+                        // If we don't have a valid replacement, use object. This will make the code at least reasonable
+                        // compared to leaving the type parameter in place.
```
SteveSandersonMS: I think that's a good plan. It means in a lot of cases there *won't* be any other errors, and people will just not get any intellisense prompts on their `context` variable (besides `ToString()`, `GetHashCode()`, etc.), which is ideal.

Technically there's the case that the component declares `where T:struct`, triggering a different error, but that would be very rare and I don't think worth digging further into.

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
```diff
@@ -81,6 +81,36 @@ public class MyComponent : BlazorComponent
             CompileToAssembly(generated);
         }
```
SteveSandersonMS: In this file, I notice that in commit d9c69366932a, the line:

    GenerateBaselines = true;

was removed, but then it was put back there in commit f288b05ab6.

Was this deliberate? Shouldn't that flag be `false` normally, so that tests fail if the baselines don't match?

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
```diff
@@ -1228,6 +1229,146 @@ public void EventHandler_OnElement_WithDelegate()
...
+    {
+        [Parameter] TItem Item { get; set; }
+    }
+}
+"));
+
+            // Act
+            var generated = CompileToCSharp(@"
+@addTagHelper *, TestAssembly
+<MyComponent TItem=string Item=""@(""hi"")""/>");
```
SteveSandersonMS: Good to see the syntax is `TItem=string` rather than `TItem=typeof(string)`. I'll be interested to see what level of IDE support we have for this - does it autocomplete on type names?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
rynowak: yeah, I was optimizing here to try and make the error that we create be the only one the user sees. You'd also have issues if the generic type was intended to be a delegate type, since lambdas aren't implicitly assignable to object, and MethodGroup to delegate conversion isn't valid in that context either.

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
rynowak: yes, and it is false by default. I've debated added a test here that always fails if it's 'true' so that it can't get checked in that way. Maybe I'll go ahead and add it now.

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
rynowak: yes, we put the cursor inside a `typeof( )` so you get the completion you'd get in that context. I think this is the correct thing to do because you'll get immediate feedback if you try to use a variable instead of a type constant.

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
SteveSandersonMS: What I mean is, it’s explicitly set to true in the checked-in code. Is that deliberate? Sounds like not.

DamianEdwards:
@rynowak any particular reason the type args don't use quote delimiters when being specified at component invocation time in the CSHTML file? Looks odd compared to the rest of the attributes of the element.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentTagHelperDescriptorProvider.cs
rynowak: I'll add a test for make it explicit. Pre-closed generics are already including in our built-in components. This code actually took some doing to get right beause their lingo is a little confusing.

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
rynowak: Added a test so this will fail if you do it. I also removed the ability to set this PER-test so that our verification will be resiliant. So now it's still easy to iterate on the baselines locally - (ignore the failing tests), but you need to undo it before pushing.

rynowak:
> type args don't use quote delimiters when being specified at component invocation time in the CSHTML file? Looks odd compared to the rest of the attributes of the element.

Someone (dunno who) decided that tag helper C# attributes allow quotes to be optional....

PR #1440, opened 2018-09-18
Update Startup.cs

IEvangelist:
Use `nameof` instead of string literal.

### DIFF
--- a/samples/StandaloneApp/Startup.cs
+++ b/samples/StandaloneApp/Startup.cs
@@ -14,7 +14,7 @@ namespace StandaloneApp
 
         public void Configure(IBlazorApplicationBuilder app)
         {
-            app.AddComponent<App>("app");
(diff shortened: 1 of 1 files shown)


adrian109:
If you're making this change then the other samples should be updated to match, e.g. ServerSideBlazor.Client

Suchiman:
@IEvangelist "app" refers to the `<app>` inside https://github.com/aspnet/Blazor/blob/dd437f091356f3d19ed6c1314a98a0970ec5ebbc/samples/StandaloneApp/wwwroot/index.html#L12
Changing this to nameof would indicate a relation to the `IBlazorApplicationBuilder` which is just coincidentally named `app` as well

IEvangelist:
That makes more sense, thanks for the quick feedback everyone.

PR #1453, opened 2018-09-20
Add Type Inference for Generic-Typed Components

rynowak:
This change allows you to use generic-typed components without
explicitly specify type arguments.

**Before:**
```
<Grid Items="@MyItems" TItem="Person">
...
</Grid>
```

**After:**
```
<Grid Items="@MyItems">
...
</Grid>
```

Where possible, the type of the component will be inferred based on the
values passed to component parameters. This won't work like magic, you
have to specify parameters that provide type arguments for all of the
component's type parameters.

This is a best effort system, and it's largely based on the limitations
and behaviours of generic type inference in C#. We think it will work
well with most Blazor features and for most reasonable components. You
should not forget it's easy to specify type arguments, because you may
still have to in some cases.

In particular you may notice issues if you are trying to use a generic
typed component where all of the parameters are delegates or templates.
Type inference for delegates/lambdas in C# is based on the context. Any
time you combine generics and delegates it's easy to get into a scenario
where the compiler won't infer the correct type, or will give up.

----
The type inference feature works by generating a 'thunk' method in
*caller* that can act as a site for type inference (C# does not support
inference on constructors).

For our grid example above, the non-inferenced code will look something
like:
```
builder.OpenComponent<Grid<Person>>(0);
builder.AddAttribute(1, "Items", MyItems);
builder.CloseComponent();
```

Note that we can only write the type `Grid<Person>` because you wrote
`Person` in your code. What you type in the `TItem` attribute value gets
inserted into the generated code such that it fills in the generic type
parameter.

On the other hand, if you want is to infer the type, we have to do some
compiler magic. That looks like:
```
__Blazor.(long generated name).TypeInference.CreateGrid_0(builder, 0, 1, MyItems);
...

// elsewhere in the file
internal static class TypeInference
{
    public static void CreateGrid_0<TItem>(RenderTreeBuilder builder, int seq, int __seq0, global::System.Collections.Generic.List<TItem> __arg0)
    {
	builder.OpenComponent<Grid<TItem>>(seq);
	builder.AddAttribute(__seq0, "Items", __arg0);
        builder.CloseComponent();
    }
}
```

This allows us to rely on the C# compiler for itype inference.


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -343,40 +343,111 @@ public override void WriteComponent(CodeRenderingContext context, ComponentExten
                 throw new ArgumentNullException(nameof(node));
             }
 
-            foreach (var typeArgument in node.TypeArguments)
+            if (node.TypeInferenceNode == null)
```
rynowak: Most of the work here is in the code generator. Basically creating a component has two code paths. I tried to make as much of it reused as is feasible.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -17,7 +17,7 @@ internal class BlazorDesignTimeNodeWriter : BlazorNodeWriter
     {
         private readonly ScopeStack _scopeStack = new ScopeStack();
 
-        private readonly static string DesignTimeVariable = "__o";
+        private static readonly string DesignTimeVariable = "__o";
```
rynowak: I think VS did this. They have a new feature that does 'cleanup' which I have turned on. So if you notice other stylistic changes like this, it's just VS doing it's thing.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -564,37 +657,69 @@ public override void WriteTemplate(CodeRenderingContext context, TemplateInterme
...
-                : BlazorApi.ElementRef.FullTypeName;
-            WriteCSharpCode(context, new CSharpCodeIntermediateNode
+            WriteReferenceCaptureInnards(context, node, shouldTypeCheck: true);
+            context.CodeWriter.WriteLine(";");
+        }
+
+        protected override void WriteReferenceCaptureInnards(CodeRenderingContext context, RefExtensionNode node, bool shouldTypeCheck)
+        {
+            // We specialize this code based on whether or not we can type check. When we're calling into
+            // a type-inferenced component, we can't do the type check. See the comments in WriteTypeInferenceMethod.
```
rynowak: @SteveSandersonMS 

So it turns out most features in Blazor are impacted by generics 😆 

Ref was the hardest to deal with. Basically we currently emit an `Action<object>` that contains a cast and assigns the field. The problem is that the calling code can't write the type name (because generics) and the thunk can't write the field name because it's not a member. 

The fix I went with was to split them up into 2 lambdas. LMK if you think there are problems with this.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -564,37 +657,69 @@ public override void WriteTemplate(CodeRenderingContext context, TemplateInterme
...
+            if (shouldTypeCheck)
+            {
+                // The runtime node writer moves the call elsewhere. At design time we
```
rynowak: This is another case where what we're doing at design time is specialized and can't work with generics. I preserved the exising design time behaviour when you don't need inference.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/DesignTimeCodeGenerationTest/ChildComponent_GenericBind/TestComponent.codegen.cs
```diff
@@ -33,14 +33,15 @@ protected override void BuildRenderTree(Microsoft.AspNetCore.Blazor.RenderTree.R
...
+            __o = Microsoft.AspNetCore.Blazor.Components.RuntimeHelpers.TypeCheck<string>(Microsoft.AspNetCore.Blazor.Components.BindMethods.GetValue(
 #line 2 "x:\dir\subdir\Test\TestComponent.cshtml"
                                     Value
 
 #line default
 #line hidden
-            );
-            __o = Microsoft.AspNetCore.Blazor.Components.BindMethods.SetValueHandler(__value => Value = __value, Value);
+            ));
+            __o = new System.Action<string>(
```
rynowak: This doesn't seem right.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/DesignTimeCodeGenerationTest/ChildComponent_GenericBind/TestComponent.codegen.cs
```diff
@@ -33,14 +33,15 @@ protected override void BuildRenderTree(Microsoft.AspNetCore.Blazor.RenderTree.R
 #line default
 #line hidden
             );
-            __o = Microsoft.AspNetCore.Blazor.Components.BindMethods.GetValue(
+            __o = Microsoft.AspNetCore.Blazor.Components.RuntimeHelpers.TypeCheck<string>(Microsoft.AspNetCore.Blazor.Components.BindMethods.GetValue(
```
rynowak: This is an improvvement, we weren't type checking this before.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/DesignTimeCodeGenerationTest/ChildComponent_GenericBind/TestComponent.codegen.cs
rynowak: Ah yes. This was a test bug. This test was testing the case where you were using bind on a parameter that isn't defined. I fixed this test, added another to cover that scenario. So these diffs are expected.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDesignTimeNodeWriter.cs
```diff
@@ -343,40 +343,111 @@ public override void WriteComponent(CodeRenderingContext context, ComponentExten
...
-                // design -time we want to render an empty lambda to provide proper scoping
-                // for any code that the user types.
-                context.RenderNode(new ComponentChildContentIntermediateNode()
+                var attributes = node.Attributes.ToList();
+                var childContents = node.ChildContents.ToList();
+                var captures = node.Captures.ToList();
+                var remaining = attributes.Count + childContents.Count + captures.Count;
+
+                // Looks like:
+                // __Blazor.SomeNamespace.TypeInference.CreateGrid_0(builder, 1, "Foo");
```
SteveSandersonMS: Minor point: looks like the value is `-1` not `1`.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
```diff
@@ -49,5 +53,169 @@ public sealed override void WriteCSharpCodeAttributeValue(CodeRenderingContext c
...
+            // attributes and child contents look like variables. 
+            //
+            // Looks like:
+            //
+            //  public static void CreateFoo_0<T1, T2>(RenderTreeBuilder builder, int seq, int __seq0, T1 __arg0, int __seq1, global::System.Collections.Generic.List<T2> __arg1, int __seq2, string __arg2)
+            //  {
+            //      builder.OpenComponent<Foo<T1, T2>>();
+            //      builder.AddAttribute(__seq0, "Attr0", __arg0);
+            //      builder.AddAttribute(__seq1, "Attr1", __arg1);
+            //      builder.AddAttribute(__seq2, "Attr2", __arg2);
```
SteveSandersonMS: Ah yes. I think last time we spoke, I suggested we could do away with sequence numbers for attribute nodes, since the diffing is meant to be order-independent anyway. But on futher inspection, I now remember we have two different merge strategies for attributes: a fast path that does use sequence numbers to avoid needing to remember what it saw earlier, and a slow path that's independent of sequence numbers.

So I agree it's better to preserve the sequence numbers like you've done here.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
```diff
@@ -49,5 +53,169 @@ public sealed override void WriteCSharpCodeAttributeValue(CodeRenderingContext c
...
+            //  }
+            //
+            // As a special case, we need to generate a thunk for captures in this block instead of using
+            // them verbatim.
+            //
+            // The problem is that RenderTreeBuilder wants an Action<object>. The caller can't write the type
+            // name if it contains generics, and we can't write the variable they want to assign to. 
+            var writer = context.CodeWriter;
+
+            writer.Write("public static void ");
```
SteveSandersonMS: Is this a method-within-a-method? If so, really handy that the compiler now allows this!

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
```diff
@@ -49,5 +53,169 @@ public sealed override void WriteCSharpCodeAttributeValue(CodeRenderingContext c
             context.Diagnostics.Add(BlazorDiagnosticFactory.Create_CodeBlockInAttribute(node.Source, content));
             return;
         }
+
+
+        // Currently the same for design time and runtime
+        public void WriteComponentTypeInferenceMethod(CodeRenderingContext context, ComponentTypeInferenceMethodIntermediateNode node)
```
SteveSandersonMS: I like how this method is implemented with a combination of recent C# features yet also is structured for a really high level of clarity (e.g., repeating certain blocks instead of adding further abstractions in the name of DRY).

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
```diff
@@ -49,5 +53,169 @@ public sealed override void WriteCSharpCodeAttributeValue(CodeRenderingContext c
...
+            {
+                writer.Write(", ");
+            }
+
+            for (var i = 0; i < parameters.Count; i++)
+            {
+                writer.Write("int ");
+                writer.Write(parameters[i].seqName);
+
+                if (parameters[i].typeName != null && parameters[i].parameterName != null)
```
SteveSandersonMS: Could you explain these conditions a bit more? It looks to me like `parameterName` is certainly always nonnull due to how `GetParameters` is implemented, and I don't know what scenario would be represented by `typeName` being null.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
```diff
@@ -45,20 +47,112 @@ public void VisitExtension(ComponentExtensionNode node)
...
                 foreach (var attribute in node.Component.GetTypeParameters())
                 {
                     bindings.Add(attribute.Name, new GenericTypeNameRewriter.Binding() { Attribute = attribute, });
                 }
 
+                // Listing all type arguments that have been specified.
+                var hasTypeArgumentSpecified = false;
                 foreach (var typeArgumentNode in node.TypeArguments)
                 {
+                    hasTypeArgumentSpecified |= true;
```
SteveSandersonMS: That's odd. Why not `hasTypeArgumentSpecified = true;`?

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
```diff
@@ -99,19 +219,86 @@ private void Process(ComponentExtensionNode node)
...
+                var @namespace = documentNode.FindPrimaryNamespace().Content;
+                @namespace = string.IsNullOrEmpty(@namespace) ? "__Blazor" : "__Blazor." + @namespace;
+                @namespace += "." + documentNode.FindPrimaryClass().ClassName;
+
+                var typeInferenceNode = new ComponentTypeInferenceMethodIntermediateNode()
+                {
+                    Bindings = bindings,
+                    Component = node,
+
+                    // Method name is generated and guaraneteed not to collide, since its unique for each
```
SteveSandersonMS: Typo: guaraneteed (and strictly speaking, its=>"it's" here, but I know that's kind of an annoying thing to mention)

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
```diff
@@ -99,19 +219,86 @@ private void Process(ComponentExtensionNode node)
...
+                    // component call site.
```
SteveSandersonMS: Did you decide not to deduplicate the methods based on the set of args? Fine with me if so - it's an optimisation we could add later if any important use case emerges.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/RuntimeCodeGenerationTest/ChildComponent_GenericWeaklyTypedAttribute_TypeInference/TestComponent.codegen.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+        #pragma warning restore 1998
+    }
+}
+namespace __Blazor.Test.TestComponent
+{
+    #line hidden
+    internal static class TypeInference
+    {
+        public static void CreateMyComponent_0<TItem>(global::Microsoft.AspNetCore.Blazor.RenderTree.RenderTreeBuilder builder, int seq, int __seq0, TItem __arg0, int __seq1, System.Object __arg1)
```
SteveSandersonMS: Is it possible that this "treat weakly-typed args as `System.Object`" breaks our use of different overloads of `AddAttribute` for different arg types?

Currently we have overloads of `AddAttribute` for `bool`, `string`, `Action`, `Action<UIEventArgs>`, `Func<Task>`, `Func<UIEventArgs, Task>`, `MulticastDelegate`, `object`, and `RenderTreeFrame`. Each has slightly different logic for converting the arg, skipping if it's a false boolean, etc.

SteveSandersonMS: Actually on reading the `AddAttribute` codes more closely, it's not necessarily problematic, since the considerable majority of the special cases are skipped in the case of components (rather than elements).

However structuring the code like this, where each `AddAttribute` overload separately makes component-vs-element decisions seems like it's inviting misunderstanding/bugs. I wonder if we should add a new `AddComponentParameter` to `RenderTreeBuilder`, whose logic is much simpler, and then we can call that from here and anywhere else that we know we're in the context of a component. Then we can change the original `AddAttribute` just to throw if it gets used for a component (maybe even rename all its overloads to `AddElementAttribute`).

If you like that and want to do it, that's great. If you like that and don't want to do it now, fine too, maybe we could have a backlog item to remember it. Or if you just don't like it, let me know!

SteveSandersonMS: It would also be good to split into `AddComponentParameter`/`AddElementAttribute` just so we're not doing so many `if` tests and branching at runtime, given we could have resolved all that at compile-time.

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
```diff
@@ -1271,10 +1381,75 @@ namespace Test
...
-        Action<TItem> ValueChanged { get; set; }
+        Action<TItem> ItemChanged { get; set; }
+    }
+}
+"));
+
+            // Act
+            var generated = CompileToCSharp(@"
+@addTagHelper *, TestAssembly
+<MyComponent bind-Item=Value/>
```
SteveSandersonMS: It's superb that this sort of thing works so naturally.

test/testapps/BasicTestApp/RazorTemplates.cshtml
```diff
@@ -6,5 +6,5 @@
     @{
         RenderFragment<OrderedList<string>.Context> template = (context) => @<li>#@context.Index - @context.Item.ToLower()</li>;
     }
-    <OrderedList Items="@items" Template="@template" TItem="string"/>
+    <OrderedList Items="@items" Template="@template" />
```
SteveSandersonMS: 👍

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
rynowak: Yeah, it seemed worth it to me regardless of whether it had an actual runtime impact. This way the behaviour of tests that render a component is the same with and without type inference.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
rynowak: Well you can use local functions, but this is not. Local functions would be fraught with peril for our use case.

This is a separate namespace/class/method from your original code. The main reason why its a separate type is so that we can avoid renaming the generic type parameters. If the thunk is part of the same class, then it shares the type parameters of the calling component, which is very counterproductive.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
rynowak: https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
rynowak: yeah, sorry this was cleanup I missed when I was iterating. I originally started with rendering the 'ref' lambdas inside the generated thunk, but it doesn't really work.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorNodeWriter.cs
```diff
@@ -49,5 +53,169 @@ public sealed override void WriteCSharpCodeAttributeValue(CodeRenderingContext c
...
+            }
+
+            foreach (var capture in node.Component.Captures)
+            {
+                context.CodeWriter.WriteStartInstanceMethodInvocation("builder", capture.IsComponentCapture ? BlazorApi.RenderTreeBuilder.AddComponentReferenceCapture : BlazorApi.RenderTreeBuilder.AddElementReferenceCapture);
+                context.CodeWriter.Write(parameters[index].seqName);
+                context.CodeWriter.Write(", ");
+
+                var cast = capture.IsComponentCapture ? $"({capture.ComponentCaptureTypeName})" : string.Empty;
+                context.CodeWriter.Write($"(__value) => {{ {parameters[index].parameterName}({cast}__value); }}");
```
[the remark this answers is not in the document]
rynowak: It's the latter. We need to pass an `Action<object>` into the builder, but we happen to have an `Action<SomethingSpecific>`. We have an `Action<SomethingSpecific>` because we can't put the cast inside the lambda that gets passed to the thunk.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorRuntimeNodeWriter.cs
```diff
@@ -355,6 +427,14 @@ public override void WriteComponentAttribute(CodeRenderingContext context, Compo
...
             context.CodeWriter.Write(", ");
 
+            WriteComponentAttributeInnards(context, node, canTypeCheck: true);
+
+            context.CodeWriter.Write(");");
+            context.CodeWriter.WriteLine();
+        }
+
+        private void WriteComponentAttributeInnards(CodeRenderingContext context, ComponentAttributeExtensionNode node, bool canTypeCheck)
+        {
```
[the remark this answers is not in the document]
rynowak: Sure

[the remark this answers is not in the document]
rynowak: I'll go ahead and do this for the component-related stuff. These are the complicated ones anyway.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
rynowak: eh, this just seemed naturl to me. If you think it's clearer the other way I don't mind.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/GenericComponentPass.cs
rynowak: Yep. I decided to keep it simple.

test/Microsoft.AspNetCore.Blazor.Build.Test/TestFiles/RuntimeCodeGenerationTest/ChildComponent_GenericWeaklyTypedAttribute_TypeInference/TestComponent.codegen.cs
rynowak: So weak typing can absolutely break type inference/or be broken when using type inference. This is a limitation of how the compiler works.  For delegates/bind/eventhandlers I'm actually using `Action<object>` which resolves a tricky case. 

I'm going to log a backlog item to consider changing the runtime in the future. I'm not sure how it helps this feature without know that specific things are broken..

test/Microsoft.AspNetCore.Blazor.Build.Test/GenericComponentRazorIntegrationTest.cs
```diff
@@ -112,6 +181,105 @@ public void Render_GenericComponent_WithChildContent()
                 frame => AssertFrame.Whitespace(frame, 6));
         }
 
+        [Fact]
+        public void Render_GenericComponent_TypeInference_WithRef()
```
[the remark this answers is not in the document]
rynowak: Render_GenericComponent_TypeInference_WithRef
Render_GenericComponent_TypeInference_WithRef_Recursive
?

PR #1470, opened 2018-09-23
Adds support for 'Context' parameters on components

rynowak:
Adds support for 'Context' parameters on components

This change allows you to set the parameter name of a parameterized child content by using the `Context` attribute on the component. The `Context` attribute will be defined (and shown by completion) when the component has one or more declared parameterized (`RenderFragment<>`) child content parameters.

This is nice for cases where you are using implicit child content:

```
<ol>
  <Repeater Items="People" Context="person">
    <li>@person.FirstName</li>
  </Repeater>
</ol>
```

 or, when you have multiple child content elements and want them all to have the same parameter name:

 ```
<MyComponent Items="People" Context="person">
    <ChildContent1><div>@person.FirstName</div></ChildContent1>
    <ChildContent2><div>@person.LastName</div></ChildContent2>
</Repeater>
```

The parameter name can be overridden by using the `Context` parameter on the child content element:

 ```
<MyComponent Items="People" Context="person">
    <ChildContent1 Context="item"><div>@item.FirstName</div></ChildContent1>
    <ChildContent2><div>@person.LastName</div></ChildContent2>
</Repeater>
```

If the component defines a `Context` parameter already then we won't synthesize one - your component's parameter will work exactly as it did before this feature.


src/Microsoft.AspNetCore.Blazor.Razor.Extensions/BlazorDiagnosticFactory.cs
```diff
@@ -308,6 +308,16 @@ public static RazorDiagnostic Create_ChildContentHasInvalidParameter(SourceSpan?
             var attributesText = string.Join(", ", attributes.Select(a => $"'{a.Name}'"));
             return RazorDiagnostic.Create(GenericComponentTypeInferenceUnderspecified, source ?? SourceSpan.Undefined, component.TagName, attributesText);
         }
-    }
 
+        public static readonly RazorDiagnosticDescriptor ChildContentHasInvalidParameterOnComponent =
+            new RazorDiagnosticDescriptor(
```
rynowak: Added a new error for component-scoped invalid parameters. The existing error message would be confusing as it mentions child content elements specifically.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
```diff
@@ -234,7 +241,7 @@ private ComponentChildContentIntermediateNode RewriteChildContent(BoundAttribute
                     }
                     else if (child is TagHelperPropertyIntermediateNode property)
                     {
-                        if (property.BoundAttribute.Kind == BlazorMetadata.ChildContent.TagHelperKind)
+                        if (property.BoundAttribute.IsChildContentParameterNameProperty())
```
rynowak: This check is stricter than it was before, which is nice. If we end up adding anything additional to a 'child content' element, then this code won't have to change too much.

rynowak: Previously we assumed that any 'component' parameter on a child content element was 'Context'.  Now we're matching on metadata.

Liander:
@rynowak Would it make sense to also include an attribute for an optional index variable name? The caller can always wrap extra information in an anonymous object but it can be good as a way to let it be controlled by the one providing the template (and the content assignment tag). The passed (invocation) index could be used for showing alternating background colors on rows for example. Like the extra Linq-Select-index. If it is present the lambda expression would loosely be resolved to `(x) => (x, i++) => Action<Builder> { f(x, i) }`. (Setting lambda variables at parent level is something I need time to get used to but setting an optional index variable I think would only make sense on content parameter tags.)

Just a thought.

SteveSandersonMS:
> What I implemented in this PR applies to a single component and it's child-content(s). We can't really recursively apply context to all arbitrary nested components due to scoping.

I know - that's what I understood! In my comment about *eventually remove all the Form=context*, the "eventually" referred to when we implement the part of #1 that's similar to React's "context" notion. This is *not* the same as the context we use for templated components, so we'll certainly need to use a different name. The notion I'm referring to might be better called something like `treeData` - it's a runtime feature that lets components populate a key-value store that is then readable by their descendants.

> Would it make sense to also include an attribute for an optional index variable name?

It's a good thought, and indeed one we've considered. It's not just index variables though - there are cases for passing any number of items with any set of names. To ensure it's discoverable by the component consumer, we decided the most easy-to-understand thing would be to force the component author to pass exactly one thing so it can have a predefined name. If you want to pass multiple things, you'd need to define a structure for that, e.g.,

```
public readonly struct MyListContext<T>
{
    public readonly T Item;
    public readonly int Index;
    public readonly int NumItems;
}
```

... then the consumer can type `@context.`, and intellisense will show them the names of all the context values. It's much easier for the consumer than having to somehow know how many args their lambda should take and what names make sense for each of those args.

Liander:
@SteveSandersonMS Yes, this would be a special case, similar to what Linq-Select has. I just want to clarify that the idea is that it will be independent to the component producer who would still pass in the same object. So when consumer is giving Context=x it would generate `(x) => Action<Builder> { f(x) }`, and when giving both Context=x, Index=i it would generate `int i; (x) => (x, i++) => Action<Builder> { f(x, i) };`. So it will be an automatic invocation counter. I just wanted to clarify if that was not understood.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/Resources.resx
```diff
@@ -135,6 +135,9 @@
   <data name="ChildContentParameterName_Documentation" xml:space="preserve">
     <value>Specifies the parameter name for the '{0}' lambda expression.</value>
   </data>
+  <data name="ChildContentParameterName_TopLevelDocumentation" xml:space="preserve">
+    <value>Specifies the parameter name for all child content lambda expressions.</value>
```
SteveSandersonMS: Calling them "lambda expressions" here is interesting - it implies the user has some understanding of the shape of the generated code. TBH I'm fine with it, but I wondered what you thought about calling them "child content blocks" or something instead (also on L136). Please don't feel pressured to change this unless you really want!

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/TagHelperBoundAttributeDescriptorExtensions.cs
```diff
@@ -94,5 +113,44 @@ public static bool IsParameterizedChildContentProperty(this BoundAttributeDescri
...
+                throw new ArgumentNullException(nameof(attribute));
+            }
+
+            return attribute.IsChildContentProperty() &&
+                !string.Equals(attribute.TypeName, BlazorApi.RenderFragment.FullTypeName, StringComparison.Ordinal);
+        }
+
+        /// <summary>
+        /// Gets a value that indicates whether the property is used to specify the name of the parameter
+        /// for a p arameterized child content property.
```
SteveSandersonMS: Typo in `parameterized`

test/Microsoft.AspNetCore.Blazor.Razor.Extensions.Test/ComponentTagHelperDescriptorProviderTest.cs
```diff
@@ -695,6 +704,82 @@ public class MyComponent : BlazorComponent
...
+
+namespace Test
+{
+    public class MyComponent : BlazorComponent
+    {
+        [Parameter]
+        RenderFragment<string> ChildContent2 { get; set; }
+
+        [Parameter]
+        string Context { get; set; }
```
SteveSandersonMS: Great, glad this case is covered!

test/testapps/BasicTestApp/MultipleChildContent.cshtml
```diff
@@ -7,7 +7,7 @@
         }
     </Footer>
     <ItemTemplate>
-        <tr><td>@context.Col1</td><td>@context.Col2</td><td>@context.Col3</td></tr>
+        <tr><td>@row.Col1</td><td>@row.Col2</td><td>@row.Col3</td></tr>
```
SteveSandersonMS: Looks better!

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentLoweringPass.cs
```diff
@@ -80,6 +80,13 @@ private ComponentExtensionNode RewriteAsComponent(TagHelperIntermediateNode node
             var visitor = new ComponentRewriteVisitor(component);
             visitor.Visit(node);
 
+            // Fixup the parameter names of child content elements. We can't do this during the rewrite
+            // because we see the nodes in the wrong order.
+            foreach (var childContent in component.ChildContents)
+            {
+                childContent.ParameterName = childContent.ParameterName ?? component.ChildContentParameterName ?? "context";
```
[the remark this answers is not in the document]
rynowak: This is the variable name, the other place is the attribute name. Sure, I'll find a home for these.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -129,8 +129,16 @@ private TagHelperDescriptor CreateDescriptor(BlazorSymbols symbols, INamedTypeSy
                 CreateProperty(builder, property.property, property.kind);
             }
 
-            var descriptor = builder.Build();
+            if (builder.BoundAttributes.Any(a => a.IsParameterizedChildContentProperty()) &&
+                !builder.BoundAttributes.Any(a => string.Equals(a.Name, "Context")))
```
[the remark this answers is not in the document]
rynowak: yeah, I think you're right.

src/Microsoft.AspNetCore.Blazor.Razor.Extensions/Resources.resx
rynowak: yeah I thought that as well when I wrote it. I'll change this.

PR #1471, opened 2018-09-23
Fix negative route params

### LINKED ISSUE #1437: negative route parameter not allowed
Having the following route:

@page "/sendemail/{RoomName}/{RoomId:int}"

I get the following error:

Uncaught (in promise) Error: System.InvalidOperationException: 'Router' cannot find any component with a route for '/confirmemail/2.01/-1'.

when trying to send a negative RoomId. 

No Problem with positive Ids or 0

KristianJakubik:
Fix #1437.

Now you should be able to use negative int and long as url parameters.


SteveSandersonMS:
Thanks, @KristianJakubik! I made a couple of further tweaks so moving this over to #1474

PR #1474, opened 2018-09-24
Fix negative route params. Fixes #1437

### LINKED ISSUE #1437: negative route parameter not allowed

SteveSandersonMS:
Tweaked version of #1471. Thanks, @KristianJakubik!


SteveSandersonMS:
Merging because I think it's uncontroversial but @rynowak please shout at me if you disagree or have concerns.

PR #1475, opened 2018-09-24
Marshal oninput events as UIChangeEventArgs

SteveSandersonMS:
- Blazor does handle the oninput event, but it is marshalled as a regular UIEventArgs
- This means that we cannot access the new value of the input element from inside our oninput handler

Addresses #821 

This is @sebbert's contribution. I just tweaked it by adding a further E2E test. Thanks @sebbert!

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts
@@ -6,6 +6,7 @@ export class EventForDotNet<TData extends UIEventArgs> {
     const element = event.target as Element;
     switch (event.type) {
 
+      case 'input':
       case 'change': {
         const targetIsCheckbox = isCheckbox(element);
         const newValue = targetIsCheckbox ? !!element['checked'] : element['value'];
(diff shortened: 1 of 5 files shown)


SteveSandersonMS:
As the E2E failures show, this is going to require some further changes in the event processing logic to avoid dropping keystrokes while the prior events are still being processed asynchronously.

We need some design around whether (and how) to still invoke event handlers that have already been removed from the render tree.

SteveSandersonMS:
Design ideas:

1. Have the client return an ACK after it processes a render batch. The server will need to keep any event handler delegates in memory until the client finishes processing the render batch in which they were disposed. We know the server can't receieve any more events for those older IDs after this ACK because client->server message delivery order is guaranteed.
2. Or, if we want to avoid the explicit ACK, have some kind of "renderer state ID" that increments with each batch. The client can include this ID with each outbound event notification. The server will need to keep event handler delegates in memory until it receives an event corresponding to a state ID that's >= the one where the event handler ID was disposed.

Option 1 is simpler. If we did option 2, we'd also need some kind of timeout so that event handlers can be disposed after a while even if the client doesn't send any more events.

We might also want to expand `UIEventArgs` to include an `IsStale` flag to signal whether the event was raised from a renderer state other than the most recent one. Then the developer can choose whether or not to process the event depending on the application requirements. Alternatively we could leave it up to the developer to track event validity (e.g., by re-checking whatever condition determines the applicability of the event, such as whether a certain button should be enabled) in the cases where it matters to them.

SteveSandersonMS:
I've implemented the "client sends back ACK" design in #1655. This PR is now rebased onto it, so its E2E tests should now pass.

But for some reason Travis/AppVeyor aren't even running builds for this PR at all. I think it does work, so I'm willing to proceed without Travis/AppVeyor's approval for this PR.

SteveSandersonMS:
Closing this PR and resubmitting as a new one in an attempt to make CI actually run. Not sure why commits on this PR don't trigger any CI hooks when all other PRs do.

PR #1476, opened 2018-09-24
Switch Hosted template to UseBlazor<Startup> for consistency with server-side template. Fixes #1473

### LINKED ISSUE #1473: Switch the ASP.NET Core hosted template to use app.UseBlazor<Client.Startup>()
Currently the ASP.NET Core hosted template points to the `Program` class in the Client project. For consistency with server-side Blazor we should switch it to point to the `Startup` class: `app.UseBlazor<Client.Startup>()`.

SteveSandersonMS:


SteveSandersonMS:
Merging because it's super trivial.

PR #1492, opened 2018-09-28
Make HttpClientHandler wiping more conservative. Fixes #1484

### LINKED ISSUE #1484: Wiped method since 0.6.0-preview1
Hi,

after upgrading from 0.5.1 to 0.6.0, no Http-Calls via Flurl are possible any more:

WASM: [Flurl.Http.FlurlHttpException] Call failed. Cannot invoke method because it was wiped. See stack trace for details.

What to do?

SteveSandersonMS:
This wiping config change doesn't have any meaningful effect on the default app payload size, but it means that Flurl.Http can work in the scenario described at #1484 [1].

It's awkward that Flurl.Http instantiates a default `HttpMessageHandler` even when you're configuring it to use a custom HttpClient, but it does, so this change is to stop wiping the constructor, and just throw if you actually call `SendAsync` (or `Dispose`, which pulls in a huge graph of dependencies from `ServicePointManager`).

Longer term we can hopefully remove the special cases around System.Net.Http, but it depends on exactly what Mono provides in the BCL when they supply a built-in working `HttpClient` and what effect that has on bundle size.

[1] The scenario is:

```csharp
FlurlHttp.Configure(settings =>
{
    settings.HttpClientFactory = new MyHttpClientFactory();
});
```

with

```csharp
class MyHttpClientFactory : DefaultHttpClientFactory
{
    public override HttpClient CreateHttpClient(HttpMessageHandler handler)
        => new HttpClient(new BrowserHttpMessageHandler());
}
```

Notice that the problematic scenario is when we're *not* overriding `CreateMessageHandler`, because if we did, then there wouldn't be a problem anyway and this PR would not be required. In fact that was the original workaround proposed at #1484.

### DIFF
--- a/samples/MonoSanityClient/Examples.cs
+++ b/samples/MonoSanityClient/Examples.cs
@@ -33,7 +33,7 @@ namespace MonoSanityClient
 
         public static void InvokeWipedMethod()
         {
-            new HttpClientHandler();
+            new HttpClientHandler().Dispose();
         }
 
         public static string EvaluateJavaScript(string expression)
--- a/src/mono/bclwipe/System.Net.Http.txt
+++ b/src/mono/bclwipe/System.Net.Http.txt
@@ -1,3 +1,14 @@
 # HttpClientHandler brings in about 500KB and will never be used under WebAssembly anyway
-# (we have a separate browser-compatible handler)
-System.Net.Http.HttpClientHandler
+# (we have a separate browser-compatible handler).
+#
+# To avoid causing issues for Flurl.Http (https://github.com/aspnet/Blazor/issues/1484),
+# we only wipe the minimal set of methods that pull in the large dependency graph and
+# don't do anything useful under WebAssembly anyway. It would be preferable if Flurl.Http
+# didn't use HttpClientHandler in the case where a custom handler has been configured,
+# but at the time of writing it does.
+
+# SendAsync contains the bulk of the outbound references to HttpWebRequest etc.
+System.Net.Http.HttpClientHandler::SendAsync
+
+# Dispose references ServicePointManager which is huge
+System.Net.Http.HttpClientHandler::Dispose


SteveSandersonMS:
@rynowak @danroth27 Proposing to include this in 0.6.0, as it's the only regression reported so far based on 0.6.0-preview1. It's a very safe change to make, because it's reducing the delta between 0.5.1 and 0.6.0.

Suchiman:
Why not use the Negation Operator?
`!System.Net.Http.HttpClientHandler::.ctor`

samples/MonoSanityClient/Examples.cs
```diff
@@ -33,7 +33,7 @@ public static void TriggerException(string message)
 
         public static void InvokeWipedMethod()
         {
-            new HttpClientHandler();
+            new HttpClientHandler().Dispose();
```
Suchiman: Should we then have another test that checks that calling the `ctor` doesn't throw?

SteveSandersonMS:
> Why not use the Negation Operator?

That is what I did at first, but then found it's insufficient because Flurl.Http also sets a few properties, so they have to be retained too. In the end, instead of whitelisting an exact set of things that Flurl does, I thought it would be safer and more precise to blacklist the specific two methods that cause the size bloat.

> Also seems like it would be handy to have a feature to replace a method body with just a ret instead of throwing

Yes, that exact thought occurred to me too. It's not trivial to implement, though. For the resulting code to be valid, you have to return whatever type the method declares, and have to supply values for any "out" params. I didn't think it was urgent enough to get side-tracked into right now. If you're interested in implementing that and think you could make enough test cases to show convincintly it's correct, please go ahead!

Longer term I'm thinking it would be better for the "wiping" feature to be moved into the Mono Linker so it can be applied based on the app's own build config rather than being hardcoded. This would also make it possible to wipe things in NuGet packages, etc., which is not possible under the existing arrangement.

samples/MonoSanityClient/Examples.cs
SteveSandersonMS: This test isn't trying to specify which parts of the BCL are wiped; it's just trying to test that when something *is* wiped, the exception it reports is the one we expect.

Suchiman:
@SteveSandersonMS i didn't intend to change the default, i rather had something in mind along the lines of
```
# wipe and throw
-System.Net.Http.HttpClientHandler::SendAsync
# wipe and ignore
~System.Net.Http.HttpClientHandler::Dispose
```
The rationale being that some Methods may be called unavoidably but they're side effect free or it's something we don't care about

SteveSandersonMS:
Yep, I agree it would be fine if used for `Dispose`. We'd have to be careful to avoid ever using it for something that is meant to have any return values or externally-observable side-effects (which is just about everything except for `Dispose` methods).

tmenier:
@SteveSandersonMS I didn't notice this specific issue until today. Thanks for the effort in making Blazor more compatible with Flurl, and sorry for any headaches Flurl might be causing. I'm working on Flurl 3.0, and I'm open to making changes to better accommodate things on my end, even if they are breaking.

> It's awkward that Flurl.Http instantiates a default `HttpMessageHandler` even when you're configuring it to use a custom HttpClient

The factory interface you're referring to supports 2 methods: `CreateMessageHandler()` and `CreateHttpClient(HttpMessageHandler)`. The former is called and its result is passed to the latter lazily when it's first needed. Flurl needs a default implementation in order for something like `url.GetJsonAsync()` to work out of the box. Typically you only need to customize the construction of `HttpClient` or `HttpMessageHandler` but not both, and most typically the latter. So you can inherit from `DefaultHttpClientFactory` and override only the one you need. It seems like just the right balance of customizability with minimal code. I'm not sure what part of it is awkward.

> so this change is to stop wiping the constructor

Which constructor? `HttpClient(HttpMessageHandler)`? If so, I'm glad to hear that was added back. I think it was the right thing to do. I'm not sure that a library like mine targeting .NET Standard should need to worry that this or any other part of the standard might be "wiped" on a particular platform. Seems to kind of go against the spirit of having a standard. :)

That said, I know your goal is to reduce Blazor's footprint, and to my knowledge you haven't had a lot of issues with libraries other than Flurl. So I want to do my part to make sure Blazor and Flurl can continue to peacefully coexist without requiring any major hacks on either end. If you have any thoughts about what I can do on my end to reduce the pains on yours, let me know here, or feel free to [open a Flurl issue](https://github.com/tmenier/Flurl/issues/new). Thanks!

PR #1494, opened 2018-09-28
Added Proteced Property for the `RenderHandle` of the `BlazorComponent`

sharpninja:
What: Added a `protected` property to `BlazorComponent` that exposes the `RenderHandle` to sub-classes.

Why: This allows sub-classes to be instantiated individually with the ability to set the `RenderHandle` from the sub-class's `IComponent.Init` method, which prevents the `BlazorComponent` to not have the ability to be rendered when accessed as `BlazorComponent`.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -32,6 +32,16 @@ namespace Microsoft.AspNetCore.Blazor.Components
 
         private readonly RenderFragment _renderFragment;
         private RenderHandle _renderHandle;
+
+        /// <summary>
+        /// Exposes the RenderHandle to any sub-classes
+        /// </summary>
+        /// <value>RenderHandle</value>
+        protected RenderHandle RenderHandle {
+            get => _renderHandle;
+            set => _renderHandle = value;
+        }
+
         private bool _hasCalledInit;
         private bool _hasNeverRendered = true;
         private bool _hasPendingQueuedRender;


sharpninja:
Yes, I want to control instantiation of the sub-class of the component.  Any subclass that wants to be able to trigger a render on itself must have a reference to _renderHandle.  This is not possible without the Get part of the new property.  The Set part of the property allows a sub-class of BlazorLayoutComponent to instantiate the sub-class of BlazorComponent and assign it the RenderHandle that is local to the sub-class of the BlazorLayoutComponent.

Why do I want to do this?  I am creating a navigation library for Blazor that directly models that of UWP, Android and iOS.  I believe that the Url based navigation is a huge hindrance to the flexibility of Blazor.  Also, this would break us free from the tyranny of the browser's back button as we can use JavaScript to disable it and force navigation to be controlled within the app.  This is desirable for embedding applications with Blazor such as Electron or custom browsers so that you can completely remove the normal browser UI elements and make a completely self-contained application environment.

I am willing to use other mechanisms to perform the instantiation of a sub-class of BlazorComponent that can still trigger it's own render.

sharpninja:
BTW, wrapping _renderHandle breaks encapsulation only in so much as _renderHandle should have been a protected property or field to begin with as it should be reasonable that any sub-class of BlazorComponent should have the ability to trigger it's own Render and other functionality requiring the _renderHandle.  I have the code for my Navigation framework at home.  I will put up a NuGet project for it today and post a link here to the pertinent code.

sharpninja:
>           My best guess is that maybe you're hoping to instantiate the components directly and then somehow tie them into the rendering process. If so, this is not a desirable pattern, as it would not match up with other aspects of the component lifecycle such as DI and disposal, or of course future changes to the lifecycle.

I've been working on performing the draw via other methods and I believe I'm on to an alternative.

Having said that, I think there's a strong argument for a blazor component factory that could take in a context, instantiate from DI, and then apply that context to the component, giving the initialized component back to the caller for manipulation.  I'm working on something similar and may focus more effort in that direction.

SteveSandersonMS:
OK, thanks for the update. I'll close this then, but please let us know how you get on with your alternate approach.

PR #1498, opened 2018-09-28
Upgrade KoreBuild

natemcmaster:
* remove sources.props and use NuGet.config. Source.props is only necessary for products that bulid in ProdCon, which Blazor does not
* Remove code signing glue code. This is part of KoreBuild now
* Update SDK to 2.1.400
* Update certificate names used for code-signing to use SHA2


natemcmaster:
This uncovered a bug in KoreBuild. :watch: on https://github.com/aspnet/BuildTools/pull/769

Directory.Build.props
```diff
@@ -5,7 +5,6 @@
 
   <Import Project="version.props" />
   <Import Project="build\dependencies.props" />
-  <Import Project="build\sources.props" />
```
rynowak: Curious? Why is nuget.config better?

src/Microsoft.AspNetCore.Blazor.Browser.JS/package-lock.json
```diff
@@ -1540,12 +1540,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
rynowak: @SteveSandersonMS  - is this expected?

SteveSandersonMS: It's hard to say. NPM has never yet managed to get `package-lock.json` updates to be consistent across machines or NPM versions. It feels like it's getting more consistent over time, but they do keep changing stuff.

On the appearance of `optional:true`, it's probably related to one of these two:

 * https://github.com/npm/npm/issues/19016 (a feature request implemented in NPM 5.6.0)
 * https://npm.community/t/package-lock-json-keeps-changing-between-platforms-and-runs/1129/8 (a bug report that is still open, specifically mentioning the `optional` flag)

I'm not worried about the appearance of this flag because it's supposedly back-compatible. Fine for this PR to include it here, in my view.

natemcmaster:
It appears from the internal build that microbuild signing tasks do not load in MSBuild for .NET Core. I believe this workaround should fix the issue: https://github.com/aspnet/BuildTools/pull/771

Directory.Build.props
rynowak: ok dope

natemcmaster:
Updated. New internal build for PR validation: https://dnceng.visualstudio.com/internal/_build/results?buildId=26221&_a=summary&view=logs

Directory.Build.props
JamesNK: Friendship ended with `RestoreSources`.
Now `NuGet.config` is my best friend.

rynowak: I still like restore sources better. There are too many files.

SteveSandersonMS:
Thanks for this, @natemcmaster!

PR #1507, opened 2018-10-01
Add required fields to build nuspec

rynowak:
These required fields are missing from the .Build nuspec, and this is blocking package upload.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.nuspec
+++ b/src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.nuspec
@@ -4,7 +4,10 @@
     <id>Microsoft.AspNetCore.Blazor.Build</id>
     <version>$version$</version>
     <authors>Microsoft</authors>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
     <description>Build mechanism for Blazor applications.</description>
+    <licenseUrl>https://raw.githubusercontent.com/aspnet/Home/2.0.0/LICENSE.txt</licenseUrl>
+    <projectUrl>https://asp.net/</projectUrl>
     <dependencies>
       <dependency id="Microsoft.AspNetCore.Razor.Design" version="$razorversion$" include="all" />
       <dependency id="Microsoft.AspNetCore.Blazor.Analyzers" version="$version$" include="all" />


src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.nuspec
```diff
@@ -4,7 +4,10 @@
     <id>Microsoft.AspNetCore.Blazor.Build</id>
     <version>$version$</version>
     <authors>Microsoft</authors>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
     <description>Build mechanism for Blazor applications.</description>
+    <licenseUrl>https://raw.githubusercontent.com/aspnet/Home/2.0.0/LICENSE.txt</licenseUrl>
+    <projectUrl>https://asp.net/</projectUrl>
```
Suchiman: Why not the link to this GitHub repository? I always find that very unhelpful when i click on Project Page on nuget just to land on the generic asp.net Homepage.

SteveSandersonMS: Sounds like a good suggestion to me, but maybe @rynowak knows of some reason why it has to be just `https://asp.net/`. I'm OK with it either way.

SteveSandersonMS:
Actually I'm going to merge this right away so we can be ready with builds ASAP.

Regarding @Suchiman's suggestion about project URL, if @rynowak thinks the change is valid we can do it after.

Regarding the AppVeyor failure, this is bogus. It's annoying that this one test is flaky - I have looked at it, and it's unclear from the sources what's different compared with other tests. And the failure only seems to happen on AppVeyor. If this continues I will have to do something about it, but not blocking this PR on it as it's unrelated.

src/Microsoft.AspNetCore.Blazor.Build/Microsoft.AspNetCore.Blazor.Build.nuspec
rynowak: 😆 - Dan and I had the same discussion before I opened this PR. Since we wanted to get this done in a hurry, I just did what the existing packages do. How about it @Eilon - is this allowed to be the github repo URL?

Meir017: I think you should also add the repositoryUrl - https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#repositoryurl (like [this](https://github.com/cake-build/cake/pull/2240/files))

Eilon: The project URL is correct - it's part of ASP.NET. But indeed the *repo* URL should point to this repo.

Issue #1483, opened 2018-09-25
Wrong data type in UIKeyboardEventArgs.Location declaration


Andrzej-W:
`UIKeyboardEventArgs.Location` is currently declared as `float` but according to this info:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/location
it can contain only one of a few unsigned integer values. I think this property should be `enum`?

SteveSandersonMS:
Thanks for the suggestion! It sounds good to me. I'll put this as "up for grabs".

KristianJakubik:
Hi, I would like to work on this, if it is possible.

Niladri24dutta:
@SteveSandersonMS  Hi , I have made some changes for this in my branch . Can you please take a look and provide your feedback if this can be a PR
https://github.com/Niladri24dutta/Blazor/blob/add-keyboardlocation-enum/src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs#L206

Niladri24dutta:
@SteveSandersonMS However I noticed that the value of `UIKeyboardEventArgs.Location` is only set during `keydown` and `keyup` event. The interface `KeyboardEvent` already have the property `location` along with 4 readonly properties to indicate the location of the key in the the type definition library. So I have made the changes in the `EventForDotNet.ts` file also .  Please let me know if it can be simplified for testing in with selenium (sending a shift or right ctrl key).

Here is the change 
https://github.com/Niladri24dutta/Blazor/blob/add-keyboardlocation-enum/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts#L171 

https://github.com/Niladri24dutta/Blazor/blob/add-keyboardlocation-enum/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/EventForDotNet.ts#L346

PR #1515, opened 2018-10-02
Fixing types of UIEventArgs properties.

KristianJakubik:
Hi, 
I found out that some properties of [UIEventArgs ](https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs)objects doesn't match with either w3 standard or with their description.

Example:
`public float Detail { get; set; }` 
should be based on [mozilla](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail) or [w3](https://www.w3.org/TR/uievents/#interface-uievent)
`public long Detail { get; set; }`
Even the description of property says, that it represents count of clicks incremented by 1, therefor  representing them by `long` makes more sense to me.

I found several other properties with mismatching types. I can fix them and propose fixes in this PR, but at first I would like to get some feedback on this.

Thanks.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs
+++ b/src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs
@@ -41,7 +41,7 @@ namespace Microsoft.AspNetCore.Blazor
         /// <summary>
         /// A count of consecutive clicks that happened in a short amount of time, incremented by one.
         /// </summary>
-        public float Detail { get; set; }
+        public long Detail { get; set; }
 
         /// <summary>
         /// The data that underlies a drag-and-drop operation, known as the drag data store.
@@ -258,7 +258,7 @@ namespace Microsoft.AspNetCore.Blazor
         /// <summary>
         /// A count of consecutive clicks that happened in a short amount of time, incremented by one.
         /// </summary>
-        public float Detail { get; set; }
+        public long Detail { get; set; }
 
         /// <summary>
         /// The X coordinate of the mouse pointer in global (screen) coordinates.
@@ -403,7 +403,7 @@ namespace Microsoft.AspNetCore.Blazor
         /// <summary>
         /// A count of consecutive clicks that happened in a short amount of time, incremented by one.
         /// </summary>
-        public float Detail { get; set; }
+        public long Detail { get; set; }
 
         /// <summary>
         /// A list of <see cref="UITouchPoint"/> for every point of contact currently touching the surface.


Andrzej-W:
Maybe you should cooperate with @Niladri24dutta here:
https://github.com/aspnet/Blazor/issues/1483

Niladri24dutta:
I have posted some more comments here regarding this #1483

KristianJakubik:
@Niladri24dutta 
Yeah, I know you have been working on the property `Location` and on changing it to `Enum`. I had no intention to solve it in this PR, because I though you have been focusing only on `Location` prop.
 
But I found several others like `Detail`, `Height` and `Width`, which should't be `float` based on docs.
I certainly didn't check all of them, so there could be more to change. 
I suppose those changes should be in different PR than PR for #1483.

If it makes sense to solve it in the same PR, then we should go for it.
@SteveSandersonMS What are you suggesting?

Andrzej-W:
@KristianJakubik are you sure that Height and Width should't be float? I have laptop with DPI scaling set to 150% and I'm sure I have seen values like 100.5px in window.innerWidth for example. Here is a note from http://api.jquery.com/innerwidth/
> The number returned by dimensions-related APIs, including .innerWidth(), may be fractional in some cases. Code should not assume it is an integer.

There a lot of different event types in https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor/UIEventArgs.cs with ScreenX/ScreenY/ClientX/ClinetY properties and all of them have type `long` which is consistent with the spec:
https://www.w3.org/TR/uievents/#interface-mouseevent
On the other hand here:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/clientX
we can read
> Originally, this property was defined as a long integer. The CSSOM View Module redefined it as a `double float`.

What should we have in Blazor? Personally I think that we can always convert from double to int or long but when we have only an integer value there is no way to restore fractional part.

KristianJakubik:
@Andrzej-W 
Thanks for your response.

> @KristianJakubik are you sure that Height and Width should't be float? I have laptop with DPI scaling set to 150% and I'm sure I have seen values like 100.5px in window.innerWidth for example. Here is a note from http://api.jquery.com/innerwidth/

I went through docs again and you are right. It says, that it should be `double` . I will edit my comment.
docs: [mozzila](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/height), [w3](https://w3c.github.io/pointerevents/#widl-PointerEvent-height)

> What should we have in Blazor? Personally I think that we can always convert from double to int or long but when we have only an integer value there is no way to restore fractional part.

My personal opinion is, that we should be as accurate as possible for objects, which are accessible from outside of the framework. If I was a user of Blazor, I wouldn't be happy, if I had to convert the value every time when I'm working with `UIEventArgs` object. That's the main reason why I created this PR.

SteveSandersonMS:
Thanks for this! It's now moved to https://github.com/aspnet/AspNetCore/pull/4595.

Issue #1513, opened 2018-10-02
support help in Blazor cli


Meir017:
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <RunCommand>dotnet</RunCommand>
    <RunArguments>blazor serve</RunArguments>
    <LangVersion>7.3</LangVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Blazor.Browser" Version="0.6.0" />
    <PackageReference Include="Microsoft.AspNetCore.Blazor.Build" Version="0.6.0" PrivateAssets="all" />

    <DotNetCliToolReference Include="Microsoft.AspNetCore.Blazor.Cli" Version="0.6.0" />
  </ItemGroup>

</Project>
```
using blazor `0.6.0`

after running `dotnet new blazor` and then `dotnet blazor -h`
an `InvalidOperationException` is thrown.

is there any reason why blazor does not use the `CommandLineApplication`?

SteveSandersonMS:
> is there any reason why blazor does not use the CommandLineApplication?

The `blazor` CLI was one of the very first parts of the system we built, and is mostly intended to be an internal implementation detail. There aren't currently any reasons why developers need to run it manually, since it only has one feature, "serve", and that's done for you automatically when you execute `dotnet run`. So, adding a help system has never been a high priority.

If this is something you feel strongly about, we'd certainly consider a PR for this if it followed the same patterns as the other `Microsoft.Extensions.CommandLineUtils`-based tools in the repo, e.g., `Microsoft.AspNetCore.Blazor.BuildTools`. Thanks!

PR #1524, opened 2018-10-05
using Microsoft.Extensions.CommandLineUtils for blazor-cli

Meir017:
related to #1513

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Cli/Microsoft.AspNetCore.Blazor.Cli.csproj
+++ b/src/Microsoft.AspNetCore.Blazor.Cli/Microsoft.AspNetCore.Blazor.Cli.csproj
@@ -11,6 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="1.1.1" />
     <PackageReference Include="Microsoft.AspNetCore" Version="$(AspNetCorePackageVersion)" />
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Blazor.Cli/Microsoft.AspNetCore.Blazor.Cli.csproj
```diff
@@ -11,6 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="1.1.1" />
```
rynowak: @SteveSandersonMS - note this is using the publicly shipped build of CommandLineUtils. We shipped this on nuget.org (by mistake) in the 1.x releases and we continue to maintain and build it internally (on myget). I'm not aware of a significant delta between 1.x and the current bits - so I think this is fine for now. When/if this code starts building along with the product we should update.

Meir017: it's the same version that you use in the build tools https://github.com/aspnet/Blazor/blob/5f6de9b6f50bdf849e661a4f85ea1400228f7e88/src/Microsoft.AspNetCore.Blazor.BuildTools/Microsoft.AspNetCore.Blazor.BuildTools.csproj#L10

PR #1526, opened 2018-10-05
Give Renderer knowledge of component parents

SteveSandersonMS:
Until now, `Renderer` knew which components were the children of another (i.e., the ones in its most recent render tree), but didn't track the relationship in the opposite direction because it didn't need to.

We now do need to be able to walk up the ancestry hierarchy, because this is essential for the upcoming "passing parameters down the hierarchy" feature (when a new child is added, we have to be able to walk up the ancestry to find matches for its ambient params).

Notes:

* It's kind of awkward, but there was no direct way to unit test this until the rest of the "hierarchical parameters" feature is implemented, because it's just an internal implementation detail. So I've temporarily added an `internal` method for exposing the hierarchy to tests. I'll remove this later.
* While implementing this, it became clear that `AssignComponentId` would need to be split into two variants. Previously it was used publicly but for root components only, and privately for non-root components. I've now reflected this in the code by renaming the `public` variant to `AssignRootComponentId`, and having only the `private` variant accept a `parentComponentId`.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Rendering/ComponentState.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Rendering/ComponentState.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Blazor/Rendering/ComponentState.cs
```diff
@@ -15,6 +15,7 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
     internal class ComponentState
     {
         private readonly int _componentId; // TODO: Change the type to 'long' when the Mono runtime has more complete support for passing longs in .NET->JS calls
+        private readonly ComponentState _parentComponentState;
```
SteveSandersonMS: Tracking this new data is the main result of this PR. We don't do anything with it yet (hence the artificialness of the unit test coverage), but later on it will be key to having the renderer wire up the deep params.

rynowak:
I am taking a look at this, but it's going to take me a little time to process.

PR #1534, opened 2018-10-07
Simple 404 Handling (Fallback Route)

TheFanatr:
I've talked about this [here](https://github.com/aspnet/Blazor/issues/293#issuecomment-427678571).

Essentially, this pull request contains a very simple error 404 handling implementation that allows users to set a fallback route for the blazor routing system to use if the no components are found for the requested route. This is a essentially a catch-all-nonexistent routing handler, where the rendering of the page body is "soft-redirected" to the component represented by the fallback route. The value is set in the `App.cshtml` file or wherever the router is defined.

```cshtml
<Router AppAssembly=typeof(Program).Assembly FallbackRoute="/404"/>
```

If `FallbackRoute` is null (unset), or also points to a nonexistent component, the regular behaviour will be observed, with the only difference being that the exception thrown will say that the router `cannot find any component with the fallback route '/<fallback>'` if the fallback route is also nonexistent.

The URI displayed by the browser does not change to represent the fallback route in order to allow the fallback component to implement JavaScript interop for handling specific URIs differently, such as the URI for a page that was moved.


vertonghenb:
I'm wondering if after this implementation it's still possible to navigate to a MVC route not specified in the Blazor Routing system itself. 

See: https://github.com/aspnet/Blazor/pull/1154

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
```diff
@@ -74,15 +79,21 @@ protected virtual void Render(RenderTreeBuilder builder, Type handler, IDictiona
...
             if (context.Handler == null)
             {
-                throw new InvalidOperationException($"'{nameof(Router)}' cannot find any component with a route for '/{locationPath}'.");
+                if (!useFallback && FallbackRoute != null)
+                {
+                    Refresh(useFallback: true);
+                    return;
+                }
+                else
+                    throw new InvalidOperationException($"'{nameof(Router)}' cannot find any component with {(useFallback ? "the fallback route" : "a route for")} '/{locationPath}'.");
```
Suchiman: This can be simplified to
```csharp
if (useFallback || FallbackRoute == null)
    throw new InvalidOperationException($"'{nameof(Router)}' cannot find any component with {(useFallback ? "the fallback route" : "a route for")} '/{locationPath}'.");

Refresh(useFallback: true);
```

TheFanatr: Thanks! Will change it.

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
```diff
@@ -31,6 +31,11 @@ public class Router : IComponent, IDisposable
         /// </summary>
         [Parameter] private Assembly AppAssembly { get; set; }
 
+        /// <summary>
+        /// Gets or sets the route that should be used when no components with are found with the requested route.
```
Suchiman: > components with are found with

typo: double `with`

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
TheFanatr: Actually, I need the return in there otherwise it will fail because the original route gets rendered anyways in certain cases.

Suchiman: Oh good point, my bad, i missed that the method didn't end there

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
TheFanatr: Fixed it and ironically put another typo in the commit message.

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
TheFanatr: Fixed.

TheFanatr:
@vertonghenb This should still be possible as long as those routes are processed before the Blazor router tries to have them rendered as if they point to components, and/or if an MVC route will override a Blazor route. Will have to test this.

vertonghenb:
@TheFanatr Alright, love the idea though!

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
```diff
@@ -74,15 +79,19 @@ protected virtual void Render(RenderTreeBuilder builder, Type handler, IDictiona
...
         {
-            var locationPath = UriHelper.ToBaseRelativePath(_baseUri, _locationAbsolute);
+            var locationPath = useFallback ? FallbackRoute : UriHelper.ToBaseRelativePath(_baseUri, _locationAbsolute);
             locationPath = StringUntilAny(locationPath, _queryOrHashStartChar);
             var context = new RouteContext(locationPath);
             Routes.Route(context);
             if (context.Handler == null)
             {
-                throw new InvalidOperationException($"'{nameof(Router)}' cannot find any component with a route for '/{locationPath}'.");
+                if (useFallback || FallbackRoute == null)
```
SteveSandersonMS: I know this is a really minor point, but for code style consistency, could we have braces around line 91?

SteveSandersonMS:
This looks like a good design to me. Thanks @TheFanatr!

We'll need to add one or more E2E tests like the other router E2E tests. If you're interested in adding that please go ahead and we might be able to merge this sooner. No worries if not, as I'll add them in a week or two when I get to merging this.

src/Microsoft.AspNetCore.Blazor/Routing/Router.cs
TheFanatr: Changed and committed.

TheFanatr:
Thanks. I'll look into the E2E tests, but may not end up actually making them before you do.

TheFanatr:
@SteveSandersonMS I added an E2E test called `CanArriveAtFallbackPageFromBadURI` which tries to navigate to the URI for a route that has no components and tests whether or not the fallback route's component was then rendered. I've set the fallback route to point to the default page, as in the `/` route in `TestRouter`. It apparently succeeds according to the continuous integration systems, so please take a look and see if there's anything else I need to add and/or change.

test/testapps/BasicTestApp/RouterTest/TestRouter.cshtml
```diff
@@ -1 +1 @@
-<Router AppAssembly=typeof(BasicTestApp.Program).Assembly />
+<Router AppAssembly=typeof(BasicTestApp.Program).Assembly FallbackRoute="/" />
```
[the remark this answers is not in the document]
TheFanatr: I added a component representing a 404 page that is routed to as the fallback.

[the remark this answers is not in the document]
TheFanatr: Sorry this took so long; I had not noticed the notification of your review.

TheFanatr:
@danroth27 I added a component called `Error404` in `BasicTestApp` and it gets routed to as the fallback on the `TestRouter`. I also merged master, so everything else is up to date. Please let me know if there is anything else I need to do.

TheFanatr:
@SteveSandersonMS should I migrate this to the Razor Components section of the ASP.NET Core repo?

TheFanatr:
I've created a pull request on the main repo now.

PR #1540, opened 2018-10-08
Updated link in README.md

thareshjose:


SteveSandersonMS:
Sorry, I don't quite understand why this is the right change to make. The existing URL forwards to the same place. Could you clarify why this should be changed? Thanks!

Eilon:
Yeah this is not what we want, so closing it. We use redirect links so that we can easily change them in the future without updating lots of docs.

(@SteveSandersonMS - this contributor sent a similar PR to another repo recently and it met the same fate.)

Issue #1009, opened 2018-06-15
AssignToProperties does not empty properties that are not on the parameter collection


RemiBou:
### Functional impact
We can't follow a link that sets a parameter then go back to the page without the parameter.

### Minimal repro steps
If I have this page
```
@page "/tag/{CurrentTag}"
@page "/tag/"
 <NavLink href="/tag/" Match="NavLinkMatch.All"  class="btn">
  no tag
</NavLink>
 <NavLink href="/tag/test" Match="NavLinkMatch.All"  class="btn">
  test
</NavLink>
Current tag : @CurrentTag 
@functions{
    [Parameter]
    private string CurrentTag { get; set; } = "";
}
```
1. I navigate on "/tag/" 
2. I click on "test" 
3. I click on "no tag" 
4. the page still displays "test" instead of "no tag"

### Further technical details
This is because here https://github.com/aspnet/Blazor/blob/29f925852b0a5e81e82780e0c1c32b486f7c3be6/src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs we just assign the parameters currently on the collection and we don't empty other parameters. We should add a loop through all the parameterWriters without a corresponding item in the collection and set them to default.

As a workaround I add to add this to my page :
```
public override void SetParameters(ParameterCollection parameters)
    {
        if (string.IsNullOrEmpty(parameters.GetValueOrDefault<string>("CurrentTag")))
        {
            CurrentTag = null;
        }
        base.SetParameters(parameters);
    }
```
I can send a PR on this.

RemiBou:
I found the Unit Test "NoIncomingParameterMatchesDeclaredParameter_LeavesValueUnchanged" so I guess it's a desired behavior, maybe we could do it on the routing phase, when a pages has multiple routes with different parameters it could add empty values to the parameter collection.

SteveSandersonMS:
> I found the Unit Test "NoIncomingParameterMatchesDeclaredParameter_LeavesValueUnchanged" so I guess it's a desired behavior

It is currently by design, but that doesn't mean the design is necessarily right. We're open to reconsidering!

Is it even possible for a `ParameterCollection` to know which unspecified `[Parameter]` properties on the target object exist (without resorting to further reflection)? If this can be achieved without any extra perf cost, then we'd consider changing this design. I can see there are cases where it would be preferable to change the meaning of `AssignToProperties` to mean "ensure *all* parameter properties match this collection".

[...]

PR #1545, opened 2018-10-09
Cascading parameters

SteveSandersonMS:
This PR includes a long list of changes, but hopefully should be reasonably understandable if read one commit at a time. Many of the commits are (simple-ish) refactorings that were necessary ahead of subsequent work, e.g., to generalize things or make more info available to the system. I've prefixed each such commit message with `Refactor:` so it should be clear which commits introduce new functionality and which don't.

Since this PR has become overwhelmingly long already, I propose to merge this once the code review is completed based on the functionality here. Then I'll do remaining work in a further PR. Specifically:

### To do in a follow-up PR

- Some indicator for whether each incoming parameter is a cascading parameter or not
- Option to declare a `<CascadingValue>` as "static" (or maybe call it `FixedValue` to avoid clash of terminology)
- E2E tests


RemiBou:
Is this linked to issue #1009 ?

SteveSandersonMS:
@RemiBou No, it's independent of that.

stavroskasidis:
I the saw a demo of this in the ASP.NET Community Standup and I have to say, good job.

I think this will help me remove some hacks/workarounds I have used in my [context menu component](https://github.com/stavroskasidis/BlazorContextMenu/blob/develop/BlazorContextMenu/Components/Item.cshtml#L196).

conficient:
Would tree parameters support multiple instances?

In your ASP.NET community standup example you demo'd both a Theme parameter and a Tab parameter. Would it be possible to have more than one, so the theme controls the styling (at a layout level) and the tabset controls the tab (from the parent control). I guess it would be by the `type` of the parameter as there was no naming being implemented.

conficient:
Oh, and presumably you can override a provider with a second instance. e.g. the Theme `Provider` on the layout is overridden on a specific page which uses a different Theme `Provider`?

SteveSandersonMS:
Yes to both.

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
```diff
@@ -0,0 +1,93 @@
...
+            // Implementing the parameter binding manually, instead of just calling
+            // parameters.AssignToProperties(this), is just a very slight perf optimization
+            // and makes it simpler impose rules about the params being required or not.
+
+            var hasSuppliedValue = false;
+            Value = null;
+            ChildContent = null;
+            Name = null;
+
+            foreach (var parameter in parameters)
```
rynowak: Unrelated node, I suspect `ParameterEnumerator` would be improved by switching to a pattern like this: https://github.com/aspnet/Routing/blob/release/2.2/src/Microsoft.AspNetCore.Routing.Abstractions/RouteValueDictionary.cs#L598 (Assign `Current` inside `MoveNext()`) - `Current` is not going to get inlined anyway with an exception inside it.

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
```diff
@@ -0,0 +1,93 @@
...
+                else if (parameter.Name.Equals(nameof(ChildContent), StringComparison.OrdinalIgnoreCase))
+                {
+                    ChildContent = (RenderFragment)parameter.Value;
+                }
+                else if (parameter.Name.Equals(nameof(Name), StringComparison.OrdinalIgnoreCase))
+                {
+                    Name = (string)parameter.Value;
+                    if (string.IsNullOrEmpty(Name))
+                    {
+                        throw new ArgumentException($"The parameter '{nameof(Name)}' for component '{nameof(Provider)}' does not allow null or empty values.");
```
rynowak: nit: there's an argument (pun intended) to be made for using another exception type here since theres error aren't strictly related to the arguments of this method. I tend not to use `ArgumentException` unless I have an argument name to provide.

I don't think it's important, but rather we should make a decision and then be consistent for the built-in components. I suspect it is already consistent since you wrote them!

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        /// <summary>
+        /// The content to which the value should be provided.
+        /// </summary>
+        [Parameter] private RenderFragment ChildContent { get; set; }
+
+        /// <summary>
+        /// The value to be provided.
+        /// </summary>
+        [Parameter] private object Value { get; set; }
```
rynowak: I assume I'll find the answer when I look at subsequent commits, but it looks like based on just this commit that `Value` doesn't go anywhere yet 😆

src/Microsoft.AspNetCore.Blazor/Components/ParameterAttribute.cs
```diff
@@ -11,5 +11,26 @@ namespace Microsoft.AspNetCore.Blazor.Components
...
+        /// provided value. When <see cref="FromTree"/> is false, this value is
+        /// not used.
+        ///
+        /// If a value is specified, the parameter value will be supplied by the
+        /// closest ancestor Provider that supplies a value with this name.
+        ///
+        /// If no value is supplied, the parameter value will be supplied by the
+        /// closest ancestor Provider that supplies a value with a compatible type.
+        /// </summary>
+        public string ProviderName { get; set; }
```
rynowak: Based on our discussion yesterday, looking forward to this becoming its own attribute since it won't have much in common with *true* parameters when we're done.

galvesribeiro: Yeah, I think it should be something like `ProviderParameterAttribute`

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.AspNetCore.Blazor.Components
     /// <summary>
     /// A component that provides one or more parameter values to all descendant components.
     /// </summary>
-    public class Provider : IComponent
+    public class Provider<T> : ProviderBase, IComponent
```
rynowak: What's the advantage of turning this into a generic? Is the idea to make it based on declared type rather than runtime type?

rynowak: My only concern is about cases that don't work super well as generics - like delegate conversions

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
```diff
@@ -66,6 +66,10 @@ public static class ParameterCollectionExtensions
             }
         }
 
+        internal static IEnumerable<PropertyInfo> GetBindableProperties(Type targetType)
+            => MemberAssignment.GetPropertiesIncludingInherited(targetType, _bindablePropertyFlags)
```
rynowak: I'm not super familiar with this code, but I suspect it's similar to https://github.com/aspnet/Common/blob/master/shared/Microsoft.Extensions.PropertyHelper.Sources/PropertyHelper.cs

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
```diff
@@ -85,6 +85,17 @@ public void SetParameters(ParameterCollection parameters)
...
+        internal override bool CanSupplyValue(Type requestedType, string requestedName)
+        {
+            if (!requestedType.IsAssignableFrom(typeof(T)))
+            {
+                return false;
+            }
+
+            return (requestedName == null && Name == null) // Match on type alone
+                || string.Equals(requestedName, Name, StringComparison.Ordinal); // Also match on name
+        }
```
rynowak: So really how we should think about `name` based on this code is that a named value cannot be bound to an unnamed reciever. I think it makes sense, I just hadn't gone through it mentally yet.

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
```diff
@@ -85,6 +85,17 @@ public void SetParameters(ParameterCollection parameters)
...
 
+        internal override bool CanSupplyValue(Type requestedType, string requestedName)
+        {
+            if (!requestedType.IsAssignableFrom(typeof(T)))
+            {
+                return false;
+            }
+
+            return (requestedName == null && Name == null) // Match on type alone
+                || string.Equals(requestedName, Name, StringComparison.Ordinal); // Also match on name
```
rynowak: Are we sure about this being case sensitive when things like attribute/parameter names are case-insensitive?

src/Microsoft.AspNetCore.Blazor/Components/ProviderBase.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Blazor.Components
+{
+    /// <summary>
+    /// Non-generic base class for <see cref="Provider{T}"/>.
+    /// </summary>
```
rynowak: Might as well say that it's the best class for all providers.

src/Microsoft.AspNetCore.Blazor/Components/ProviderBase.cs
```diff
@@ -0,0 +1,18 @@
...
+    public abstract class ProviderBase
+    {
+        // This base class exists only so that TreeParameterState has a way
+        // to work with all Provider types regardless of their generic param.
+
+        internal abstract bool CanSupplyValue(Type valueType, string providerName);
+    }
```
rynowak: Is the idea of this that the base class is public so that can expose `Provider` but so that no one but us can implement it? Consider an interface as well, which wouldn't need to be public.

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
```diff
@@ -0,0 +1,129 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Blazor.Components
+{
+    internal readonly struct TreeParameterState
+    {
+        private readonly static IDictionary<Type, ReflectedTreeParameterInfo[]> _cachedInfos
```
rynowak: nit: don't use `IDictionary<>` here. This will be more efficient, and concurrent dictionary exposes lots of API that `IDictionary<>` doesn't.

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
```diff
@@ -0,0 +1,129 @@
...
+        public string LocalName { get; }
+        public ProviderBase FromProvider { get; }
+
+        public TreeParameterState(string localName, ProviderBase fromProvider)
+        {
+            LocalName = localName;
+            FromProvider = fromProvider;
+        }
+
+        public static ICollection<TreeParameterState> FindTreeParameters(ComponentState componentState)
```
rynowak: nit: consider `IReadOnlyList<>`  or just a List or array if this is an implementation detail. Those types are way more efficient to consume.

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
```diff
@@ -0,0 +1,129 @@
...
+                    return true;
+                }
+
+                componentState = componentState.ParentComponentState;
+            } while (componentState != null);
+
+            // No match
+            provider = null;
+            return false;
+        }
```
rynowak: elegant and cool.

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
```diff
@@ -0,0 +1,129 @@
...
+                    }
+
+                    result.Add(new ReflectedTreeParameterInfo(
+                        prop.Name,
+                        prop.PropertyType,
+                        parameterAttribute.ProviderName));
+                }
+            }
+
+            return result?.ToArray();
```
rynowak: consider using using `Array.Empty<>` as a sentinel, then you don't need null checks elsewhere, and it's free.

test/Microsoft.AspNetCore.Blazor.Test/TreeParameterStateTest.cs
```diff
@@ -0,0 +1,419 @@
...
+
+            // Assert
+            Assert.Collection(result, match => {
+                Assert.Equal(nameof(ComponentWithGenericTreeParam<object>.LocalName), match.LocalName);
+                Assert.Same(states[0].Component, match.FromProvider);
+            });
+        }
+
+        [Fact]
+        public void FindTreeParameters_ComponentRequestsImplementedInterface_ReturnsMatches()
```
rynowak: I think there's a decision point here about the combination of `is` based on the declared type of `T` for some `Provider<T>`. Generally you do one of these things or the other.

The pairings I expect to see are:

**declared type + `typeof(T) == ParameterType`** => Type is used as a 'key' or identity, rather then representing a set of optional policies/protocols

**runtime type + `ParameterType.IsAssignableFrom(Value.GetType())`** => Value can implement a set of protocols and binding is flexible

I expect the reason why it is like it is that you didn't have the value available in all cases in `CanSupplyValue` OR because you wanted the binding between to ancestor and parent to stay stable when the same set of components renders?

rynowak: Another consideration that I can see you thought of is `typed null`. Maybe what's here is great, I just want to talk a little more about the rationale.

test/Microsoft.AspNetCore.Blazor.Test/TreeParameterStateTest.cs
```diff
@@ -0,0 +1,419 @@
...
+
+            // Assert
+            Assert.Collection(result, match => {
+                Assert.Equal(nameof(ComponentWithNamedTreeParam.SomeLocalName), match.LocalName);
+                Assert.Same(states[0].Component, match.FromProvider);
+            });
+        }
+
+        [Fact]
+        public void FindTreeParameters_MultipleMatchingAncestors_ReturnsClosestMatches()
```
rynowak: is it worth having the same test, but where a `null` is the resulting value?

src/Microsoft.AspNetCore.Blazor/RenderTree/RenderTreeFrame.cs
```diff
@@ -112,11 +113,17 @@ namespace Microsoft.AspNetCore.Blazor.RenderTree
         /// </summary>
         [FieldOffset(16)] public readonly Type ComponentType;
 
+        /// <summary>
+        /// If the <see cref="FrameType"/> property equals <see cref="RenderTreeFrameType.Component"/>,
+        /// gets the child component state object. Otherwise, the value is undefined.
+        /// </summary>
+        [FieldOffset(24)] internal readonly ComponentState ComponentState;
```
rynowak: I assume not serialized

src/Microsoft.AspNetCore.Blazor/Rendering/ComponentState.cs
```diff
@@ -96,5 +96,11 @@ public void DispatchEvent(EventHandlerInvoker binding, UIEventArgs eventArgs)
 
         public void NotifyRenderCompleted()
             => (_component as IHandleAfterRender)?.OnAfterRender();
+
+        public void SetDirectParameters(ParameterCollection parameters)
+        {
+            // TODO: Attach tree parameters also
+            Component.SetParameters(parameters);
```
rynowak: I'm a little confused by this because it doesn't seem like tree parameters belong in `SetDirectParameters`

src/Microsoft.AspNetCore.Blazor/Components/Parameter.cs
```diff
@@ -11,31 +9,20 @@ namespace Microsoft.AspNetCore.Blazor.Components
...
-        public string Name
-            => _frames[_frameIndex].AttributeName;
+        public string Name { get; }
 
         /// <summary>
         /// Gets the value of the parameter.
         /// </summary>
-        public object Value
-            => _frames[_frameIndex].AttributeValue;
+        public object Value { get; }
```
rynowak: Now they are arguments tho

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
```diff
@@ -18,7 +18,7 @@ public static class ParameterCollectionExtensions
     {
         private const BindingFlags _bindablePropertyFlags = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.IgnoreCase;
 
-        private delegate void WriteParameterAction(ref RenderTreeFrame frame, object target);
+        private delegate void WriteParameterAction(object target, object parameterValue);
```
rynowak: The explicit delegate type isn't really needed now that this isn't `ref`

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
```diff
@@ -86,9 +86,9 @@ internal static IEnumerable<PropertyInfo> GetBindableProperties(Type targetType)
                         $"name '{propertyName.ToLowerInvariant()}'. Parameter names are case-insensitive and must be unique.");
                 }
 
-                result.Add(propertyName, (ref RenderTreeFrame frame, object target) =>
+                result.Add(propertyName, (object target, object parameterValue) =>
                 {
-                    propertySetter.SetValue(target, frame.AttributeValue);
+                    propertySetter.SetValue(target, parameterValue);
```
rynowak: You might want to consider filling in `PropertyHelper` here if these are cached.

src/Microsoft.AspNetCore.Blazor/Components/ParameterEnumerator.cs
```diff
@@ -28,9 +28,20 @@ internal ParameterEnumerator(RenderTreeFrame[] frames, int ownerIndex)
...
+                if (_currentIndex > _ownerIndex)
+                {
+                    ref var frame = ref _frames[_currentIndex];
+                    return new Parameter(frame.AttributeName, frame.AttributeValue);
+                }
+                else
+                {
+                    throw new InvalidOperationException("Iteration has not yet started.");
+                }
+            }
```
rynowak: At this point I'd suggest just doing this inside `MoveNext()` and not bothering to throw an exception - the behavior of accessing an enumerator before calling `MoveNext()` isn't defined, so returning a default value is totally acceptable.

rynowak:
Will be pausing for a bit - at this point I've gotten as far as 20c9f5a

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
SteveSandersonMS: Correct

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
SteveSandersonMS: > Is the idea to make it based on declared type rather than runtime type?

Yes. The key scenario I have in mind is if you're doing `<Provider Value="someObject">`, but `someObject` is sometimes `null`. If we did it based on runtime type, then the "overriding upstream values" logic would be very strange: it would override if it was non-null, but would not override if it was null. In fact there would be no possible way to override a non-null value with a null one. Given the way we identify the `Provider` just once per tree parameter usage, super bizarre things would happen whereby if the *initial* value was null, you wouldn't be able to observe it become non-null.

SteveSandersonMS: BTW it was really impressive how making the component generic just worked in every way I could think of (type inference, etc.).

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
SteveSandersonMS: Yes. I'm open to changing the design, but the idea was simply that if you choose to use "named slots", you're opting into that extra strictness on both the provider and consumer side.

My reasoning is that you'd only use the named slots if the value you're supplying is of a common type that might reasonably clash with other uses (e.g., `string`, `int`, `IPrincipal`, etc.). In that case you absolutely don't want your value showing up randomly in unrelated places just because they are asking for a `string`, so it's safer to force the consumer to declare the same named slot.

If the type you're supplying is unique enough (e.g., it's a special type you've defined purely for the purpose of cascading info to descendants), then you would have no reason for declaring a name too.

If it turns out I'm misjudging the use cases then we can certainly tweak the rules.

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
SteveSandersonMS: Honestly I'm not sure about it. My real preference would be for attribute/parameter names to be case sensitive too, but we went with insensitive because tag helpers are insensitive, which of course goes back to the history with it being a DSL for HTML (thanks, Berners-Lee!).

In this case I picked sensitive just because it's safer to be stricter, in that we can (more) safely loosen it in the future. But TBH you're right that the inconsistency is awkward. I'll change it to be case-insensitive.

src/Microsoft.AspNetCore.Blazor/Components/ProviderBase.cs
SteveSandersonMS: We make the best classes.

src/Microsoft.AspNetCore.Blazor/Components/ProviderBase.cs
SteveSandersonMS: You're right. I'll do that.

SteveSandersonMS:
> Would the provider value be an object that gets passed down as a whole or can we say we are interested in only a part of it.

It's a whole object. If you want multiple independent things, use multiple independent providers.

> Can this provider value be updated in descendants?

The flow is downwards. The expected pattern is you'd provide a callback that descendants want to call if they want their ancestor to do something (such as change the value being provided).

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
SteveSandersonMS: I have been using `ArgumentException` in the broader sense of considering component arguments to be arguments, even though they aren't CLR method args. Totally accept the point that if someone thinks `ArgumentException` precisely means CLR method arguments then this would be confusing.

If we want to change this later I'd be fine with that. Since you've said you don't think it's important, and I'm OK with it, I'll leave it as-is.

src/Microsoft.AspNetCore.Blazor/Components/ParameterAttribute.cs
SteveSandersonMS: It's now implemented as `CascadingParameterAttribute`, since we decided to use the term "cascading" for this whole feature area.

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
SteveSandersonMS: Changed to case-insensitive.

src/Microsoft.AspNetCore.Blazor/Components/ProviderBase.cs
SteveSandersonMS: As per other discussions, replaced with internal interface `ICascadingValueComponent`, so the XML doc is gone.

src/Microsoft.AspNetCore.Blazor/Components/ProviderBase.cs
SteveSandersonMS: Done

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
SteveSandersonMS: Good point. Done.

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
SteveSandersonMS: It is somewhat similar. I don't have any concrete reason to pull in the common implementation right now, but let me know if you think we should for some reason.

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
SteveSandersonMS: It was changed to `IReadOnlyList<>` in a subsequent commit.

dlr1:
> The flow is downwards. The expected pattern is you'd provide a callback that descendants can call if they want their ancestor to do something (such as change the value being provided).

I was asking about a provider at parent provides 'foo' as value, immediate child (child1) updates the value to 'bar'.  I want child of child1 to be able to say I want the value provided to it be from root parent not the immediate parent, In WPF we can say I want to bind to an ancestor of specific type.

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
SteveSandersonMS: Simplified slightly further by dropping the `Try...` pattern and just returning the `ICascadingValueComponent` or null, since there's no case where null can mean anything other than "there wasn't one".

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
SteveSandersonMS: I've just tried it, but TBH I don't prefer it. It does remove one null check from elsewhere, but only if you're reading enough of the code to realise why the value can't be null. It doesn't help with perf because it's replacing a null check with executing (effectively) an array length check which is presumably about the same cost. If we had non-nullable types it would be clearer, but since we don't I'll leave as-is unless you feel strongly (in which case I am happy to change it really).

test/Microsoft.AspNetCore.Blazor.Test/TreeParameterStateTest.cs
SteveSandersonMS: OK, done.

src/Microsoft.AspNetCore.Blazor/Components/Parameter.cs
SteveSandersonMS: Fair point. Rephrased in such a way as to still avoid using the word "argument" 😼

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
SteveSandersonMS: Do you mean we could replace its usages with `Action<object, object>`? If so yes, but personally I'd prefer to give it a custom name and custom parameter names for clarity. If you mean something else, or if there's some reason why it's practically better to use `Action<object, object>`, please let me know!

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
SteveSandersonMS: Happy to consider it. Since there's extra cost involved in figuring out what exactly that does, and whether it has extra generality that we don't need but has perf implications, and then just in making the code changes, do you happen to know already if there are any functionality differences? Any practical advantages or potential drawbacks?

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
SteveSandersonMS: Sounds good. Done.

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
rynowak: Yeah it's something to think about. Deciding that 'error-related-to-component-parameter' is a first class concept might be valuable in the future for diagnostics and logging.

src/Microsoft.AspNetCore.Blazor/Components/Provider.cs
rynowak: Yeah - I personally prefer that all identifiers in programming systems I use are case-sensitive. I feel like case-insensitive is something we only do at microsoft and I've never understood why it's valuable.

src/Microsoft.AspNetCore.Blazor/Components/TreeParameterState.cs
rynowak: I don't think it matters. I expect implementation details like this to require a detailed knowledge of what's going on.

test/Microsoft.AspNetCore.Blazor.Test/TreeParameterStateTest.cs
rynowak: > Yes, this is an invariant I'm relying on to make the updates cheap. If this were not the case, then whenever any Provider value changes, we have to reconsider every possible descendant consumer to see whether the new runtime type matches, not just the ones that are currently matched to its value.

OK then it makes sense that we need a declared type. We'll just have to call this out clearly. My experience with similar features in model binding is that developers generally expect any system they use to have the maximum theoretical flexibility - but I think we have good reasons to make this based on declared typed. Ultimately we provide value by sometimes designing systems that limit flexibility where it's important.

----

> Do you have examples of any scenarios where doing it based on declared type would be problematic/inconvenient for the developer?

No, I don't think it's problematic, since the binding to a type is static. 

I was trying to rationalize this based on systems like WPF's static resources, or `IPropertyOwner` from VS's editor APIs where the type is used as a key. 

The only case that would become problematic is if you wanted to bind values to a common framework type like `IEquatable` or `IConvertable`. I think our guidance would be "don't do that". For instance with this system, it would be possible to define a cascading parameter of type `IComponent` and then bind to the nearest ancestor that is a component.  These things are of somewhat minimal utility, and somewhat average level of wierdness. None of it feels broken.

What this does allow you to do is treat a value you provide as a feature collection, but exporting a bunch of common interfaces. You could image defining a control system based on capability, and exporting commands. Consider something similar to [UIA control patterns](https://docs.microsoft.com/en-us/dotnet/framework/ui-automation/ui-automation-control-patterns-overview) - now you have bubbling commands 😁 

It seems really powerful as it is defined.

src/Microsoft.AspNetCore.Blazor/Rendering/ComponentState.cs
rynowak: OK DOPE

src/Microsoft.AspNetCore.Blazor/Components/Parameter.cs
rynowak: Good. I'd like to avoid an argument about arguments.

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
rynowak: Nope, no difference, but pointing out in case you missed it, but I don't mind if you prefer it.

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollectionExtensions.cs
rynowak: After thinking about this more I think the part of property helper that's valuable is the algorithm for enumerating 'visible' properties.

I don't think the 'set property' logic is useful because I'm not sure it will work well in wasm. PropertyHelper is slightly faster on CoreCLR/NetFx.

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollection.cs
```diff
@@ -174,5 +174,27 @@ internal bool DefinitelyEquals(ParameterCollection oldParameters)
...
+            // describes the length of the buffer so that ParameterCollection
+            // knows how far to iterate through it.
+            var owner = RenderTreeFrame.PlaceholderChildComponentWithSubtreeLength(1 + numEntries);
+            builder.Append(owner);
+
+            if (numEntries > 0)
+            {
+                builder.Append(_frames, _ownerIndex + 1, numEntries);
+            }
+        }
```
rynowak: I know I'm not super familiar with this code, but I'm totally lost looking at this.... what's this doing and why?

src/Microsoft.AspNetCore.Blazor/Components/ChangeDetection.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                return true; // One's null and the other isn't, so different
+            }
+            else if (oldIsNotNull) // i.e., both are not null (considering previous check)
+            {
+                var oldValueType = oldValue.GetType();
+                var newValueType = newValue.GetType();
+                if (oldValueType != newValueType            // Definitely different
+                    || !IsKnownImmutableType(oldValueType)  // Maybe different
+                    || !oldValue.Equals(newValue))          // Somebody says they are different
```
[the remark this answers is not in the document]
SteveSandersonMS: It's because conceptually, the values we're comparing aren't really `oldValue` and `newValue`. The values we want to compare are *the state reachable transitively from `oldValue` at some unknown point in the past* versus *the state reachable transitively from `newValue` now*. The purpose of the check is to determine whether there's any chance that the deeply-reachable state is different now versus before.

If this was public API, we'd have to be much clearer about the semantics of this check and do some naming review to try to communicate this better. We could still change the name, but since it's `internal`, it's not as urgent.

With these semantics in mind, the only case where we can return `true` is for types known to be immutable. For, say, `object`, it's not enough for `oldValue.Equals(newValue)`, because in most cases `oldValue` and `newValue` will be handles to the same object instance but that doesn't tell us anything about whether it's mutated.

It's unfortunate that .NET doesn't have any native concept of (im)mutability. That limits us to hard-coding a set of commonly-used known-immutable types.

src/Microsoft.AspNetCore.Blazor/Components/ChangeDetection.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Blazor.Components
+{
+    internal class ChangeDetection
+    {
+        public static bool MayHaveChanged<T1, T2>(T1 oldValue, T2 newValue)
```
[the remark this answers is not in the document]
SteveSandersonMS: Yes, I considered doing this without the generics, but TBH I didn't have any reason not to use them. Even if we can't benefit from having the types statically right now, maybe we will later. AFAICT it's not going to make any difference either way right now.

I know it means JITting different versions of the method for each combination of value types, but that's a limited fixed cost overall for any given app so I didn't think it was a drawback to worry about.

Let me know if you have concerns as I'm happy to change this. For now will proceed as-is.

src/Microsoft.AspNetCore.Blazor/Components/ParameterCollection.cs
SteveSandersonMS: Yeah, I'm not super thrilled with this, but it's creating a clone of the parameter collection data that matches the shape of what `ParameterCollectionEnumerator` knows how to step through.

I previously tried making a new data structure but that forced either having two versions of `ParameterCollectionEnumerator` that understands each of them, or actually copying the parameter data to a new buffer even in the non-cascading case. Both seemed worse than this relatively contained bit of copying here.

SteveSandersonMS:
@dotnet-bot test ci please

PR #1560, opened 2018-10-13
Add ability to override blazor config

csnewman:
Adds the ability within the BlazorOptions to set a custom BlazorConfig, if none is provided, the standard .blazor.config file is loaded.

Quite useful when using customised development environments. I remotely start and debug the blazor server straight from visual studio on a different machine using ssh what means the paths need to be set at runtime instead of being read from the config file.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/BlazorConfig.cs
@@ -7,45 +7,77 @@ using System.Linq;
 
 namespace Microsoft.AspNetCore.Blazor.Server
 {
-    internal class BlazorConfig
+    /// <summary>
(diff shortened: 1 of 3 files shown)


csnewman:
I'm basically using the following setup: https://github.com/Microsoft/MIEngine/wiki/Offroad-Debugging-of-.NET-Core-on-Linux---OSX-from-Visual-Studio

The main issue is that the local paths do not lineup with the remote maths in the .blazor.config file. I currently set the correct paths as environment variables and then overwrite the config values at runtime. (the .blazor.config file needs to stay as is, as the folders are shared and that would cause issues if modified to have the correct path at compile time for example).

danroth27:
We appreciate the contribution, but we've decided that adding extensibility in this area is not something we want to do at this point. We're open to reconsidering this if we hear more feedback in this area.

PR #1563, opened 2018-10-15
Fix grammatical errors

amitmerchant1990:
Hi,

I've fixed couple of grammatical errors in the readme. Kindly please check the same and let me know if we can merge those or not.

Thanks!


SteveSandersonMS:
Thanks for the suggestion! I think the content is fine as-is without the extra commas. Hope that's OK!

PR #1564, opened 2018-10-15
Add Cascading flag to Parameter

SteveSandersonMS:
... so that components that pass through an arbitrary set of parameters can choose to filter out cascading ones if they want.

This is a follow-up to #1545.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/Parameter.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/Parameter.cs
@@ -23,13 +23,13 @@ namespace Microsoft.AspNetCore.Blazor.Components
         /// Gets a value to indicate whether the parameter is cascading, meaning that it
         /// was supplied by a <see cref="CascadingValue{T}"/>.
         /// </summary>
-        public bool IsCascading { get; }
+        public bool Cascading { get; }
 
-        internal Parameter(string name, object value, bool isCascading)
+        internal Parameter(string name, object value, bool cascading)
         {
             Name = name;
             Value = value;
-            IsCascading = isCascading;
+            Cascading = cascading;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Blazor/Components/Parameter.cs
```diff
@@ -19,10 +19,17 @@ namespace Microsoft.AspNetCore.Blazor.Components
         /// </summary>
         public object Value { get; }
 
-        internal Parameter(string name, object value)
+        /// <summary>
+        /// Gets a value to indicate whether the parameter is cascading, meaning that it
+        /// was supplied by a <see cref="CascadingValue{T}"/>.
+        /// </summary>
+        public bool IsCascading { get; }
```
SteveSandersonMS: Is there guidance on whether we're meant to prefix names of flags like this with `Is` or not?

To mean it feels more natural to have the "is" prefix when you're reading the value, but more natural not to have it when you're writing the value (perhaps because "is" implies past tense, as in, that's already how it is and you can't change it). Ultimately I'm fine with either.

SteveSandersonMS: Ah forget that. My sense is that we don't usually have the "is" prefix so I've removed it.

danroth27: Guidelines are [here](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-properties), but they're a bit vague: "DO name Boolean properties with an affirmative phrase (CanSeek instead of CantSeek). Optionally, you can also prefix Boolean properties with "Is," "Can," or "Has," but only where it adds value."

PR #1566, opened 2018-10-15
"Fixed" mode and E2E tests for <CascadingValue>

SteveSandersonMS:
This is a follow-up to #1545.

It's a perf optimization that allows the framework to skip the subscription tracking and parameter snapshotting that it normally has to do when cascading parameters might change value. The intended use case is for things like "form context" or anywhere you do:

    <CascadingValue Value=this>...</CascadingValue>

... because if the `Value` instance is constant for the lifetime of the `<CascadingValue>`, we can make this really cheap.

If developers fail to apply this optimization when they could, nothing bad happens. We only expect this optimization to be used by our framework components and more advanced component authors who have proactively read docs about this.

**Also, E2E tests**

It was convenient to chain the E2E tests for cascading parameter/value in general to this PR. As usual, the E2E tests are not meant to cover every possible combination of use cases and error scenarios (that's in the unit tests); the E2E tests are a representative sample of high-level common scenarios.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/CascadingValue.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/CascadingValue.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNetCore.Blazor.Components
     {
         private RenderHandle _renderHandle;
         private HashSet<ComponentState> _subscribers; // Lazily instantiated
+        private bool _hasSetParametersPreviously;
 
         /// <summary>
         /// The content to which the value should be provided.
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNetCore.Blazor/Components/CascadingValue.cs
```diff
@@ -35,8 +36,18 @@ public class CascadingValue<T> : ICascadingValueComponent, IComponent
         /// </summary>
         [Parameter] private string Name { get; set; }
 
+        /// <summary>
+        /// If true, indicates that <see cref="Value"/> will not change. This is a
+        /// performance optimization that allows the framework to skip setting up
+        /// change notifications. Set this flag only if you will not change
+        /// <see cref="Value"/> during the component's lifetime.
+        /// </summary>
+        [Parameter] private bool Fixed { get; set; }
```
SteveSandersonMS: My other idea for the name was `FixedValue` (or even `IsFixed`). It's a bit clearer, but there's a lot of repetition in the word `Value` if you have:

    <CascadingValue Value=something FixedValue=true>

@danroth27 @rynowak Got any preference? I'm happy either way.

kant2002: If you mind, let me drop my 2 cents here.
Value `IsValueFixed` seems to be best of words and very Microsofty in the name.
Another word which is used in UI word is `Frozen` so maybe `IsValueFrozen`. I rarely seen Fixed value in the context of UI frameworks.

danroth27: My vote is for `IsFixed`

SteveSandersonMS: OK, it is fixed to `IsFixed`

PR #1574, opened 2018-10-16
When tracing dependencies *without* linking, don't fail the build if there are unresolved dependencies

SteveSandersonMS:
There are legitimate use cases for referencing BCL assemblies that *aren't* in the Mono WebAssembly BCL, particularly for Blazor-on-the-server which runs on CoreCLR (and hence supports a broader BCL) anyway.

Developers can already do this themselves if they are using the linker (by passing `--skip-unresolved true` to the linker). But before this PR there was no way to get around the problem if you're not using the linker, which you typically aren't for server-side Blazor.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Build/Core/RuntimeDependenciesResolver.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Build/Core/RuntimeDependenciesResolver.cs
@@ -76,7 +76,7 @@ namespace Microsoft.AspNetCore.Blazor.Build
 
             internal void ResolveAssemblies()
             {
-                var visitedAssemblies = new Dictionary<string, AssemblyEntry>();
+                var visitedAssemblies = new HashSet<string>();
                 var pendingAssemblies = new Stack<AssemblyNameReference>();
                 pendingAssemblies.Push(Assembly.Definition.Name);
                 ResolveAssembliesCore();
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Merging now because I need a build to be on MyGet for some other work I'm doing, and it's pretty safe. @rynowak please comment if you have any concerns.

src/Microsoft.AspNetCore.Blazor.Build/Core/RuntimeDependenciesResolver.cs
```diff
@@ -85,15 +85,24 @@ void ResolveAssembliesCore()
                 {
                     while (pendingAssemblies.TryPop(out var current))
                     {
-                        if (!visitedAssemblies.ContainsKey(current.Name))
+                        if (!visitedAssemblies.Contains(current.Name))
                         {
+                            visitedAssemblies.Add(current.Name);
```
rynowak: You can just use `.Add` inside the if you want #RyanPoints

src/Microsoft.AspNetCore.Blazor.Build/Core/RuntimeDependenciesResolver.cs
```diff
@@ -85,15 +85,24 @@ void ResolveAssembliesCore()
...
-                            foreach (var reference in references)
+                            if (resolved != null)
                             {
-                                pendingAssemblies.Push(reference);
+                                Results.Add(resolved);
+                                var references = GetAssemblyReferences(resolved);
+                                foreach (var reference in references)
+                                {
+                                    pendingAssemblies.Push(reference);
+                                }
```
rynowak: Is it worth logging/reporting these as warnings? My experience with MVC is that being intolerant of missing assemblies has prevented a lot of pain for the team (reflection loading/runtime compilation). I understand in this case it's a valid use case, but this will make it easier for mistakes to creep in.

Issue #1530, opened 2018-10-06
An error while parsing .cshtml


panda-eye:
I have a line in Teachers.cshtml:
`<button onclick="window.location.href = '/teacher/@(teacher.Id)/';">More info...</button>`
After complation I have the following line in Teachers.g.cs (/obj/Debug/netstandart2.0/Pages/):
`builder.AddAttribute(24, "onclick", (Microsoft.AspNetCore.Blazor.Components.BindMethods.GetEventHandlerValue<Microsoft.AspNetCore.Blazor.UIMouseEventArgs>() + "window.location.href" + " =" + " \'/teacher/" + (teacher.Id) + "/\';" + ()));`
And an error message:

> CS1525	Invalid expression term ")"

panda-eye:
The same result if I remove whitespaces

codeinstinct:
Here's what I would do:

Change this
`<button onclick="window.location.href = '/teacher/@(teacher.Id)/';">More info...</button>`

to
`<button onclick=@Navigate(teacher.Id)>More info...</button>`

Add this line at the top of the .cshtml
`@inject Microsoft.AspNetCore.Blazor.Services.IUriHelper UriHelper`

Add this method `Navigate` to the functions block
`
@functions {
    private void Navigate(string id)
    {
        UriHelper.NavigateTo($"/teacher/{id}");
    }
}`

adrian109:
Believe this issue might be a similar issue with double and single quote combinations not being correctly escaped as raised in issue #1527

PR #1577, opened 2018-10-17
escape quotes in event handler attribute content

### LINKED ISSUE #1527: Blazor Codegen does not escape quotes
If you use double quotes inside single quotes in attributes, this will result in broken codegen due to unescaped double quotes inside double quotes.
Input:
```html
<input onfocus='alert("Test");' />
```
Output:
```csharp
builder.OpenElement(0, "input");
builder.AddAttribute(1, "onfocus", Microsoft.AspNetCore.Blazor.Components.BindMethods.GetEventHandlerValue<Microsoft.AspNetCore.Blazor.UIFocusEventArgs>("alert("Test");"));
builder.CloseElement();
```

BickelLukas:
Fix broken codegen due to double quotes not escaped inside single quoted event handlers.
fixes #1527


adrian109:
Does this also fix #1530?

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
```diff
@@ -2443,6 +2443,20 @@ public void OnComponentHover(UIMouseEventArgs e)
             CompileToAssembly(generated);
         }
 
+        [Fact]
+        public void EventHandlerTagHelper_EscapeQuotes()
+        {
+            // Act
+            var generated = CompileToCSharp(@"
+<input onfocus='alert(""Test"");' />
```
adrian109: I see that you have added a test for the following:
`<input onfocus='alert("Test");' />`

What about the swapping the single and double quotes around?  The following format would be my preferred way of writing the same input element:
`<input onfocus="alert('Test');" />`

Maybe worth adding another test to cover both formats if they are not already covered.

BickelLukas:
@adrian109 no it does not fix #1530. That error is not because of quotes not being escaped. I started doing some investigation but didn't have enough time to find a fix yet. Maybe i'll take a look at it in the following days

test/Microsoft.AspNetCore.Blazor.Build.Test/CodeGenerationTestBase.cs
BickelLukas: I can also add a test case for that but it wouldn't serve much purpose since a single quote inside double quotes is valid c# code. Hence the single quotes don't need to be escaped at all.

rynowak:
Looks great!

PR #1579, opened 2018-10-17
Remove useless call to StateHasChanged

RemiBou:
This removes an unecessary call to StateHasChanged if task is not finished when we reach the "if" line 172

or we could remove StateHasChanged from ContinueAfterLifecycleTask as it's already called line 178 and 212

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/BlazorComponent.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Blazor.RenderTree;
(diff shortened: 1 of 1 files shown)


Suchiman:
@RemiBou both calls are necessary. The StateHasChanged Invokation on 177 / 178 will refresh the component after the synchronous event handlers (OnInit / OnParametersSet) were called or the async one's either completed synchronously or were not executed. The StateHasChanged from ContinueAfterLifecycleTask will be fired after asynchronous work completes. All in all, this boils down to the little bit weird behavior when having for example https://github.com/aspnet/Blazor/blob/d8d368c066061230d40d584ff49d5d21b282e643/src/Microsoft.AspNetCore.Blazor.Templates/content/BlazorStandalone-CSharp/Pages/FetchData.cshtml#L40-L43
where StateHasChanged will be fired once when method yields due to await, and then a second time after it got the result and truly completed.

RemiBou:
The "else" prevents the first call to StateHasChanged in this case doesn't it ? Do we need to call it when the method yields instead of only when it completes ?

Suchiman:
Yes'ish, the else only prevents it for the `OnParametersSetAsync`, not for the `OnInitAsync`.
Also if you were to override both `OnParametersSet` and `OnParametersSetAsync`, StateHasChanged won't be called for the synchronous version, but this is an odd design in the first place.

danroth27:
We think the current code is ok as is.

PR #1584, opened 2018-10-18
Return completed task rather than null for empty virtual async methods.

hutchcodes:
Also changed the OnAfterRenderAsync task to only ContinueWith if the task has not RunToCompletion so that it won't run on the completed task. OnInitAsync and OnParameterSetAsync were already handling this.


hutchcodes:
It looks like the AppVeyor CI build failed for transient reasons not related to my code changes.

> Failures
>  - googlechrome (exited 1) - googlechrome not installed. An error occurred during installation:
>  The remote server returned an error: (503) Server Unavailable. Service Unavailable

Anything I need to do here?

hutchcodes:
Closing because the build failed for what appears to be transient reasons. I'll reopen a new PR.

PR #1591, opened 2018-10-19
Add webpack to prereqs on readme page for the less initiated.

eeevans:
Was looking to contribute and ran into missing prerequisite webpack.  That was the only other thing needed for building.


SteveSandersonMS:
We don't require a global install of Webpack. The build takes care of installing and using a local copy.

I'm not certain why it seemed necessary in your case - perhaps there is something unusual about your Node setup.

Closing - hope that's OK.

PR #1595, opened 2018-10-21
Add routing enhancements

csnewman:
Attempts to add some functionality from #293 plus some other features/bug fixes.

- Makes router extensible (Many aspects are now virtual methods)
- Adds 404 handling via virtual method, defaults to a fallback route parameter
- Adds error handling via virtual method, defaults to a error route parameter
- Cache components per assembly
- Allow adding and removing of assemblies from router
- Makes route tables "routes" get only
- Enables route inheritance (from component base classes)
- Fixes component assembly searching (Now checks to see whether blazor is indirectly referenced)
- Step closer to nested routers (BaseURI is now a parameter, and you have a lot more control over which assemblies are loaded)
- Step closer to late/lazy loaded modules (You can add and remove assemblies from routers at runtime)

One important change was that RouteTables cannot be cached per assembly, as the order of the routes inside the table specifies the matching order, meaning instead the routes must be cached (route collection) and then merged and ordered. 

At the moment the unit tests etc have not been updated, only the core logic just to see what you think.

In the future if an error occurs elsewhere inside the application, it might be an idea to hook into the "handleerror" method to route to an error page. For now it just handles routing errors.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Components/ComponentResolver.cs
@@ -11,8 +11,11 @@ namespace Microsoft.AspNetCore.Blazor.Components
     /// <summary>
     /// Resolves components for an application.
     /// </summary>
-    internal static class ComponentResolver
+    public static class ComponentResolver
     {
+        private static readonly IDictionary<Assembly, IList<Type>> _componentCache =
+            new Dictionary<Assembly, IList<Type>>(new AssemblyComparer());
+
         /// <summary>
         /// Lists all the types 
         /// </summary>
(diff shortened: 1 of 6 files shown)


SteveSandersonMS:
Thanks very much for this, @csnewman! I appreciate this must have been a lot of work, and you have gone pretty deep with routing enhancements here.

> At the moment the unit tests etc have not been updated, only the core logic just to see what you think.

It's good to provide a preview of what you're implementing, so thanks for that.

As it is, I wouldn't be able to review or merge this PR because it contains too many changes of behavior all mixed together. Rather than having 10 new bugfixes and features combined, we'd really each change as a separate PR with its own tests.

Some of the changes proposed here extend the API surface in ways that are potentially expensive or undesirable, because it's not clear that the demand from developers is high enough to warrant the long-term back-compatibility costs. This is especially true for "make X more extensible"-type changes - sometimes those are needed, but the necessity has to reach a high bar.

To proceed, I'd recommend posting an issue on https://github.com/aspnet/aspnetcore for each change you want to propose, describing what the change is and any pros/cons. We can then discuss the design and if applicable it can be followed up with a PR.

Is that OK? I'm sorry for the extra time it takes to go though each change in detail like that, but it's the only way we can maintain the framework! Hope you don't mind.

PR #1599, opened 2018-10-23
Fix NullReferenceException when href is null

Gaulomatic:
Fix `NavLink` when the given href is null (or an empty string in the markup).

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
+++ b/src/Microsoft.AspNetCore.Blazor/Routing/NavLink.cs
@@ -71,7 +71,7 @@ namespace Microsoft.AspNetCore.Blazor.Routing
 
             // Update computed state and render
             _hrefAbsolute = href == null ? null : UriHelper.ToAbsoluteUri(href).AbsoluteUri;
-            _isActive = ShouldMatch(UriHelper.GetAbsoluteUri());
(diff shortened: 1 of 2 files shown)


danroth27:
Looks good, but could you please also add a test for this?

Gaulomatic:
The `ActiveClass` has always been applied when href was not defined or simply empty. The change circumwents the whole determination logic behind that when the href is null or empty. Thus the NullReferenceException is not being thrown and subsequently the `_isActive` field is always false and CSS is applied like I would it expect it to. The tests are checking for that.

SteveSandersonMS:
I don't think the logic in this PR is correct. For example:

```csharp
_isActive = !string.IsNullOrEmpty(href) && this.ShouldMatch(this.UriHelper.GetAbsoluteUri());
```

This would prevent the link from ever being active if `href` was an empty string. But empty string is a perfectly valid relative URI and should be active if the user is at that location (i.e., they are at the base URI). The existing implementation handles this scenario correctly, but the PR would break it.

> There is one case left: Currently the home Uri "/" is always active. Question is, if this bevavior is wanted. Logically it makes sense, but in practice it's not what you usually want.

The developer gets to make the choice by specifying the `Match` parameter on `NavLink`.

I'll close this because I think the existing behavior is already correct, but if you want to follow up with any changes to this PR, would you please be able to submit them at https://github.com/aspnet/aspnetcore? Thanks!

PR #1616, opened 2018-10-26
Asp

VictoriaLasso:
asp

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Blazor
 
-**An experimental .NET web framework using C#/Razor and HTML that runs in the browser via WebAssembly**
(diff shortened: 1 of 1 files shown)


Andrzej-W:
Although Blazor can be used in/with ASP.NET application it is not required (see: https://blazor.net/docs/host-and-deploy/index.html#standalone-deployment). I'm not sure this change is necessary.

danroth27:
Thanks for this contribution! We prefer to refer to .NET in this case because ASP.NET is typically associated with using .NET on the server. We don't want a bunch of .NET devs to tune out because the expect that ASP.NET is server-side only. I hope that's ok!

PR #1620, opened 2018-10-27
Returns completed tasks (not null) from BlazorComponent.OnInitAsync

duracellko:
BlazorComponent methods `OnInitAsync` and `OnParametersSetAsync` should return Task.Completed and not null.

For example following code will fail with NullReferenceException. And there is FxCop rule to run base methods.
```csharp
    {
        await base.OnInitAsync();
        forecasts = await Http.GetJsonAsync<WeatherForecast[]>("sample-data/weather.json");
    }
```

This PR has following changes:
- `BlazorComponent` methods `OnInitAsync` and `OnParametersSetAsync` return `Task.Completed`.
- Replaces ContinueWith with async/await in `SetParameters`.
- Add few unit tests for BlazorComponent to cover cases above.

### DIFF
--- a/src/Microsoft.AspNetCore.Components/ComponentBase.cs
+++ b/src/Microsoft.AspNetCore.Components/ComponentBase.cs
@@ -211,12 +211,7 @@ namespace Microsoft.AspNetCore.Components
             // at the end of every event callback.
             StateHasChanged();
 
-            if (task.Status == TaskStatus.RanToCompletion)
-            {
-                return;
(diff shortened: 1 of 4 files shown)


duracellko:
Oh, I just noticed, there is similar PR #1619

SteveSandersonMS:
Thanks for this! It's now moved to https://github.com/aspnet/AspNetCore/pull/4790, where it will be merged shortly.

PR #1647, opened 2018-11-07
Update to latest Mono (for 0.7.0 release)

SteveSandersonMS:
As of this PR, we're now using Mono's prebuilt binaries instead of building custom ones for Blazor. This greatly simplifies the process of changing Mono versions.

This PR also drops the historical support for asm.js fallback. We're making this choice because better alternatives have now emerged for all the primary asm.js fallback scenarios:

* For Safari, it now supports WebAssembly properly so there's no remaining need for asm.js fallback
* For old browsers such as IE11, the best solution is server-side Blazor. This avoids putting heavy load on the browser, meaning you can get excellent compatibility and perf right back to really ancient versions of IE (given suitable JS polyfills). In the long run it wouldn't be realistic to rely on asm.js fallback because the JS engine in these older browsers typically wouldn't supply acceptable perf.

The reason we're making this change (removing asm.js fallback) now is that it's a necessary part of switching to use prebuilt Mono binaries (they don't ship asm.js versions).

**Debugging improvements**

This PR also updates to a newer version of @kumpera's debug proxy. You can now remove breakpoints and step in/over properly 👍


SteveSandersonMS:
Merging now because @rynowak is not around to review at the moment, the changes are pretty straightforwards, and we want to be ready for the 0.7.0 release soon.

But @rynowak please feel free to review this when you are back if you wish. I'm happy to make any follow-up changes.

Suchiman:
@SteveSandersonMS Maybe also update https://github.com/aspnet/Blazor/blob/22d3709414897af69bbb0d00bd9e3389c1096ddd/src/Microsoft.AspNetCore.Blazor.Server/Microsoft.AspNetCore.Blazor.Server.csproj#L21 to `0.10.0` like in https://github.com/mono/mono/blob/5eb4698bb23f9fd5226ae641abd2708b79c43c0b/sdks/wasm/Mono.WebAssembly.DebuggerProxy/Mono.WebAssembly.DebuggerProxy.csproj#L8 ?

SteveSandersonMS:
@Suchiman We're using 0.10.0-beta7 in a few places in the repo, so I'd prefer to keep it consistent with those until we update all of them (which we could do now, but I'm not aware of any problem it would solve).

IyyappanAmirthalingam:
@SteveSandersonMS 

In think the document should be updated accordingly for the following line

> " It works in older browsers too by falling back to an asm.js based .NET runtime."

PR #1651, opened 2018-11-08
Route events based on delegate target, not component ID

SteveSandersonMS:
This fixes multiple issues and potential areas of confusion related to event routing (in effect, the scenarios where you have to call `StateHasChanged` or not).

Previously, if one component passed an event handler (e.g., an `Action`, `Func<Task>`, `MulticastDelegate`, etc.) through to another component, either as a `[Parameter]` or simply by including it in some `ChildContent`, then when the event was fired it would be routed not to the component that owns the event handler, but the one that rendered it into the tree.

### Example

MyButton.cshtml:

```
<button onclick=@OnClick>Click me<button>
@functions {
    [Parameter] Action OnClick { get; set; }
}
```

Consumer:

```
<MyButton OnClick=@SomeMethod /> or <MyButton OnClick=@(() => { clickCount++; }) />

Clicks: @clickCount

@functions {
    int clickCount = 0;

    void SomeMethod()
    {
        clickCount++;
    }
}
```

You'd expect that clicking the button would update the click count, but it wouldn't. You'd have to fix it by calling `StateHasChanged` manually in the consumer's `SomeMethod` or the lambda. This was annoying and confusing.

### Example 2:

PassthroughComponent.cshtml:

```
@ChildContent

@functions {
    [Parameter] RenderFragment ChildContent { get; set; }
}
```

Consumer.cshtml:

```
<PassthroughComponent>
    <button OnClick=@(() => { clickCount++; })>Click me</button>
</PassthroughComponent>

Clicks: @clickCount

@functions {
    int clickCount = 0;
}
```

Again, clicking the button would seem to have no effect, unless you manually put `StateHasChanged` into the lambda.

### Fix

We now route events to the component matching `eventHandlerDelegate.Target`, not the component that inserted the delegate into the render tree. This fixes both of the common cases above, and works with event handlers expressed as both methods and lambdas.

This also simplifies cases around binding. Components that expose custom bindables can now trigger updates in their consumers easily.

### New API

In the case where a custom component wants to trigger a supplied event handler delegate indirectly (e.g., inside one of its own methods), there's now a public static API:

```
ComponentEvents.InvokeEventHandler(someDelegate, eventArgs)
```

This works with delegates that are `Action`, `Action<T>`, `Func<Task>`, `Func<T, Task>` (the fast paths), or arbitrary `MulticastDelgate` (slower path). It automatically causes the recipient to trigger its own re-rendering logic, including responses to async tasks (i.e., if it's a `Func<Task>`, then the recipient renders once synchronously, then again asynchronously after the task, just like normal event handlers).

### Breaking change

There's one scenario where this may break something people were already doing, i.e., relying on the older strange behavior. Example:

```
<button onclick=@SomeOtherObject.SomeMethod>Click me</button>
```

Previously, this would run `SomeOtherObject.SomeMethod` and then re-render the component. That might be useful if this method updated some global state that you read back during rendering.

Now it will no longer re-render the component, because `SomeMethod` isn't on that component. If you actually do want to re-render the component in this case, the solution is to ensure you're actually calling one of its methods or lambdas, e.g.:

```
<button onclick=@(() => SomeOtherObject.SomeMethod())>Click me</button>
```

Alternatively, if you're actually updating some global state container, it's better still to have the state container raise some event that triggers all the necessary UI updates on all components, not just the one that raised the action.

Although it's a slight drawback to have this breaking change, I still think this is the right thing to do. The new rules about when events update components are much simpler and easier to reason about (i.e., we re-render the component that owns the event handler method/lambda, regardless of who passed it where).

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
+++ b/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
@@ -1,6 +1,4 @@
-import { System_Array, MethodHandle } from '../Platform/Platform';
-import { RenderBatch, ArraySegment, ArrayRange, RenderTreeEdit, RenderTreeFrame, EditType, FrameType, ArrayValues } from './RenderBatch/RenderBatch';
-import { platform } from '../Environment';
+import { RenderBatch, ArraySegment, RenderTreeEdit, RenderTreeFrame, EditType, FrameType, ArrayValues } from './RenderBatch/RenderBatch';
 import { EventDelegator } from './EventDelegator';
 import { EventForDotNet, UIEventArgs } from './EventForDotNet';
 import { LogicalElement, toLogicalElement, insertLogicalChild, removeLogicalChild, getLogicalParent, getLogicalChild, createAndInsertLogicalContainer, isSvgElement } from './LogicalElements';
@@ -9,16 +7,14 @@ const selectValuePropname = '_blazorSelectValue';
 const sharedTemplateElemForParsing = document.createElement('template');
 const sharedSvgElemForParsing = document.createElementNS('http://www.w3.org/2000/svg', 'g');
 const preventDefaultEvents: { [eventType: string]: boolean } = { submit: true };
-let raiseEventMethod: MethodHandle;
-let renderComponentMethod: MethodHandle;
 
 export class BrowserRenderer {
   private eventDelegator: EventDelegator;
   private childComponentLocations: { [componentId: number]: LogicalElement } = {};
 
   constructor(private browserRendererId: number) {
-    this.eventDelegator = new EventDelegator((event, componentId, eventHandlerId, eventArgs) => {
-      raiseEvent(event, this.browserRendererId, componentId, eventHandlerId, eventArgs);
+    this.eventDelegator = new EventDelegator((event, eventHandlerId, eventArgs) => {
+      raiseEvent(event, this.browserRendererId, eventHandlerId, eventArgs);
     });
   }
 
@@ -32,7 +28,7 @@ export class BrowserRenderer {
       throw new Error(`No element is currently associated with component ${componentId}`);
     }
 
-    this.applyEdits(batch, componentId, element, 0, edits, referenceFrames);
+    this.applyEdits(batch, element, 0, edits, referenceFrames);
   }
 
   public disposeComponent(componentId: number) {
@@ -47,7 +43,7 @@ export class BrowserRenderer {
     this.childComponentLocations[componentId] = element;
   }
 
-  private applyEdits(batch: RenderBatch, componentId: number, parent: LogicalElement, childIndex: number, edits: ArraySegment<RenderTreeEdit>, referenceFrames: ArrayValues<RenderTreeFrame>) {
+  private applyEdits(batch: RenderBatch, parent: LogicalElement, childIndex: number, edits: ArraySegment<RenderTreeEdit>, referenceFrames: ArrayValues<RenderTreeFrame>) {
     let currentDepth = 0;
     let childIndexAtCurrentDepth = childIndex;
 
@@ -67,7 +63,7 @@ export class BrowserRenderer {
           const frameIndex = editReader.newTreeIndex(edit);
           const frame = batch.referenceFramesEntry(referenceFrames, frameIndex);
           const siblingIndex = editReader.siblingIndex(edit);
-          this.insertFrame(batch, componentId, parent, childIndexAtCurrentDepth + siblingIndex, referenceFrames, frame, frameIndex);
+          this.insertFrame(batch, parent, childIndexAtCurrentDepth + siblingIndex, referenceFrames, frame, frameIndex);
           break;
         }
         case EditType.removeFrame: {
@@ -81,7 +77,7 @@ export class BrowserRenderer {
           const siblingIndex = editReader.siblingIndex(edit);
           const element = getLogicalChild(parent, childIndexAtCurrentDepth + siblingIndex);
           if (element instanceof Element) {
-            this.applyAttribute(batch, componentId, element, frame);
+            this.applyAttribute(batch, element, frame);
           } else {
             throw new Error(`Cannot set attribute on non-element child`);
           }
@@ -145,12 +141,12 @@ export class BrowserRenderer {
     }
   }
 
-  private insertFrame(batch: RenderBatch, componentId: number, parent: LogicalElement, childIndex: number, frames: ArrayValues<RenderTreeFrame>, frame: RenderTreeFrame, frameIndex: number): number {
+  private insertFrame(batch: RenderBatch, parent: LogicalElement, childIndex: number, frames: ArrayValues<RenderTreeFrame>, frame: RenderTreeFrame, frameIndex: number): number {
     const frameReader = batch.frameReader;
     const frameType = frameReader.frameType(frame);
     switch (frameType) {
       case FrameType.element:
-        this.insertElement(batch, componentId, parent, childIndex, frames, frame, frameIndex);
+        this.insertElement(batch, parent, childIndex, frames, frame, frameIndex);
         return 1;
       case FrameType.text:
         this.insertText(batch, parent, childIndex, frame);
@@ -161,7 +157,7 @@ export class BrowserRenderer {
         this.insertComponent(batch, parent, childIndex, frame);
         return 1;
       case FrameType.region:
-        return this.insertFrameRange(batch, componentId, parent, childIndex, frames, frameIndex + 1, frameIndex + frameReader.subtreeLength(frame));
+        return this.insertFrameRange(batch, parent, childIndex, frames, frameIndex + 1, frameIndex + frameReader.subtreeLength(frame));
       case FrameType.elementReferenceCapture:
         if (parent instanceof Element) {
           applyCaptureIdToElement(parent, frameReader.elementReferenceCaptureId(frame)!);
@@ -178,7 +174,7 @@ export class BrowserRenderer {
     }
   }
 
-  private insertElement(batch: RenderBatch, componentId: number, parent: LogicalElement, childIndex: number, frames: ArrayValues<RenderTreeFrame>, frame: RenderTreeFrame, frameIndex: number) {
+  private insertElement(batch: RenderBatch, parent: LogicalElement, childIndex: number, frames: ArrayValues<RenderTreeFrame>, frame: RenderTreeFrame, frameIndex: number) {
     const frameReader = batch.frameReader;
     const tagName = frameReader.elementName(frame)!;
     const newDomElementRaw = tagName === 'svg' || isSvgElement(parent) ?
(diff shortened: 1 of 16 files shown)


src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
```diff
@@ -1,6 +1,4 @@
-import { System_Array, MethodHandle } from '../Platform/Platform';
```
rynowak: I'm assuming these are removed as unused (same with `platform`)

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
```diff
@@ -355,14 +350,13 @@ function countDescendantFrames(batch: RenderBatch, frame: RenderTreeFrame): numb
...
 
-function raiseEvent(event: Event, browserRendererId: number, componentId: number, eventHandlerId: number, eventArgs: EventForDotNet<UIEventArgs>) {
+function raiseEvent(event: Event, browserRendererId: number, eventHandlerId: number, eventArgs: EventForDotNet<UIEventArgs>) {
   if (preventDefaultEvents[event.type]) {
     event.preventDefault();
   }
 
   const eventDescriptor = {
     browserRendererId,
-    componentId,
```
rynowak: I'm assuming that a lot of this got simpler because we no longer track the componentId associated with an eventHandler on the JS side. If this isn't right let me know 😆

src/Microsoft.AspNetCore.Blazor/Components/EventHandlerInvoker.cs
```diff
@@ -7,49 +7,53 @@
...
                 case MulticastDelegate @delegate:
-                    return @delegate.DynamicInvoke(e) as Task ?? Task.CompletedTask;
+                    return @delegate.DynamicInvoke(eventArg) as Task ?? Task.CompletedTask;
 
                 case null:
                     return Task.CompletedTask;
             }
         }
+
+        internal bool CheckDelegateTarget(object target)
```
rynowak: IsSameTarget? I would expect a method named like this to throw

src/Microsoft.AspNetCore.Blazor/Components/IHandleEvent.cs
```diff
@@ -11,8 +13,8 @@ public interface IHandleEvent
         /// <summary>
         /// Notifies the component that one of its event handlers has been triggered.
         /// </summary>
-        /// <param name="binding">The event binding.</param>
-        /// <param name="args">Arguments for the event handler.</param>
-        void HandleEvent(EventHandlerInvoker binding, UIEventArgs args);
+        /// <param name="invoker">The event invoker.</param>
+        /// <param name="eventArgs">Arguments for the event handler.</param>
+        Task HandleEvent<TArg>(EventHandlerInvoker<TArg> invoker, TArg eventArgs);
```
rynowak: HandleEventAsync? - have we been using async as a suffix on lifecycle methods?

src/Microsoft.AspNetCore.Blazor/Rendering/ComponentState.cs
```diff
@@ -95,20 +95,6 @@ public void DisposeInBatch(RenderBatchBuilder batchBuilder)
...
 
-        public void DispatchEvent(EventHandlerInvoker binding, UIEventArgs eventArgs)
-        {
-            if (_component is IHandleEvent handleEventComponent)
-            {
-                handleEventComponent.HandleEvent(binding, eventArgs);
-            }
-            else
-            {
-                throw new InvalidOperationException(
```
rynowak: I'm guessing that this validation is lost now that we've generalized this. I don't have a big concern about it because I haven't seen anyone complain about this error message.

src/Microsoft.AspNetCore.Blazor/Components/EventHandlerInvoker.cs
SteveSandersonMS: That's reasonable. I'll change it.

test/Microsoft.AspNetCore.Blazor.Test/ComponentEventsTest.cs
```diff
@@ -0,0 +1,184 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Blazor.Components;
```
rynowak: Sort order is wrong

src/Microsoft.AspNetCore.Blazor/Components/IHandleEvent.cs
SteveSandersonMS: Good point - will change this.

src/Microsoft.AspNetCore.Blazor/Rendering/ComponentState.cs
SteveSandersonMS: This error is no longer applicable as a concept.

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
SteveSandersonMS: Exactly

src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Rendering/BrowserRenderer.ts
SteveSandersonMS: Correct

test/Microsoft.AspNetCore.Blazor.Test/RendererTest.cs
```diff
@@ -280,51 +280,83 @@ public void CanDispatchEventsToNestedComponents()
...
         }
 
         [Fact]
-        public void ThrowsIfComponentDoesNotHandleEvents()
+        public void DispatchesEventToTheDelegateTargetComponent()
+        {
+            // Arrange: Render parent component
+            var renderer = new TestRenderer();
+            var parentComponent = new CustomButtonParentComponent();
+            var parentComponentId = renderer.AssignRootComponentId(parentComponent);
```
rynowak: `parentComponentId` appears to be unused.

test/testapps/BasicTestApp/CascadingValueTest/CascadingValueSupplier.cshtml
```diff
@@ -16,7 +16,7 @@
     </CascadingValue>
 </CascadingValue>
 
-<p><button id="increment-count" onclick=@counterState.IncrementCount>Increment</button></p>
+<p><button id="increment-count" onclick=@(() => counterState.IncrementCount())>Increment</button></p>
```
rynowak: Is this an intentional change or did we create another gotcha here?

Based on what I can think of there's now a semantic difference between a capturing lambda and a method-group-to-delegate conversion in this case.

SteveSandersonMS: This is the "breaking change" mentioned in the PR description.

There's now a semantic difference between passing delegates (e.g., lambdas) on the component type versus passing delegates on other types. There's no difference between lambdas versus methodgroup-to-delegate on the same component type. Please let me know if you think I'm incorrect or if you disagree with this design.

SteveSandersonMS: No wait, there is a problem. Hmm... investigating options.

SteveSandersonMS:
OK, turns out this is fundamentally flawed, and after some consideration, I don't have a sufficiently simple and robust plan to salvage the concept. I should have spotted this earlier, especially given your hints about how capturing lambdas might cause issues.

I had been testing with lambdas that capture `this` (which is why I thought it worked), but not capturing anything else. As soon as you capture something other than `this`, then `delegate.Target` has to start referring to the generated closure data class (of course!). That breaks this whole approach, because it means there's a semantic difference between `onclick=@(() => DoSomething())` and `onclick=@(() => DoSomething(someLoopVariable))` (in that the latter will no longer trigger an automatic re-render). This semantic difference is not acceptable.

This whole idea of relying on the original target getting stamped on the delegate on creation seemed like a great solution because it survives the delegate getting passed through properties on any number of intermediate components. But if `delegate.Target` *doesn't* refer to the original target, by the time it's been passed through the first layer, we've totally lost track of which component instance created it.

Frustratingly the exact data I need is right there on the public `<>4__this` property on `delegate.Target`, but since there's no supported API for reading that, I don't think we can take a dependency on reading it directly.

Possible approach to salvage this:

* We could change the codegen for events on elements so that instead of generating `GetEventHandlerValue(yourSource)`, we generate `GetEventHandlerValue(this, yourSource)`
* Also change the codegen for events on components so that instead of generating `new Func<Task>(yourSource)` (etc., for all the supported delegate types), we'd generate a similar call to `GetEventHandlerValue` passing `this` and `yourSource`.
* In `GetEventHandlerValue`, we stamp the `this` value onto the constructed delegate by writing it to some `WeakDictionary<MulticastDelegate, object>` (except if there's already an entry for this delegate).

Apart from the complexity, what I dislike about this is that, if you're writing a `RenderFragment` in plain C#, you wouldn't realise that your event handler values have to be wrapped in `GetEventHandlerValue(this, value)`, so your event handlers would just appear not to work properly.

Other possible approaches:

* Declare that we never trigger re-rendering automatically after events. Developers *always* have to put `StateHasChanged` at the end of event handlers manually.
* Or, make it a compiler feature, whereby the runtime assumes you will call `StateHasChanged` manually, and have the compiler detect usages of lambdas and insert `StateHasChanged();` at the end (likewise, detect use of methodgroups, and replace them with lambdas that end with `StateHasChanged();`). This would not preserve the existing behavior of re-rendering both synchronously and asynchronously after any returned task - it would just do the latter.

@rynowak Do you agree with this assessment? Any other ideas? I'm not expecting anything in this area to make it in for 0.7.0 any more.

SteveSandersonMS:
Further idea (noting to myself for future reference): we may be able to avoid declaring any `UIEventHandler` type, and instead codegenning something like `CreateEventHandler(StateHasChanged, yourSource)`, where `CreateEventHandler` is some static method like:

```csharp
static Func<Task> CreateEventHandler(Action postEventCallback, Action eventHandler)
static Func<T, Task> CreateEventHandler<T>(Action postEventCallback, Action<T> eventHandler)
static Func<Task> CreateEventHandler(Action postEventCallback, Func<Task> eventHandler)
static Func<T, Task> CreateEventHandler<T>(Action postEventCallback, Func<T,  Task> eventHandler)
```

each of which would do something like:

```csharp
if (eventHandler.Target == typeof(TypeContainingCreateEventHandler))
{
    return eventHandler; // Pass through existing wrapped delegates without extra wrapping
}
else
{
    // Create a new delegate whose Target == typeof(TypeContainingCreateEventHandler)
    // as a signal not to wrap it further later
    return () =>
    {
        var result = eventHandler();
        postEventCallback();
        return result.ContinueWith(() => postEventCallback());
    };
}
```

Then, the recipient gets a regular `Func<Task>` they call in the usual way without needing any equivalent to the `ComponentEvents.Invoke` from this original PR, and it's safe to pass through into further levels of `CreateEventHandler` because it won't be wrapped further (except if you explicitly call it from your own lambda/method, in which case the extra wrapping is desirable anyway).

People writing `RenderFragment` in C# have the choice to call `CreateEventHandler` or to just pass a manually-written `Func<Task>` which may or may not end with `StateHasChanged()`.

Still need to think through exactly what allocations this leads to.

SteveSandersonMS:
Closing this because we're not concluded on the design yet. @rynowak and I are still discussing where we're going with bind and event handling.

Issue #1223, opened 2018-07-28
Server-side Blazor 0.5.1 doesn't update active class on NavLink element


Andrzej-W:
I have just published "default" server side application to Azure. I have made only one change in WeatherForecastService.cs
```c#
return Task.FromResult(Enumerable.Range(1, 500).Select(index => new WeatherForecast
```
As you can see service now returns 500 instead of 5 forecasts.

We can click on `Home` and `Counter` menu links and everything works as expected. Try to click on `Fetch data`. At first everything works, but `active` class on `NavLink` (strictly speaking on `<a>` element) is not removed from `Home` or `Counter` and not added to `Fetch data`.

http://blazorserwer.azurewebsites.net/

Andrzej-W:
@MarkStega it is not related to #1211 at all. This is standard project created by VS. It doesn't require any JavaScript.

Problem is visible when I try to return 500 forecasts. When I return 5 or 50 forecasts it works correctly. Probably there is subtle bug somewhere in async code on the server or on the client.

srowan:
Seeing this as well...  It's not limited to just the navlink.  Other simple bindings can fail to update as well, and it's fairly easy to reproduce.

Seems to be some kind of timing issue.

Andrzej-W:
This problem still exists in 0.6.0

vertonghenb:
Any Error message in the console?

Andrzej-W:
@vertonghenb I don't see any errors in web browser console.

enetstudio:
@Andrzej-W, did you try to place the NavLink for FetchData between the NavLink for Home and that of the Counter, and see if the issue remains ?

PR #1655, opened 2018-11-09
Handle overlapping events

SteveSandersonMS:
This is to fix the general issue with server-side Blazor that the client can trigger events more rapidly than they get processed and the results flushed back to the client. A typical scenario would be when the user is typing and triggering `input` events on each keystroke.

Previously, if the client triggered the same event multiple times in quick succession, then the server might handle the first instance by re-rendering a component and hence disposing an event handler, then it might try to process the next instance and fail because the event handler was already disposed.

With this PR, the server delays the event handler disposals until it gets confirmation from the client that it's finished updating the UI with whatever renderbatch contains the disposal.

This should eliminate the issue, because once the client has updated the UI with whatever renderbatch contains the disposal, it can no longer legitimately try to trigger new instances of the event (the client knows it has gone from the UI). Also since client/server message delivery order is guaranteed, there's no way that an earlier event triggering message could come through *after* the client's message to confirm the completion of the renderbatch in which the event is disposed.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
@@ -65,8 +65,8 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
             var componentId = AssignRootComponentId(component);
 
             // The only reason we're calling this synchronously is so that, if it throws,
-            // we get the exception back *before* attempting the first UpdateDisplay
-            // (otherwise the logged exception will come from UpdateDisplay instead of here)
+            // we get the exception back *before* attempting the first UpdateDisplayAsync
+            // (otherwise the logged exception will come from UpdateDisplayAsync instead of here)
             // When implementing support for out-of-process runtimes, we'll need to call this
             // asynchronously and ensure we surface any exceptions correctly.
             ((IJSInProcessRuntime)JSRuntime.Current).Invoke<object>(
@@ -87,7 +87,7 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
         }
 
         /// <inheritdoc />
-        protected override Task UpdateDisplay(in RenderBatch batch)
+        protected override Task UpdateDisplayAsync(in RenderBatch batch)
         {
             if (JSRuntime.Current is MonoWebAssemblyJSRuntime mono)
             {
--- a/src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Blazor.Rendering;
 using Microsoft.JSInterop;
 using Mono.WebAssembly.Interop;
 using System;
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
 {
@@ -86,7 +87,7 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
         }
 
         /// <inheritdoc />
-        protected override void UpdateDisplay(in RenderBatch batch)
+        protected override Task UpdateDisplay(in RenderBatch batch)
         {
             if (JSRuntime.Current is MonoWebAssemblyJSRuntime mono)
             {
@@ -94,12 +95,12 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
                     "Blazor._internal.renderBatch",
                     _browserRendererId,
                     batch);
+                return Task.CompletedTask;
             }
             else
             {
-                // When implementing support for an out-of-process JS runtime, we'll need to
-                // do something here to serialize and transmit the RenderBatch efficiently.
-                throw new NotImplementedException("TODO: Support BrowserRenderer.UpdateDisplay on other runtimes.");
+                // This renderer is not used for server-side Blazor.
+                throw new NotImplementedException($"{nameof(BrowserRenderer)} is supported only with in-process JS runtimes.");
             }
         }
     }
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/AutoCancelTaskCompletionSource.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/AutoCancelTaskCompletionSource.cs
@@ -5,7 +5,7 @@ using System;
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
+namespace Microsoft.AspNetCore.Blazor.Server.Circuits
 {
     /// <summary>
     /// Behaves like a <see cref="TaskCompletionSource{T}"/>, but automatically times out
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/AutoCancelTaskCompletionSource.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/AutoCancelTaskCompletionSource.cs
@@ -1,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
```diff
@@ -77,7 +79,8 @@ private ComponentState AttachAndInitComponent(IComponent component, int parentCo
         /// Updates the visible UI.
         /// </summary>
         /// <param name="renderBatch">The changes to the UI since the previous call.</param>
-        protected abstract void UpdateDisplay(in RenderBatch renderBatch);
+        /// <returns>A <see cref="Task"/> to represent the UI update process.</returns>
+        protected abstract Task UpdateDisplay(in RenderBatch renderBatch);
```
SteveSandersonMS: This is the central change in this PR. The renderer now has a way to know when the client later finishes processing the `UpdateDisplay` request. It's no longer just "fire and forget".

src/Microsoft.AspNetCore.Blazor/Rendering/Renderer.cs
```diff
@@ -217,13 +221,31 @@ private void RenderInExistingBatch(RenderQueueEntry renderQueueEntry)
...
-            var array = eventHandlerIds.Array;
-            var count = eventHandlerIds.Count;
-            for (var i = 0; i < count; i++)
+            if (eventHandlerIds.Count == 0)
             {
-                _eventBindings.Remove(array[i]);
+                return;
+            }
+
+            if (afterTask.IsCompleted)
```
SteveSandersonMS: For WebAssembly execution (where the UI updates are synchronous), this condition is always true, so the changes in this PR have no effect on either behavior or perf.

Andrzej-W:
Maybe #1223 will be fixed?

SteveSandersonMS:
@Andrzej-W #1223 is fixed by #1666.

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.JSInterop;
 using Mono.WebAssembly.Interop;
 using System;
+using System.Threading.Tasks;
```
rynowak: Sort order is wrong

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
```diff
@@ -86,20 +87,20 @@ public void Dispose()
         }
 
         /// <inheritdoc />
-        protected override void UpdateDisplay(in RenderBatch batch)
+        protected override Task UpdateDisplay(in RenderBatch batch)
```
rynowak: I know I commented on this in another PR - but we should eventually make a decision about whether we're using `Async` on async methods. We've been making lifecycle methods async as the framework matures. This is a good topic for later I suppose.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/AutoCancelTaskCompletionSource.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
```
rynowak: Is this the right namespace?

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
```diff
@@ -12,10 +14,17 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
 {
     internal class RemoteRenderer : Renderer
     {
+        // The purpose of the timeout is just to ensure server resources are released at some
+        // point if the client disconnects without sending back an ACK after a render
+        private const int TimeoutMilliseconds = 60 * 1000;
```
rynowak: Hmmm is this the right place to fix this? 

I'm going to read on assuming that we need this here, but I'm wondering if this problem with crop up other places.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
```diff
@@ -12,10 +14,17 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
...
+
         private readonly int _id;
         private readonly IClientProxy _client;
         private readonly IJSRuntime _jsRuntime;
         private readonly RendererRegistry _rendererRegistry;
+        private readonly ConcurrentDictionary<long, AutoCancelTaskCompletionSource<object>> _pendingRenders
```
rynowak: One question that I have a this point - are we OK with out-of-order delivery? Or are we confident that SignalR will enforce ordering correctly. 

For instance `_client.SendAsync` could process an arbitrary number of batches before the browser sends any acks. 

Suppose that 1000000 renders get queued up on the server, and the client is taking the time to ack and process each one. Eventually I would expect some of the render operations to time out, and the order that those exceptions are delivered to the client is arbitrary.

This seems OK - I just wanted to mention it.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
```diff
@@ -85,10 +94,48 @@ public void Dispose()
...
-            var task = _client.SendAsync("JS.RenderBatch", _id, batch);
-            CaptureAsyncExceptions(task);
+            // Prepare to track the render process with a timeout
+            var renderId = Interlocked.Increment(ref _nextRenderId);
+            var pendingRenderInfo = new AutoCancelTaskCompletionSource<object>(TimeoutMilliseconds);
+            _pendingRenders[renderId] = pendingRenderInfo;
+
+            // Send the render batch to the client
+            // If the "send" operation fails, abort the whole render with that exception
+            _client.SendAsync("JS.RenderBatch", _id, renderId, batch).ContinueWith(sendTask =>
```
rynowak: can `_client.SendAsync` throw synchronously? We'd leak in that case.

SteveSandersonMS: Good point - I'll catch that.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
SteveSandersonMS: AFAIK nothing depends on the ordering of batch completion notifications, so I think we are fine. Also in the success cases, the completion notifications are sure to be in order.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
SteveSandersonMS: Totally fine to have some way of configuring this at some point. Right now this is just a backstop to ensure we're not actually just leaking.

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
SteveSandersonMS: I don't follow. It's alphabetical. Were you expecting `System.*` to go at the top? It's regular alphabetical in other files.

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
SteveSandersonMS: Good point. Renamed to `UpdateDisplayAsync`.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/AutoCancelTaskCompletionSource.cs
SteveSandersonMS: Hmm, that was weird. Fixed.

src/Microsoft.AspNetCore.Blazor.Browser/Rendering/BrowserRenderer.cs
rynowak: We do system first everywhere

PR #1660, opened 2018-11-11
Add error detection when sending content

ebekker:
As it stands, if the attempt to write (PUT/POST) content fails, for example, due to authorization (401) or some other server-side issue (50x), the error gets swallowed up.  Ensuring the call was successful will cause it to throw otherwise.

This also preserves symmetry with the `GetJsonAsync` counterpart since its underlying call to `GetStringAsync` already exhibits this behavior.


SteveSandersonMS:
Thanks! This is now moved to https://github.com/aspnet/AspNetCore/pull/4788/files, where it will be merged shortly.

PR #1666, opened 2018-11-12
Fix intermittent render omissions. Fixes #1223.

### LINKED ISSUE #1223: Server-side Blazor 0.5.1 doesn't update active class on NavLink element

SteveSandersonMS:
As per #1223, there were some timing-based cases where server-side rendering would just fail to apply certain updates to the UI.

The underlying reason turned out to be a misunderstanding about how the SignalR APIs work: I thought that `SendAsync` would at least *send* the data synchronously, and that the returned `Task` represented the network transfer. In fact I was wrong, and `SendAsync` doesn't necessarily serialize the inputs synchronously - if the network channel is currently busy, it waits until it's free before serializing. This was problematic because we reuse `RenderBatch` instances and their buffers for subsequent renders, so if multiple renders occurred sufficiently close together in time, the 2nd or later render could have its data overwritten by the 3rd or later render before its data was dispatched to the client.

The fix here is the obvious one, i.e., to snapshot the `RenderBatch` data synchronously at the time when `UpdateDisplay` is called. This has the drawback that we are now allocating an extra `byte[]` on every render. Later on, we should consider putting in some kind of object pool for this.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/MessagePackBufferStream.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/MessagePackBufferStream.cs
@@ -8,28 +8,19 @@ using System.IO;
 namespace Microsoft.AspNetCore.Blazor.Server.Circuits
 {
     /// <summary>
-    /// A write-only stream that outputs its data to an underlying expandable
-    /// buffer in the format for a MessagePack 'Bin32' block. Supports writing
-    /// into buffers up to 2GB in length.
+    /// Provides Stream APIs for writing to a MessagePack-supplied expandable buffer.
     /// </summary>
-    internal class MessagePackBinaryBlockStream : Stream
+    internal class MessagePackBufferStream : Stream
     {
-        // MessagePack Bin32 block
-        // https://github.com/msgpack/msgpack/blob/master/spec.md#bin-format-family
-        const int HeaderLength = 5;
-
         private byte[] _buffer;
         private int _headerStartOffset;
         private int _bodyLength;
 
-        public MessagePackBinaryBlockStream(byte[] buffer, int offset)
+        public MessagePackBufferStream(byte[] buffer, int offset)
         {
             _buffer = buffer ?? throw new ArgumentNullException(nameof(buffer));
             _headerStartOffset = offset;
             _bodyLength = 0;
-
-            // Leave space for header
-            MessagePackBinary.EnsureCapacity(ref _buffer, _headerStartOffset, HeaderLength);
         }
 
         public byte[] Buffer => _buffer;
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Blazor.Server/Circuits/MessagePackBufferStream.cs
```diff
@@ -8,28 +8,19 @@
...
-    /// buffer in the format for a MessagePack 'Bin32' block. Supports writing
-    /// into buffers up to 2GB in length.
+    /// Provides Stream APIs for writing to a MessagePack-supplied expandable buffer.
     /// </summary>
-    internal class MessagePackBinaryBlockStream : Stream
+    internal class MessagePackBufferStream : Stream
     {
-        // MessagePack Bin32 block
-        // https://github.com/msgpack/msgpack/blob/master/spec.md#bin-format-family
-        const int HeaderLength = 5;
```
SteveSandersonMS: We no longer have to format the actual Bin32 block ourselves, because we're passing the `byte[]` to SignalR's `SendAsync` (instead of passing a `RenderBatch`), so SignalR now calls the appropriate MessagePack APIs to format the `byte[]` as a Bin32 block.

Hence we no longer need to write out the header manually, and I've removed all the code that did that.

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
```diff
@@ -87,7 +89,15 @@ public void Dispose()
         /// <inheritdoc />
         protected override void UpdateDisplay(in RenderBatch batch)
         {
-            var task = _client.SendAsync("JS.RenderBatch", _id, batch);
+            // Send the render batch to the client
+            // Note that we have to capture the data as a byte[] synchronously here, because
+            // SignalR's SendAsync can wait an arbitrary duration before serializing the params.
+            // The RenderBatch buffer will get reused by subsequent renders, so we need to
+            // snapshot its contents now.
+            // TODO: Consider using some kind of array pool instead of allocating a new
```
rynowak: I think we discussed this before. It would actually be very easy if not for the fact that the message pack library is the one creating the arrays. I think we need to be really concerned about LOH allocations but not in 0.7

src/Microsoft.AspNetCore.Blazor.Server/Circuits/MessagePackBufferStream.cs
SteveSandersonMS: Reopened #1132

src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
SteveSandersonMS: Reopened #1132

PR #1672, opened 2018-11-13
Serialize server-side renders. Fixes #1573

### LINKED ISSUE #1573: Blazor Server side, child component not always updates view
I notices that a child component that used async entityframework not always got updated. So I tried to simulate the situation without Entity framework. 

I created a child component with this code:

```
@if(list == null)
{
    <div>
        Loading....
    </div>
}
else
{
    foreach(var i in list)
    {
        <div>
            @i
        </div>
    }
}

@functions
{
    protected List<string> list = null;

    {
        await Task.Run(async () =>
        {
            await Task.Delay(3000);

            list = new List<string>();
          
            list.Add("Some item 1");
            list.Add("Some item 2");
            list.Add("Some item 3");
            list.Add("Some item 4");
        });

        this.StateHasChanged();
    }
}
```

In de main page: index.cshtml insert A few child components like:

```
<Child></Child>
<Child></Child>
<Child></Child>
<Child></Child>
```

Open debugging with f5.
When the page has been loaded in chrome, reload with f5.

After a few time, a few components won't show but keep saying: loading....

SteveSandersonMS:
Another case where the asynchrony of server-side rendering needed some extra work to ensure the correct behaviors. This was a simple thread safety issue inside `Renderer`.

To a first approximation, I could have fixed it just by putting a `lock(_renderer) { ... }` around `RenderHandle`'s call to `_renderer.AddToRenderQueue`. However that could lead to deadlock because rendering can be re-entrant (a child can cause its own parent to re-render, e.g., by invoking some callback passed by the parent).

The solution here is to enqueue render calls on the circuit sync context. This causes some extra allocations per-render but provides an easy-to-reason-about solution. Plus it ensures that component lifecycle methods such as `Init` (which is invoked by the renderer) get executed serially within the sync context, which better matches up with the WebAssembly execution style.

### DIFF
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/DefaultCircuitFactory.cs
@@ -40,8 +40,8 @@ namespace Microsoft.AspNetCore.Blazor.Server.Circuits
             var scope = _scopeFactory.CreateScope();
             var jsRuntime = new RemoteJSRuntime(client);
             var rendererRegistry = new RendererRegistry();
-            var renderer = new RemoteRenderer(scope.ServiceProvider, rendererRegistry, jsRuntime, client);
             var synchronizationContext = new CircuitSynchronizationContext();
+            var renderer = new RemoteRenderer(scope.ServiceProvider, rendererRegistry, jsRuntime, client, synchronizationContext);
 
             var circuitHost = new CircuitHost(
                 scope,
--- a/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
+++ b/src/Microsoft.AspNetCore.Blazor.Server/Circuits/RemoteRenderer.cs
@@ -24,6 +24,7 @@ namespace Microsoft.AspNetCore.Blazor.Browser.Rendering
         private readonly IClientProxy _client;
         private readonly IJSRuntime _jsRuntime;
         private readonly RendererRegistry _rendererRegistry;
+        private readonly SynchronizationContext _syncContext;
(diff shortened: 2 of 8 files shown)


SteveSandersonMS:
Merging now so I can produce the 0.7.0 binaries.

@rynowak I'd still appreciate knowing of any comments/concerns you have about this as we can still make changes before 0.7.0 ships (or after).

rynowak:
> Plus it ensures that component lifecycle methods such as Init (which is invoked by the renderer) get executed serially within the sync context, which better matches up with the WebAssembly execution style.

This should already be the case...

I think this solution is fine. I agree with your assessment that we would need to create a rendering queue rather than using simple locking, and the sync context is already a queue so it works well.

SteveSandersonMS:
> This should already be the case...

If a component does something to run code off the sync context (e.g. queue a thread pool work item), that code can trigger a render by calling StateHasChanged. Prior to this PR, the triggered render would run in the arbitrary context, potentially simultaneously with other renders of the same component due to the thread safety problem. At least that’s my understanding, since previously we were not forcing the execution back onto the sync context.

rynowak:
ah I see. Yeah I was thinking about this in the context of async/await code. This makes things a LOT more safer - other UI frameworks like winforms/wpf would throw in this case, but I like what we're doing.

SteveSandersonMS:
I *was* seeing the issue in the context of async/await code (just awaiting Task.Delay as in the E2E test in this PR). Would you expect it not to be an issue in the case of this E2E test? If so maybe we also have another problem.

rynowak:
I would expect the code after the await to resume on the sync context - like this test: https://github.com/aspnet/Blazor/blob/master/test/Microsoft.AspnetCore.Blazor.Server.Test/Circuits/CircuitSynchronizationContextTest.cs#L377

PR #1678, opened 2018-11-14
(TS) Object is possibly 'null' fix

StevenTCramer:
micro fix for #1670 
add the non-null assertion operator.


SteveSandersonMS:
Thanks! This change was already applied as part of migrating the code from this repo into the AspNetCore repo.

Issue #1388, opened 2018-09-05
warning BL9993: Component parameter 'any' is marked public, but component parameters should not be public


Lupusa87:
**This warning has no sense!**

1. If I want to access child **property or action** from parent using component reference I need public property (why should I add extra public method to modify private property) or action (especially).

2. I often access component's properties or actions from any place inside application, logically I need public properties or actions to have control of them.

I use public access level but VS is giving this warning.

I can use **internal** access level but I think it will be also against the logic based with blazor analyzer requires private.

My questions:
**1. What are arguments for requesting private?**

**2. Why can not be used public or internal?**

**3. If there is good reason then what can we do in above cases?**

@SteveSandersonMS

Lupusa87:
This contains cases mentioned above https://github.com/aspnet/Blazor/issues/1355#issuecomment-418591770

SteveSandersonMS:
The analyzer that tells you to use nonpublic (i.e., `private`, `protected`, or `internal`) only applies to component *parameters*, not to any other methods, properties, or fields. So if you want to have public methods on your components you absolutely can and should - that's a totally standard pattern.

The only thing we're recommending not to do is have public *parameters* (i.e., `[Parameter]`). That's because there's only one correct way to set them, which is using the attributes on components in markup. If you try to set them any other way (e.g., from procedural C# code), the effect won't be what you want, because the receiving component won't re-render automatically. If you *do* want it to re-render, then give it some public method that receives the new value, assigns its own property, and calls `StateHasChanged`.

Finally, this is only guidance. If you don't like this recommendation, you can turn it off in your own project: https://stackoverflow.com/questions/34720443/suppressing-issues-from-roslyn-code-analyzers

SteveSandersonMS:
I agree we should do that. PRs welcome!

kswoll:
@SteveSandersonMS alright, here's my stab at this! :)  https://github.com/aspnet/Blazor/pull/1702

PR #1702, opened 2018-11-19
Ignore non-public setters in analyzer

kswoll:
Currently if you have a property like:

```
[Parameter] public string Foo { get; private set; }
```

The analzyer emits a warning about public properties with a `[Parameter]` attribute specified.  Since the reason for this warning is about calling the setter (and mutating state at improper points in the rendering cycle) explicit non-public setters should be ignored.

See https://github.com/aspnet/Blazor/issues/1388#issuecomment-439650736 for more elaboration


kswoll:
Any thoughts on the failure?  It passes locally for me and doesn't seem at all related to my changes.  Flaky?

```

  Starting test execution, please wait...
  [xUnit.net 00:00:01.2289096]     Microsoft.JSInterop.Test.JsonUtilTest.CanSerializeClassToJson [FAIL]
  Failed   Microsoft.JSInterop.Test.JsonUtilTest.CanSerializeClassToJson
```

SteveSandersonMS:
Thanks for this! It's now moved to https://github.com/aspnet/AspNetCore/pull/4594.

PR #1712, opened 2018-11-22
linking to some existing docs

aghanem:
This is to add links to the existing Blazor docs site. In addition, we may need to convert this item"Live reloading in the browser during development" to a link to the debugging page but I first the debugging docs need some enhancements like adding IDE debugging and details about live reloading (maybe using dotnet command line tool ?)


SteveSandersonMS:
Thanks! This PR is now moved to https://github.com/aspnet/AspNetCore/pull/4592.

Issue #1709, opened 2018-11-21
BeforeRender event


grahamehorner:
Please consider adding a IHandleBeforeRender interface with the method void OnBeforeRender(); to allow components to perform some action prior to rendering; i.e. changing parameters as a result of validation errors or fetching/triggering data cache expiry/refresh

vertonghenb:
`OnInit`
`OnParametersSet`
Are already present.
https://blazor.net/docs/components/index.html#lifecycle-methods

grahamehorner:
@vertonghenb for my understanding is OnInit/OnParametersSet fired every time and before the component is rendered?  my use case is that I wish to do some work just prior to the component is rendered, and then have this show in the component UI and/or trigger some background action.

If OnInit is fired every time the component is rendered, then I would propose that the name is misleading and should be changed.

michaelvolz:
@vertonghenb  `OnBeforeRender` would be useful if data binding changed and the state is updated and only 
`OnAfterRender`|`OnAfterRenderAsync` is triggered but not the other lifecycle events, If I am not mistaken.
So `OnInit` or `OnParametersSet` wouldn't help here.

Or did I miss something?

grahamehorner:
Also a developer may wish to prevent rendering after initial and settings parameters, based on some computation or interop call.

PR #1716, opened 2018-11-22
OnBeforeRender

grahamehorner:
Added OnBeforeRender event to component ref: https://github.com/aspnet/Blazor/issues/1709

### DIFF
--- a/src/Microsoft.AspNetCore.Components.Browser.JS/package-lock.json
+++ b/src/Microsoft.AspNetCore.Components.Browser.JS/package-lock.json
@@ -16,7 +16,7 @@
       "integrity": "sha512-Q/rSs/UDzw0iwhqZILzMMP0m/lTVNSQDS9aNkBTkt8aAAKAEMriCMTy0Ahx9xPuMTzFtyPRjNgEZfEkWUHajcg==",
       "dev": true,
       "requires": {
-        "msgpack5": "4.2.0"
+        "msgpack5": "^4.0.2"
       }
     },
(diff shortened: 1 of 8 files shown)


grahamehorner:
@SteveSandersonMS I totally agree a virtual StateHasChanged would be great; I'll open an issue and do a POC with reference back to this

michaelvolz:
The less lifecycle methods the better, great idea!

markzielinski:
@SteveSandersonMS, it would be really nice to have OnBeforeRender as there are many use cases for wanting to do work before the rendering actually takes place, just as there will be many use cases for wanting to do work after the rendering takes place. As there is already a OnAfterRender it makes complete sense here to have a OnBeforeRender for the exact same reasons and to provide symmetry. Adding one lifecycle method isn't suddenly going to turn into a whole bunch of lifecycle methods added.

This was originally discussed two years ago and at that time there was a pull request submitted for this functionality, but it was not accepted in favor of making StateHasChanged virtual as discussed here. As there hasn't been progress made on this in those two years because of the complexity of determining what impacts this would have and what the tradeoffs are, this tells me that this was the wrong decision and the original request and solution should be revisited as it's the simplest solution and shouldn't cause any impacts.

grahamehorner:
Happy to do another PR for just the virtual StateHasChanged along with unit tests @markzielinski if I ping you when I have this ready would you mind reviewing and commenting to see if your use case is covered then @SteveSandersonMS can look at the implications and/or implementation

PR #1706, opened 2018-11-21
Code migration, step 1

SteveSandersonMS:
We're in the process of migrating code from the Blazor repo into the [central ASP.NET Core repo](https://github.com/aspnet/aspnetcore). This is because (1) we're moving most of our repos into the central one, and (2) we've committed to ship the Razor Components programming model in ASP.NET Core 3.0.

For more info about what we've announced will ship in 3.0, see [this blog post](https://blogs.msdn.microsoft.com/webdev/2018/10/02/blazor-0-6-0-experimental-release-now-available/#razor-components-to-ship-with-asp-net-core-in-net-core-3-0). TLDR: server-side Blazor ships in 3.0 and gets the name "Razor Components", while client-side (WebAssembly) Blazor remains experimental to give more time to make the underlying runtime truly production grade.

### Migration steps

1. **Reorganise all the files <-- this PR**
2. Rename projects/packages/namespaces
3. Obtain Mono binaries from some new external package instead of hosting them here
4. Actually move the code into the mondo-repo
5. All ongoing refactoring, e.g., to take dependencies on 3.0 packages/tooling, to separate out WASM-specific aspects (e.g., in build system and middleware), etc. This is an open-ended task.

### Review notes

There's nothing interesting in this PR. It's just a vast amount of moving files, tracking down all their references to each other, and updating all those references. The output from the build system should be unchanged.

The reason I separated the "reorganise files" step from "rename content" (e.g., namespaces) is because it's the "reorganise" step that causes all the unreadable noise in the Git diffs. Also, the "reorganise" step is nearly impossible to reconcile with unrelated changes, so until this is merged, we can't make any other changes in the repo.

So, @rynowak @javiercn @pranavkm @danroth27 @mkArtakMSFT: **please don't commit any other changes to the Blazor repo until this PR is merged**. I don't think I could realistically rebase this PR onto any new work!

PR #1718, opened 2018-11-23
Code migration phase 3 - Move Mono binaries to their own package

SteveSandersonMS:
Continuation of #1706 and #1708

### Migration steps

1. ~~Reorganise all the files~~ Done in #1706
2. ~~Rename projects/packages/namespaces~~ Done in #1708 
3. **Obtain Mono binaries from some new external package instead of hosting them here <-- this PR**
4. Actually move the code into the mondo-repo
5. All ongoing refactoring, e.g., to take dependencies on 3.0 packages/tooling, to separate out WASM-specific aspects (e.g., in build system and middleware), etc. This is an open-ended task.


SteveSandersonMS:
Merged now because I need the CI to produce an official build of the new `Microsoft.AspNetCore.Blazor.Mono` package so it can be published prior to the next stage of work.

PR #1725, opened 2018-11-26
Decouple migrating code from non-migrating code

SteveSandersonMS:
This PR changes the code that will migrate into the mondorepo so that it only consumes the nonmigrating code (`Microsoft.AspNetCore.Blazor.Mono`, `Microsoft.AspNetCore.Blazor.BuildTools`, JSInterop) via package references and no longer assumes the presence of their source files.

The nonmigrating code will continue to be built and shipped from the Blazor repo.


src/Microsoft.AspNetCore.Components.Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.csproj
```diff
@@ -12,16 +12,12 @@
...
-    <ProjectReference Include="..\..\blazor\src\Microsoft.AspNetCore.Blazor.BuildTools\Microsoft.AspNetCore.Blazor.BuildTools.csproj" PrivateAssets="all" ReferenceOutputAssembly="false" />
     <PackageReference Include="Microsoft.Extensions.FileProviders.Embedded" Version="$(AspNetCorePackageVersion)" />
     <WebpackInputs Include="**\*.ts" Exclude="node_modules\**" />
     <WebpackInputs Include="..\..\modules\jsinterop\src\Microsoft.JSInterop.JS\src\**" />
   </ItemGroup>
 
-  <Import Project="..\..\blazor\src\Microsoft.AspNetCore.Blazor.BuildTools\BuildToolsExe.props" />
-
   <Target Name="EnsureNpmRestored" Condition="!Exists('node_modules')">
-    <Exec Command="$(BlazorBuildToolsExe) checknodejs -v 8.3.0" />
```
SteveSandersonMS: I think we can just live without this. It only affects people building from source, not people consuming the shipped packages, and it only affects the case where the build was going to fail anyway because you don't have Node on `PATH`.

src/Microsoft.AspNetCore.Components.Browser.JS/src/Boot.Server.ts
```diff
@@ -1,6 +1,5 @@
-import '../../../modules/jsinterop/src/Microsoft.JSInterop.JS/src/Microsoft.JSInterop';
+import '@dotnet/jsinterop';
 import './GlobalExports';
-import * as Environment from './Environment';
```
SteveSandersonMS: This import was unused.

src/Microsoft.AspNetCore.Components.Browser.JS/src/Boot.WebAssembly.ts
```diff
@@ -1,10 +1,9 @@
-import '../../../modules/jsinterop/src/Microsoft.JSInterop.JS/src/Microsoft.JSInterop';
+import '@dotnet/jsinterop';
 import './GlobalExports';
 import * as Environment from './Environment';
 import { monoPlatform } from './Platform/Mono/MonoPlatform';
 import { getAssemblyNameFromUrl } from './Platform/Url';
 import { renderBatch } from './Rendering/Renderer';
-import { RenderBatch } from './Rendering/RenderBatch/RenderBatch';
```
SteveSandersonMS: This import was unused.

src/Microsoft.AspNetCore.Components.Browser.JS/src/BootCommon.ts
```diff
@@ -24,7 +24,7 @@ function loadResourceFromElement(element: HTMLElement) {
   return new Promise((resolve, reject) => {
     element.onload = resolve;
     element.onerror = reject;
-    document.head.appendChild(element);
+    document.head!.appendChild(element);
```
SteveSandersonMS: Newer versions of the TypeScript compiler complain without this additional type hint.

SteveSandersonMS:
Merging now because (1) it's just a boring bunch of build config changes and (2) I need it to be in to proceed with the merge work.

PR #1734, opened 2018-11-28
Clean up after mondorepo migration

SteveSandersonMS:
This PR removes all the migrated projects, so the Blazor repo only builds the following:

 * `Microsoft.AspNetCore.Blazor.Mono`, because we want to keep the Mono runtime binaries separate from the mondorepo
 * `Microsoft.AspNetCore.Blazor.BuildTools`, because this is only needed when packaging the Mono runtime binaries above. It contains the `ILWipe` mechanism.

This PR also updates the repo's `README.md` to explain this change.


Eilon:
Sorry for the very long delay; I've disabled AppVeyor and Travis on this repo.

PR #1807, opened 2019-05-14
Update Mono (early preview 6)

SteveSandersonMS:


SteveSandersonMS:
This brings in another good perf improvement from the Mono team - benchmarks show roughly a 10% speed gain across the board.

However, we can't yet merge it because the updated linker does something that breaks all Blazor apps by default. I had to disable linking in order to run the benchmarks. With the linker on, the initial call to `Program.Main` fails with:

image

@lewing You mentioned that you anticipated us running into trouble with the updated linker. Do you have any hints on what specifically has changed and what we should to to avoid this? Are there some new command-line options we should be using?

Presumably the linker has stripped out some method that is required, but how can we diagnose which methods it is? Once we know what part of the config needs to be customised, we can add it to our built-in linker config [here](https://github.com/aspnet/AspNetCore/blob/master/src/Components/Blazor/Build/src/targets/BuiltInBclLinkerDescriptor.xml).

cc @danroth27

lewing:
@SteveSandersonMS it's possible adding `--explicit-reflection --disable-opt unreachablebodies` to the linker args will resolve this but I haven't had a chance to verify.

SteveSandersonMS:
The uncompressed `mono.wasm` is now 144KB smaller than before, and the uncompressed `mscorlib.dll` is 503KB smaller! This is independent of the linker improvements. Not sure how that was achieved but this is awesome.

cc @danroth27

Suchiman:
@SteveSandersonMS could be https://github.com/mono/mono/pull/14481 when looking at the recent history.

rynowak:
> 
> 
> The uncompressed `mono.wasm` is now 144KB smaller than before, and the uncompressed `mscorlib.dll` is 503KB smaller! This is independent of the linker improvements (as in, these are the changes _before_ running the linker). Not sure how that was achieved but this is awesome.
> 
> cc @danroth27

Wow that's a super impressive result!

PR #1809, opened 2019-05-23
Convert repo to use Arcade

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore-Internal/issues/2360


natemcmaster:
FYI @SteveSandersonMS

anurse:
Will this also publish a manifest to BAR? We should include this in our discussions for automatic publishing.

natemcmaster:
Yes.

Directory.Build.props
```diff
@@ -1,29 +1,15 @@
...
+
+  <PropertyGroup>
+    <Product>Microsoft ASP.NET Core Blazor</Product>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageProjectUrl>https://asp.net/</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/aspnet/Blazor</RepositoryUrl>
+    <RepositoryType>git</RepositoryType>
+    <StrongNameKeyId>MicrosoftAspNetCore</StrongNameKeyId>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <LangVersion>7.3</LangVersion>
```
JunTaoLuo: Why are we specifying 7.3? Why not 8.0?

natemcmaster: This is the value that was here before this change. Git is just showing a diff because I deleted unnecessary code that was on the lines before and after. @SteveSandersonMS if you want a higher lang version, you're welcome to change it.

dougbu:
We have not published anything since this went in
- Created dotnet/arcade#2886 to complete async publishing on-boarding (FYI @JohnTortugo and @natemcmaster)
- @rynowak @SteveSandersonMS looks like we're not signing the generated assemblies included in Microsoft.AspNetCore.Blazor.Mono.0.10.0-preview6.19273.9.nupkg. See https://dnceng.visualstudio.com/internal/_releaseProgress?_a=release-environment-logs&releaseId=8012&environmentId=22206 which failed due to this issue.
  - If they shouldn't be signed, add them to the exclusions file
  - If they should, do the needful 😺 

Side question:
- @natemcmaster do we need https://dnceng.visualstudio.com/internal/_release?view=all&definitionId=7 anymore?

natemcmaster:
Thanks Doug. I'm aware of all those problems and fixed them yesterday. I added signing exclusions and updated arcade this morning to support publishing to the aspnet-blazor static feed.

Re: https://dnceng.visualstudio.com/internal/_release?view=all&definitionId=7 - still working on getting blazor into the official .net core 3 release pipelines. Until then, lets keep this. I changed triggers to manual for now since the only stage left is deployment to nuget.org.

PR #1818, opened 2019-06-24
Upgrade Mono binaries for preview 7 release

SteveSandersonMS:


SteveSandersonMS:
This looks like another good upgrade according to benchmarks:

 * Compressed default app size increases by about 30KB, but...
 * Speed benchmarks show 10-20% improvements across the board

PR #1834, opened 2019-07-22
Upgrade Mono binaries for Preview 8

SteveSandersonMS:


src/Microsoft.AspNetCore.Blazor.Mono/tools/binaries/illink/illink.runtimeconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "runtimeOptions": {
-    "tfm": "netcoreapp3.0",
+    "tfm": "netcoreapp2.0",
     "framework": {
       "name": "Microsoft.NETCore.App",
-      "version": "3.0.0-alpha"
+      "version": "2.0.0"
```
rynowak: Do you know the reason who this isn't pointing in 3.0 anymore? IIRC in the past this was needed to work for users that didn't have a 2.0 runtime installed. Has something changed here?

SteveSandersonMS: > Has something changed here?

For a while we've had this super risk-averse technique of specifying `"version": "3.0.0-alpha"`, but it doesn't really make sense given that this file also specifies `"rollForwardOnNoCandidateFx": 2` (that was added *after* we previously reported the issues with running it on machines with only .NET Core 3.0 previews).

I think by now we can trust that `"rollForwardOnNoCandidateFx": 2` will do its job properly in the current 3.0 preview SDKs.

PR #1838, opened 2019-07-27
Update branding to 3.0.0-preview8

dougbu:


dougbu:
This should go in _before_ we branch for preview8

PR #1852, opened 2019-08-12
Upgrade Mono for preview 9

SteveSandersonMS:


SteveSandersonMS:
From benchmarking, it looks like the app size is unchanged (to within < 10KB), and the runtime perf has regressed by about 3%.

Assuming this 3% reduction in speed is due to some improved functionality or correctness it seems OK and within the normal range of fluctuations between versions, so I'll proceed with it.

SteveSandersonMS:
Merging as there's nothing to code review.

PR #1896, opened 2019-09-23
Switch to embedded PackageIcon

wtgodbe:
Part of aspnet/AspNetCore-Internal#3146

Local build passes the smell test with these changes. I'm still unsure on whether or not we need to include a reference to the package icon in the `.nuspec` - the "old way" did so automatically, the "new way" does not (https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file). CC @nkolev92 who may know more about that question.


wtgodbe:
Looks like we don't need to worry about the `.nuspec`, as per https://github.com/aspnet/EntityFramework6/pull/1286#issuecomment-534307966

wtgodbe:
With the last change, this now includes the icon metadata in all .nuspec's, including the manually generated ones - @dougbu PTAL

wtgodbe:
Closing in favor of https://github.com/aspnet/Blazor/pull/1897

PR #1897, opened 2019-09-24
[release/3.0] Switch to embedded package icon

wtgodbe:
Contributes to aspnet/AspNetCore-Internal#3103


wtgodbe:
Just updated to the GA SDK

dougbu:
@wtgodbe the branding for Blazor won't change in 3.0.x. Yes, the build number will increase but that's it.

PR #1904, opened 2019-09-30
Update Mono WebAssembly for 3.1 preview2

SteveSandersonMS:
This only changes the Mono binaries.

It doesn't yet change any of the boot scripts as needed to work around the iOS 13 issue - that will be done separately in the aspnetcore repo, since those scripts live there.


SteveSandersonMS:
I checked, and this makes no difference to either app size or benchmark results versus the previous drop. Still worth updating to latest, though!

SteveSandersonMS:
Merging because I've manually checked it works, and it's meaningless to code review this change.

PR #1903, opened 2019-09-29
Update branding to 3.1.0-preview2

dougbu:

PR #1906, opened 2019-09-30
Update branding to 3.1.0-preview2

dougbu:


dougbu:
This reanimates #1903. I reverted that change (29e34d1ea8d3) directly in the 'release/3.1' branch to avoid intermediate merges.

PR #1937, opened 2019-11-18
Enable building `blazor-wasm` branch

mkArtakMSFT:
Making sure the blazor-wasm branches are being built.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,6 +24,7 @@ trigger:
   batch: true
   branches:
     include:
+    - blazor-wasm
     - master
     - release/*
 


dougbu:
Why not rename the branch? `blazor-wasm` is one we'll be releasing from, correct?

mkArtakMSFT:
> Why not rename the branch? `blazor-wasm` is one we'll be releasing from, correct?

We've already included this name in multiple communications / documents, and don't want to start changing stuff as that way we may miss something somewhere.

dougbu:
> We've already included this name in multiple communications / documents, and don't want to start changing stuff as that way we may miss something somewhere.

Alright. Given that, should the `blazor-wasm` branches have branch protection rules at least similar to those we have for `master` in the relevant repos?

PR #1938, opened 2019-11-20
Stop wiping System.Net.Http

SteveSandersonMS:
This used to be necessary to avoid pulling in 500KB of code that wouldn't work anyway.

However now Mono's BCL has a native WebAssembly-compatible implementation of `HttpClient`, `HttpMessageHandler`, etc., we no longer need to strip this out. And in fact we need to stop stripping it out, as it's breaking scenarios like:

```cs
var handler = new HttpClientHandler();
var httpClient = new HttpClient(handler);
httpClient.GetJsonAsync(...);
```


SteveSandersonMS:
@danroth27 I know some people were saying our `HttpClientHandler` stuff still isn't working despite our move to the Mono built-in bindings. The change here fixes the last remaining issue I know about currently.

PR #1950, opened 2019-11-26
Upgrade Mono (November 2019 edition)

SteveSandersonMS:
This updates Mono to c4136ebd945, which is the first build that includes @vargaz's non-recursive interpreter change. Once CI produces this package I'll try to verify whether it now works on iOS 13.

Note: this is going into `master`, which currently builds using the 5.0 SDK. This isn't really the configuration we want, since we're flowing this into `blazor-wasm` which uses the 3.1 SDK. We should fix this before the Dec 3rd release.


pranavkm:
> Note: this is going into master, which currently builds using the 5.0 SDK. 

Is there a reason this couldn't target the `blazor-wasm` branch?

SteveSandersonMS:
> Is there a reason this couldn't target the blazor-wasm branch?

There isn't one - it was deleted.

PR #1958, opened 2019-12-13
Update mono (2019-12-13)

SteveSandersonMS:
Updates Mono to a build that hopefully includes the new client websocket support.

Instead of doing the upgrade manually, I've finally added a script that does it, plus [updated the instructions to describe the process of using that script](https://github.com/aspnet/Blazor/blob/4b49ece533d4e03e236000b9d04cf77cfec05ac0/src/Microsoft.AspNetCore.Blazor.Mono/HowToUpgradeMono.md). @javiercn @pranavkm Could you please check [these instructions](https://github.com/aspnet/Blazor/blob/4b49ece533d4e03e236000b9d04cf77cfec05ac0/src/Microsoft.AspNetCore.Blazor.Mono/HowToUpgradeMono.md) and let me know if you think this is sufficient for someone other than me to do a future upgrade? I don't mind doing it, but it's better if the whole team is enabled to do so.


javiercn:
Looks great!

SteveSandersonMS:
I checked, and the app sizes and perf numbers are more or less the same as before.

*However*, with this build, `System.Net.WebSockets.ClientWebSocket` now just works (at least in the "hello world" scenario I tried)! Great job, @kjpou1 👍 

@anurse Now this is merged here, the change will propagate to the `blazor-wasm` branch of the aspnetcore repo within the next day or so. Do you want to re-try the SignalR client at that point?

PR #1960, opened 2020-01-02
Stop wiping HttpClientHandler

SteveSandersonMS:
Context: in the build process for `Microsoft.AspNetCore.Blazor.Mono`, we have an optimization step that "wipes" (i.e., replaces with a single `throw` instruction) specified methods from the Mono BCL. Historically this allowed us to disconnect large pieces of the BCL from the linker-rooted parts so the resulting apps were much smaller. We used this on `System.Net.Http.HttpClientHandler` because historically it could not be used (you had to use a different WASM-specific `HttpClientHandler`) and it pulled in large amounts of code from `ServicePointManager` and so on.

However, we no longer need to wipe `HttpClientHandler` (because it no longer references large dependencies in the current Mono BCL) and we don't want to (because you actually can use it now - it now works with WebAssembly by default). Additionally, it's really valuable to stop wiping it, because doing so is breaking the gRPC-Web prototype.

Technically after this PR we no longer use the wiping infrastructure at all. Eventually we might want to remove it, but I don't want to do so immediately because we haven't finished our size-optimization investigations yet, and they may well want to use this.


SteveSandersonMS:
BTW I checked and this change doesn't result in the default app size growing. And it does result in gRPC-Web working.

rynowak:
:shipit:

PR #1966, opened 2020-01-14
Upgrade Mono binaries to c32414966c7

mkArtakMSFT:
Also updated the `Upgrade` script to use the new `dotnat.wasm` and `dotnet.js` names instead of `mono.wasm` and `mono.js` respectively.

Addresses https://github.com/dotnet/aspnetcore/issues/18316


mkArtakMSFT:
I'm not going to merge this yet, as need to still validate these changes. However, @SteveSandersonMS and @pranavkm, I'm thinking about the rename from 'mono.wasm' to `dotnet.wasm`. Do we also need to react to that change somewhere?

SteveSandersonMS:
> Do we also need to react to that change somewhere?

Definitely. Probably in several places. I’d start by searching through all the sources in the Components dir in the aspnetcore repo. I’m happy to help track this down.

mkArtakMSFT:
Here is the reaction: https://github.com/dotnet/aspnetcore/pull/18345

s4lish:
it's very nice and clean if we can write @if and @for or @foreach inside HTML Elements looks like Angular or Vue.js.

PR #1976, opened 2020-02-21
Updating build to https://jenkins.mono-project.com/job/test-mono-mainline-wasm/4695

pranavkm:


pranavkm:
The merging part isn't automated unfortunately. I had asked @anurse about onboarding this to the bot that knows about auto-merge, but it sounds like there's work there. The other alternative is to make the `master` branch non-protected, but I don't think we necessarily want that.

PR #1977, opened 2020-02-22
Updating build to https://jenkins.mono-project.com/job/test-mono-mainline-wasm/4699

pranavkm:


SteveSandersonMS:
@pranavkm Not sure how meaningful it is to "review" this kind of PR.

Normally the acceptance criteria is:

 * All our E2E tests still pass (which I guess is getting checked upstream when `dotnet/blazor` is flowed into `dotnet/aspnetcore`)
 * Size and execution speed is not significantly regressed

Is there anything checking the second point?

pranavkm:
@SteveSandersonMS the `master` branch is protected, so pushing to master isn't an option. The only option there is to send a PR and merge it in once the build passes. We have a bot that can automate merging in other repos, but it wasn't clear how much work that is (/cc @anurse). 

>  Is your recommendation that we merge the PRs and then later, if the numbers regress, we look back to see which change was the first one introduced before the regression? 

Benchmarking has automation to file issues if it notices regressions, we should rely on that. We include details of the Git commit per run, so that should help us pinpoint the range of commits that could have regressed a build.

anurse:
> the `master` branch is protected, so pushing to master isn't an option.

Anything is an option. Perhaps this is something that should be raised at DoI or on an email thread explaining what you're trying to accomplish. If removing the branch protection is the best option, I'm not opposed to it out of principle.

SteveSandersonMS:
Thanks for the answers!

I agree it would be good to find some way to auto-merge if we can. This repo isn't used for anything apart from repackaging Mono WebAssembly so there's no reason we should impose the same protection rules that we do for (say) dotnet/aspnetcore.

pranavkm:
> Anything is an option. Perhaps this is something that should be raised at DoI or on an email thread explaining what you're trying to accomplish. If removing the branch protection is the best option, I'm not opposed to it out of principle.

Sounds good.

anurse:
(Testing msftbot's automerge, FYI)

PR #1985, opened 2020-03-02
Use version numbers in dotnet.js pack output

pranavkm:


pranavkm:
image

While this means you get different stamps between different builds even if the content is identical, I don't think that's particularly problematic. Especially once Blazor RTMs.

pranavkm:
> I have a slight preference for the filename format looking more like our assembly filenames, e.g., dotnet.3.2.0-suffix.js

Ok. I'd assumed Blazor would be confused if it saw multiple "dots" in the filename. Should be trivial to change it to use the original file name as is.

pranavkm:
Removed the transformation. We can figure out if the client needs to be updated in the integration PR.

PR #1987, opened 2020-03-03
Delete DELETEME.md

anurse:
I was hoping to beat the auto-merge but I failed :(

### DIFF
--- a/DELETEME.md
+++ b/DELETEME.md
@@ -1,3 +1,0 @@
-# Testing Page
-
-Testing msftbot Auto-Merge. If you see this file in the main branch, feel free to delete it.


anurse:
@pranavkm do we want auto-merge to be immediate like this? In dotnet/aspnetcore we require that the PR be open for at least 60 minutes (just to make sure someone is able to look at it). I notice we don't require review approvals here, but given the automated nature of this, that seems OK. Note that it is possible to have the msftbot give review approval for a PR, so we could require approval for PRs *in general* but exempt any automation from that process (that's what we do in dotnet/aspnetcore).

pranavkm:
Yeah, that should be fine. We're gating the changes in the dependency update PR in to dotnet/aspnetcore. There's really nothing interesting for us to vet in the automated PRs we have in this repo.

pranavkm:
Also thanks for setting this up!

PR #1995, opened 2020-03-08
[release/3.1] Update hosted and BYOC agents used in our builds

dougbu:
- dotnet/aspnetcore-internal#3540


dougbu:
**NO-MERGE** because the branches aren't open

dougbu:
@Pilchie FYI if we're right about the merge bot and GitHub behaviours, my merge here will lead to a new merge PR that shows changes in 'master' even though 4eb192390b9ba800b6e01d94ba6d7d4c62db64a5 is already part of that branch. The problem seems to be the final merge commit (chicken and egg).

PR #2043, opened 2020-05-06
Updating build to commit https://github.com/mono/mono/commit/1a6e64a9381

pranavkm:


pranavkm:
@mkArtakMSFT should we turn off automatic updates to mono? This build was triggered as a result of a test change.

